PR #6, opened 2018-03-20
Better associations handling

goldoraf:


packages/cozy-client/src/associations/HasManyAssociation.js
```diff
@@ -0,0 +1,62 @@
+export default class HasManyAssociation {
+  constructor(target, name, { get, query, mutate }) {
+    this.target = target
+    this.name = name
+    this.get = get
+    this.query = query
+    this.mutate = mutate
+  }
+
+  get data() {
```
ptbrowne: With getter it seems it is not possible to target IE8.

packages/cozy-client/src/associations/HasManyAssociation.js
```diff
@@ -0,0 +1,62 @@
...
+      ...target,
+      relationships: {
+        ...target.relationships,
+        [this.name]: {
+          ...target.relationships[this.name],
+          ...updateFn(target.relationships[this.name])
+        }
+      }
+    }
+  }
```
ptbrowne: Shouldn't this two functions be located in Association, they seem generic enough ?

packages/cozy-client/src/associations/HasManyAssociation.js
goldoraf: And I don't give a shit ;)

packages/cozy-client/src/associations/HasManyAssociation.js
goldoraf: They will probably move in Association, yes.

PR #7, opened 2018-03-29
OAuth support

y-lohse:
*Please ignore the changes right now, i just made them to get the PR started.*

The goal of this PR is to add OAuth support to Cozy Client. I'll base my work in large parts on the [cozy-client-js](https://github.com/cozy/cozy-client-js/blob/master/src/auth_v3.js) version, and we should get feature parity. Maybe.

I haven't written any code yet, but I'd like some feedback on the general plan.

### Public API

Here are the things I'd like to keep from the public API — maybe renaming a few things in the process, we'll see:

- `registerClient`: well, registers a new OAuth client with the stack. Sends general information like client name, type, etc. This is required for the rest of the OAuth flow anyway
- `getClient` : gets the registered information from the stack for a given client. To be confirmed, but I think it gets different info whether or not you provide a token.
- `updateClient` : not sure if it's actually used by any app, but allows to update the registered information
- `unregisterClient` : duh, opposite of register. Used when we're revoking a client

So essentially, CRUDing for clients.

Also, purely OAuth things:

- `getAuthCodeURL` : gets the page that you send the user to
- `getAccessToken` : exchange a code for an access token
- `refreshToken` : exchange a refresh token for an access token

These 3 functions are wrapped in the more abstracted function `client.authorize`, and we'll probably have something similar.

### Integration

[`CozyStackClient`](https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/CozyStackClient.js#L66) wants a token to make requests. So eventually, that's what we want to provide.

I think we'll allow a new `OAuth` constructor parameter, and then probably expose the functions listed above as methods — that way we can re-use some of the cozy-client-js code.

I still need to look into the Storage thing — `cozy-client-js` stored some information in a `localStorage` for re-use, I want to understand exactly why.

Let me know if this sounds sensible to you.

Looping in @goldoraf @ptbrowne @jinroh

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -392,7 +392,7 @@ export default class CozyClient {
   /**
    * Performs a complete OAuth flow, including upating the internal token at the end.
    * @param   {function} openURLCallback Receives the URL to present to the user as a parameter, and should return a promise that resolves with the URL the user was redirected to after accepting the permissions.
-   * @returns {object}   Contains the fetched token and the client informations. These should be stored and used to restore the client.
+   * @returns {object}   Contains the fetched token and the client information. These should be stored and used to restore the client.
    */
   async startOAuthFlow(openURLCallback) {
     const client = this.getOrCreateStackClient()
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -394,7 +394,7 @@ export default class CozyClient {
    * @param   {function} openURLCallback Receives the URL to present to the user as a parameter, and should return a promise that resolves with the URL the user was redirected to after accepting the permissions.
    * @returns {object}   Contains the fetched token and the client informations. These should be stored and used to restore the client.
    */
-  async oauthFlow(openURLCallback) {
+  async startOAuthFlow(openURLCallback) {
     const client = this.getOrCreateStackClient()
     
     await client.register();
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -397,12 +397,12 @@ export default class CozyClient {
   async startOAuthFlow(openURLCallback) {
     const client = this.getOrCreateStackClient()
     
-    await client.register();
-    const state = client.generateStateCode()
-    const url = client.getAuthCodeURL(state, client.scope)
-    
+    await client.register()
+    const stateCode = client.generateStateCode()
+    const url = client.getAuthCodeURL(stateCode)
+
     const redirectedURL = await openURLCallback(url)
-    const code = client.getAccessCodeFromURL(redirectedURL, state)
+    const code = client.getAccessCodeFromURL(redirectedURL, stateCode)
     const token = await client.fetchAccessToken(code)
     
     client.setCredentials(token)
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -1,7 +1,7 @@
 import StackLink from './StackLink'
 import Association from './associations'
 import { QueryDefinition, Mutations } from './dsl'
-import CozyStackClient from 'cozy-stack-client'
+import CozyStackClient, { OAuthClient } from 'cozy-stack-client'
 import {
   default as reducer,
   createStore,
@@ -375,7 +375,7 @@ export default class CozyClient {
       associations
     }
   }
-
+  
   doctypeModelExists(doctype) {
     if (!this.schema) return false
     return (
@@ -387,6 +387,27 @@ export default class CozyClient {
 
   getDocumentFromStore(type, id) {
     return getDocumentFromStore(this.store.getState(), type, id)
+  }
+  
+  /**
+   * Performs a complete OAuth flow, including upating the internal token at the end.
+   * @param   {function} openURLCallback Receives the URL to present to the user as a parameter, and should return a promise that resolves with the URL the user was redirected to after accepting the permissions.
+   * @returns {object}   Contains the fetched token and the client informations. These should be stored and used to restore the client.
+   */
+  async oauthFlow(openURLCallback) {
+    const client = this.getOrCreateStackClient()
+    
+    await client.register();
+    const state = client.generateStateCode()
+    const url = client.getAuthCodeURL(state, client.scope)
+    
+    const redirectedURL = await openURLCallback(url)
+    const code = client.getAccessCodeFromURL(redirectedURL, state)
+    const token = await client.fetchAccessToken(code)
+    
+    client.setCredentials(token)
+    
+    return {token, infos: client.oauthOptions}
   }
 
   setStore(store) {
@@ -402,7 +423,7 @@ export default class CozyClient {
 
   getOrCreateStackClient() {
     if (!this.client) {
-      this.client = new CozyStackClient(this.options)
+      this.client = this.options.oauth ? new OAuthClient(this.options) : new CozyStackClient(this.options)
     }
     return this.client
   }
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -1,18 +1,11 @@
-import {
-  MutationTypes,
-  CozyLink,
-  getDoctypeFromOperation
-} from 'cozy-client'
+import { MutationTypes, CozyLink, getDoctypeFromOperation } from 'cozy-client'
 import PouchDB from 'pouchdb'
 import PouchDBFind from 'pouchdb-find'
 import mapValues from 'lodash/mapValues'
 import fromPairs from 'lodash/fromPairs'
 import omit from 'lodash/omit'
 import toPairs from 'lodash/toPairs'
-import {
-  getIndexNameFromFields,
-  getIndexFields
-} from './mango'
+import { getIndexNameFromFields, getIndexFields } from './mango'
 
 PouchDB.plugin(PouchDBFind)
 
@@ -50,23 +43,26 @@ const pouchResToJSONAPI = (res, isArray, doctype) => {
     }
   } else {
     return {
-      data: ensureHasBothIds(res),
+      data: ensureHasBothIds(res)
     }
   }
 }
 
 const createPouches = doctypes => {
-  return fromPairs(doctypes.map(
-    doctype => [doctype, {
-      db: new PouchDB(doctype)
-    }]
-  ))
+  return fromPairs(
+    doctypes.map(doctype => [
+      doctype,
+      {
+        db: new PouchDB(doctype)
+      }
+    ])
+  )
 }
 
 const sanitized = doc => omit(doc, '_type')
 
 const parseMutationResult = (original, res) => {
-  return {...original, ...omit(res, 'ok')}
+  return { ...original, ...omit(res, 'ok') }
 }
 
 const doNothing = () => {}
@@ -99,15 +95,15 @@ export default class PouchLink extends CozyLink {
     }
   }
 
-  getAllDBs () {
+  getAllDBs() {
     return Object.values(this.pouches).map(x => x.db)
   }
 
-  resetAllDBs () {
+  resetAllDBs() {
     return Promise.all(this.getAllDBs().map(db => db.destroy()))
   }
 
-  getDBInfo (doctype) {
+  getDBInfo(doctype) {
     const dbInfo = this.pouches[doctype]
     if (!dbInfo) {
       throw new Error(`${doctype} not supported by cozy-pouch-link instance`)
@@ -115,11 +111,11 @@ export default class PouchLink extends CozyLink {
     return dbInfo
   }
 
-  getDB (doctype) {
+  getDB(doctype) {
     return this.getDBInfo(doctype).db
   }
 
-  syncOne (doctype) {
+  syncOne(doctype) {
     return new Promise(resolve => {
       const info = this.getDBInfo(doctype)
       if (info.syncing) {
@@ -155,12 +151,12 @@ export default class PouchLink extends CozyLink {
     return (client.uri + '/data/' + doctype).replace('//', `//${basicAuth}`)
   }
 
-  supportsOperation (operation) {
+  supportsOperation(operation) {
     const impactedDoctype = getDoctypeFromOperation(operation)
     return !!this.pouches[impactedDoctype]
   }
 
-  request(operation, result=null, forward=doNothing) {
+  request(operation, result = null, forward = doNothing) {
     if (!this.synced) {
       return forward(operation)
     }
@@ -177,7 +173,7 @@ export default class PouchLink extends CozyLink {
     }
   }
 
-  hasIndex (name) {
+  hasIndex(name) {
     return Boolean(this.indexes[name])
   }
 
@@ -233,16 +229,16 @@ export default class PouchLink extends CozyLink {
     switch (mutation.mutationType) {
       case MutationTypes.CREATE_DOCUMENT:
         pouchRes = await this.createDocument(mutation)
-      break
+        break
       case MutationTypes.UPDATE_DOCUMENT:
         pouchRes = await this.updateDocument(mutation)
-      break
+        break
       case MutationTypes.DELETE_DOCUMENT:
         pouchRes = await this.deleteDocument(mutation)
-      break
+        break
       case MutationTypes.ADD_REFERENCES_TO:
         pouchRes = await this.addReferencesTo(mutation)
-      break
+        break
       case MutationTypes.UPLOAD_FILE:
         return forward(mutation, result)
       default:
@@ -251,7 +247,7 @@ export default class PouchLink extends CozyLink {
     return pouchResToJSONAPI(pouchRes, false, getDoctypeFromOperation(mutation))
   }
 
-  createDocument (mutation) {
+  createDocument(mutation) {
     return this.dbMethod('post', mutation)
   }
 
@@ -263,7 +259,7 @@ export default class PouchLink extends CozyLink {
     return this.dbMethod('remove', mutation)
   }
 
-  async dbMethod (method, mutation) {
+  async dbMethod(method, mutation) {
     const doctype = getDoctypeFromOperation(mutation)
     const { document } = mutation
     const db = this.getDB(doctype)
--- a/packages/cozy-stack-client/src/OAuthClient.js
+++ b/packages/cozy-stack-client/src/OAuthClient.js
@@ -1,0 +1,135 @@
+import CozyStackClient from './CozyStackClient'
+import AccessToken from './AccessToken'
+
+const defaultOAuthOptions = {
+  clientID: '',
+  clientName: 'a',
+  clientKind: '',
+  clientSecret: '',
+  clientURI: '',
+  registrationAccessToken: '',
+  redirectURI: 'http://localhost',
+  softwareID: 'a',
+  softwareVersion: '',
+  logoURI: '',
+  policyURI: '',
+  notificationPlatform: '',
+  notificationDeviceToken: '',
+}
+
+export default class OAuthClient extends CozyStackClient{
+  constructor({ oauthOptions, uri = '' }) {
+    super({token: null, uri})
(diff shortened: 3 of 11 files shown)


y-lohse:
[(2) Implémenter la partie auth mobile (OAuth)](https://trello.com/c/A00Uavk3/888-2-impl%C3%A9menter-la-partie-auth-mobile-oauth)

y-lohse:
I've implemented a first draft of the OAuth flow. Lots of finer points that I want to rework, but it's working. The API looks like this:

```
import { OAuthClient } from "cozy-stack-client";
// OAuthClient extends CozyStackClient, not sure if it will stay like that. But anyway right now it inherits all the jazz.

async function test(){
  const client = new OAuthClient({uri: 'http://cozy.tools:8080'}); // here we should also configure OAuth client infos
  // register the client with the stack
  await client.register();
  //construct the URL where we will send the user
  const state = client.generateStateCode()
  const url = client.getAuthCodeURL(state, ['io.cozy.contacts'])
  console.log(url)
  //at that point you must visit the URL and accept the permissions

  setTimeout(async () => {
    // then you paste the resulting URL in the following prompt. On cordova, this is done with the inAppBrowser
    const returnUrl = prompt('url?')
    // The new URL contains a code that we can exchange for a token
    const code = client.getAccessCodeFromURL(returnUrl, state)
    const token = await client.getAccessToken(code)
    client.setAccessToken(token)
  
    //and then we can just use the client as usual
    const contacts = client.fetch('POST', '/data/io.cozy.contacts/_all_docs', {}).then(console.log)
  }, 7000)
}
```

y-lohse:
I'm done with the low level functions! More things to do:

- add tests
- document everything
- probably add a high-level function that does the whole oauth flow
- detect revocations, maybe?
- have a clear-ish migration path for cozy drive and cozy banks

y-lohse:
Onto the last parts!

### High level function

Going through the whole OAuth process with the low level functions isn't too complicated, but not too interesting either. So i'm going to add an abstraction which is essentially going to be the replacement for the old [`oauthFlow`](https://github.com/cozy/cozy-client-js/blob/master/src/auth_v3.js#L259). 

### Restoring a client ?

In order to restart an OAuth client, we need 2 things:

- Some basic client informations, such as `client_id`, `client_secret`, `registration_access_token`, etc. These informations were stored by the apps, usually in the redux store.
- The previous token (which also contains a refresh token that can be used for renewal). This token was stored by cozy-client-js itself. More interestingly, this token is automatically restored when a new call to fetch is made — [`authorize()`](https://github.com/cozy/cozy-client-js/blob/master/src/fetch.js#L14) [returns a token for a webapp](https://github.com/cozy/cozy-client-js/blob/master/src/index.js#L252) but [calls `oauthFlow`](https://github.com/cozy/cozy-client-js/blob/master/src/index.js#L239) for oauth clients, [which loads the token](https://github.com/cozy/cozy-client-js/blob/master/src/auth_v3.js#L284). Our apps rely on this behavior

I'm not sure what I should do about this yet. Everything works without it, but it forces the dev to manually save and restore both informations before using cozy-client. Should cozy-client take care of that, or should it be the apps responsibility ?

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+    this.oauthOptions = this.camelCaseOAuthData(result)
+
+    return this.oauthOptions
+  }
+
+  /**
+   * Generates a random state code to be used during the OAuth process
+   * @returns {string}
+   */
+  generateStateCode() {
```
y-lohse: Replacing @ptbrowne comment from [here](https://github.com/cozy/cozy-client/pull/18#discussion_r179415626)

> It is a bit confusing to see state here. generateNone ? > https://fr.wikipedia.org/wiki/Nonce_(cryptographie) 

I found the name very confusing too, but it's called `state` on the stack as well](https://github.com/cozy/cozy-stack/blob/master/docs/auth.md#get-authauthorize) so I think it's better to keep the same name. And I've called it `stateCode` everywhere, instead of simply `state`.

ptbrowne: If it's confusing, let's change it also in the stack ?

y-lohse: Yeah that could be a thing. @nono or @jinroh do you see any reason not to do it (renaming `state` to `nonce` in the oauth process) ?

nono: I see a good reason to not do that: [the RFC about OAuth 2.0](https://tools.ietf.org/html/rfc6749#section-4.1.1) says that this parameter is named `state`

packages/cozy-client/src/CozyClient.js
```diff
@@ -402,7 +423,7 @@ export default class CozyClient {
 
   getOrCreateStackClient() {
     if (!this.client) {
-      this.client = new CozyStackClient(this.options)
+      this.client = this.options.oauth ? new OAuthClient(this.options) : new CozyStackClient(this.options)
```
enguerran: I see that `OAuthClient` and `CozyStackClient` return the same concept and take the same parameters. Could it be abstracted in this way?

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -1,18 +1,11 @@
-import {
-  MutationTypes,
-  CozyLink,
-  getDoctypeFromOperation
-} from 'cozy-client'
+import { MutationTypes, CozyLink, getDoctypeFromOperation } from 'cozy-client'
```
enguerran: Some may have different prettier/formatting rules. Could you check it's not you?

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+
+    // special case: turn redirect_uris into an array
+    if (
+      result['redirect_uris'] &&
+      result['redirect_uris'] instanceof Array === false
+    )
+      result['redirect_uris'] = [result['redirect_uris']]
+
+    return result
+  }
```
enguerran: I am a bit impressed by this function. You put in inside a class not really related and the function's behavior own too much responsibilities: keep the full list of properties, build a binding object, transform a non array property into an array.

We could use regexp to transform property name, see an example here https://jsfiddle.net/LkmaLgsn/3/

And get the array transformation out of this function as it's not express by the name of the function.

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+    const result = {}
+
+    Object.keys(data).forEach(fieldName => {
+      const key = mappedFields[fieldName] || fieldName
+      const value = data[fieldName]
+      result[key] = value
+    })
+
+    return result
+  }
```
enguerran: Yet another function that could be written with a simple and more straight forward regexp. See above.

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+        softwareID: this.oauthOptions.softwareID,
+        clientKind: this.oauthOptions.clientKind,
+        clientURI: this.oauthOptions.clientURI,
+        logoURI: this.oauthOptions.logoURI,
+        policyURI: this.oauthOptions.policyURI,
+        softwareVersion: this.oauthOptions.softwareVersion,
+        notificationPlatform: this.oauthOptions.notificationPlatform,
+        notificationDeviceToken: this.oauthOptions.notificationDeviceToken
+      })
+    )
```
enguerran: I think that this is too much complicated. What if a property is missing? Can't you loop over every object's properties? I don't trust hard-coded exhaustive list.

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+
+  /**
+   * Registers the currenly configured client with the OAuth server. 
+   * @throws {Error} When the client is already registered
+   * @returns {promise} A promise that resolves with a complete list of client information, including client ID and client secret.
+   */
+  async register() {
+    if (this.isRegistered()) throw new Error('Client already registered')
+
+    const data = await this.fetch(
```
enguerran: You don't catch fetch exception here.

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+      }
+    )
+  }
+
+  /**
+   * Fetches the complete set of client informations from the server after it has been registered.
+   * @throws {NotRegisteredException} When the client doesn't have it's registration information
+   * @returns {promise}
+   */
+  async fetchInformations() {
```
enguerran: _"information"_ in english may always be written in a plural form and don't take a final "s".

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+      softwareID: this.oauthOptions.softwareID
+    }
+    const data = this.snakeCaseOAuthData({
+      ...mandatoryFields,
+      ...informations
+    })
+
+    if (resetSecret) data['client_secret'] = this.oauthOptions.clientSecret
+
+    const result = await this.fetch(
```
enguerran: You don't catch possible exception here.

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+      for (let i = 0; i < buffer.length; i++) {
+        buffer[i] = Math.floor(Math.random() * 255)
+      }
+    }
+
+    return btoa(String.fromCharCode.apply(null, buffer))
+      .replace(/=+$/, '')
+      .replace(/\//g, '_')
+      .replace(/\+/g, '-')
+  }
```
enguerran: It's not very easy to see what this code do. Is it ok if it returns `0123456789abcdef`? What is a sample? Is it tested somewhere?

enguerran: Another confusion comes from the owner of this function: how is it related to this class? It may be written in its own file or in a private area of the file of the class.

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+    const data = new URLSearchParams({
+      grant_type: 'authorization_code',
+      code: accessCode,
+      client_id: this.oauthOptions.clientID,
+      client_secret: this.oauthOptions.clientSecret
+    })
+
+    const result = await this.fetch('POST', '/auth/access_token', data, {
+      headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
+    })
```
enguerran: You don't catch possible exception here.

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+    const data = new URLSearchParams({
+      grant_type: 'refresh_token',
+      refresh_token: this.token.refreshToken,
+      client_id: this.oauthOptions.clientID,
+      client_secret: this.oauthOptions.clientSecret
+    })
+
+    const result = await this.fetch('POST', '/auth/access_token', data, {
+      headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
+    })
```
enguerran: You don't catch possible exception here.

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+      refresh_token: this.token.refreshToken,
+      ...result
+    })
+  }
+
+  /**
+   * Updates the client's stored token
+   * @param {string} token = null The new token to use — can be a string, a json object or an AccessToken instance.
+   */
+  setCredentials(token = null) {
```
enguerran: I checked [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters), as `undefined` or `null` are falsy value, then calling `setCredentials(null)` or `setCredentials(undefined)` will go in the `else` statement, even if there is no default parameter.

> Default function parameters allow formal parameters to be initialized with default values if no value or undefined is passed.

```js
function test(num = 1) {
  console.log(typeof num);
}

test();          // 'number' (num is set to 1)
test(undefined); // 'number' (num is set to 1 too)

// test with other falsy values:
test('');        // 'string' (num is set to '')
test(null);      // 'object' (num is set to null)
```

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -0,0 +1,358 @@
...
+  }
+}
+
+class NotRegisteredException extends Error {
+  constructor(message = 'Client not registered or missing OAuth informations') {
+    super(message)
+    this.message = message
+    this.name = 'NotRegisteredException'
+  }
+}
```
enguerran: I know there's some weird stuff with classes extending the Error class. See https://stackoverflow.com/questions/31089801/extending-error-in-javascript-with-es6-syntax-babel

packages/cozy-stack-client/src/__tests__/__snapshots__/OAuthClient.spec.js.snap
```diff
@@ -0,0 +1,159 @@
...
+
+exports[`OAuthClient without registration should throw on other server interactions 1`] = `"Client not registered or missing OAuth informations"`;
+
+exports[`OAuthClient without registration should throw on other server interactions 2`] = `"Client not registered or missing OAuth informations"`;
+
+exports[`OAuthClient without registration should throw on other server interactions 3`] = `"Client not registered or missing OAuth informations"`;
+
+exports[`OAuthClient without registration should throw on other server interactions 4`] = `"Client not registered or missing OAuth informations"`;
+
+exports[`OAuthClient without registration should throw on other server interactions 5`] = `"Client not registered or missing OAuth informations"`;
```
enguerran: Do you check the content of snapshots? I can't review it, but I think it is a good idea to review it carefuly. So if you could throw a double attention here as we will be ablt to rely on those snapshots later.

packages/cozy-client/src/CozyClient.js
```diff
@@ -370,6 +370,27 @@ export default class CozyClient {
...
+   * Performs a complete OAuth flow, including upating the internal token at the end.
+   * @param   {function} openURLCallback Receives the URL to present to the user as a parameter, and should return a promise that resolves with the URL the user was redirected to after accepting the permissions.
+   * @returns {object}   Contains the fetched token and the client informations. These should be stored and used to restore the client.
+   */
+  async oauthFlow(openURLCallback) {
+    const client = this.getOrCreateStackClient()
+    
+    await client.register();
+    const state = client.generateStateCode()
+    const url = client.getAuthCodeURL(state, client.scope)
```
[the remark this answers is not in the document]
y-lohse: Ah yes good catch, I added the property quite late and didn't properly review this use.

packages/cozy-client/src/CozyClient.js
y-lohse: Not sure what you mean by that ? `OAuthClient` is `CozyStackClient` + OAuth. But keeping them separated reduces their individual complexity, and it's transparent for the user.

packages/cozy-stack-client/src/OAuthClient.js
y-lohse: Maybe the functions are misnamed? What they really do is transform data in a format suitable for the server., and the other way around.

The regexp approach wouldn't work for names like `clientURI`, or the regexp would become quite complex, whereas the mapping here is very explicit.

packages/cozy-stack-client/src/OAuthClient.js
y-lohse: We do want an [exhaustive list though](https://github.com/cozy/cozy-stack/blob/master/docs/auth.md#post-authregister). We *could* just send the whole object because the server probably ignores the ones it doesn't know, but I don't think that's better.
And if a property is missing, it's totally ok.

packages/cozy-stack-client/src/OAuthClient.js
y-lohse: Uhm no, why do you think we should? I think it's desirable that an exception gets thrown if the `fetch` fails.

packages/cozy-client/src/CozyClient.js
enguerran: I mean, could we have an unique class that implements what it has to

```
const options = {
  oauth: {}
}
const client = new CozyClient(options)
```

Maybe I misunderstood what you wanted to do here.

packages/cozy-stack-client/src/OAuthClient.js
y-lohse: > Another confusion comes from the owner of this function: how is it related to this class? It may be written in its own file or in a private area of the file of the class.

You're right, it could be a standalone function. You would always use it in conjunction with the rest of the class though, so you'd have to import it along with the class everywhere. 
I don't mind moving it, but it would make the DX worse. Do you think it's worth sacrificing a bit of DX for the clarity it will give us?

Otherwise, it just… generates a random string. `0123456789abcdef` is cool. It has a test attached to it, and it's documented. Do you think a sample return value in the doc would help?

packages/cozy-stack-client/src/OAuthClient.js
y-lohse: Interesting, although in that case everything seems to work fine. The error messages are in the snapshots for example :)

packages/cozy-stack-client/src/__tests__/__snapshots__/OAuthClient.spec.js.snap
y-lohse: Yes :) I checked them as I was writing the tests and actually caught a few problems with them.

packages/cozy-stack-client/src/OAuthClient.js
enguerran: Here are my thoughts:
- this function is too big.
- this function has too much responsibilities.
- this function should not be public.

The regexp proposal was a proposition to help, I have no clue if it works or not, but it should as this function just transform a string into another string.

I can imagine something like a usage of this 3 one-lined functions:
```js
const toSnakeCase = prop =>
  prop.replace(/([A-Z]+)/g, "_$1").toLowerCase();

const redirect_uri_correction = prop =>
  /redirect_uri/.test(prop) ? `${prop}s` : prop;

const toArray = val => val instanceof Array ? val : [val]
```

packages/cozy-client/src/CozyClient.js
goldoraf: Yeah, I think it'd be better to let `CozyStackClient` instantiate its OAuth utility if there is an `oauth` option. But it's a simple change that can be done later.

packages/cozy-stack-client/src/OAuthClient.js
enguerran: I mean, if the protocol changes and needs a new property, then the developer will add that property to their call, and the library won't take that new property into account without a change in the library code.

packages/cozy-pouch-link/src/CozyPouchLink.js
goldoraf: Patrick didn't format its code correctly using prettier, and as I don't have added a git hook yet... So, no it's not @y-lohse fault ;)

packages/cozy-client/src/CozyClient.js
y-lohse: I'm afraid it will lead to a lot of if/else-ing inside `CozyStackClient`, but we'll see.

packages/cozy-stack-client/src/OAuthClient.js
y-lohse: Yes ok, it will reject the promise with the exception. But still, that's what we want, isn't it?

packages/cozy-stack-client/src/OAuthClient.js
enguerran: If you don't need to export it, you could put it outside the class in the same file.

Yes, an example of returned valued may help.

y-lohse:
I'll go ahead and merge this now. 
I think the main remaining issue is the snake-camel-case functions, I'll try to come up with a better approach in a separate PR. The test coverage is quite exhaustive so we should be fine.

PR #8, opened 2018-03-30
chore: update yarn.lock

y-lohse:
When installing the dependencies, i get the error `An unexpected error occurred: "https://registry.yarnpkg.com/jest/-/jest-23.0.0-alpha.0.tgz: Request failed \"404 Not Found\"".`. I guess the package was removed or something?

Anyway, new lock file, and git-ignored the yarn error file.

I also suggest to pin dependencies and to add the renovate app to avoid issues like that :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ packages/*/yarn-error.log
 lerna-debug.log
 npm-debug.log
 coverage/*
+yarn-error.log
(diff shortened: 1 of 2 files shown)


goldoraf:
Sorry for not looking at this PR sooner :/ I had the same problem, and already commited a new yarn.lock 
Thanks anyway!

PR #9, opened 2018-03-30
chore: don't lint before tests

y-lohse:
Just a suggestion, but I don't think we should lint files before running tests. 
I assume we all sometimes write some crap code while writing tests, and it's really annoying if the test don't even run because the temporary code has linting errors.


goldoraf:
I fully agree with you, would you mind to rebase this PR and merge it? Thanks!

PR #18, opened 2018-04-05
Auto refresh

y-lohse:
This si based on my work in #7 , the only new commit to look at is [this one](https://github.com/cozy/cozy-client/commit/fd4779bb5e33b3a85159f15daf3b6337f41be880).

So essentially, when using the OAuth client and the token has expired, we transparently renew it. It works and I think it's a good first step. 

But for transparency, there are 2 things I don't like with this:

1. This function silently changes the internal state of the class by updating the token.
2. If the app has stored the token (in local storage, typically), the stored token won't be updated. It will still work fine, but every time the client is restored, the first request will trigger a refresh


y-lohse:
[(2) Implémenter la partie auth mobile (OAuth)](https://trello.com/c/A00Uavk3/888-2-impl%C3%A9menter-la-partie-auth-mobile-oauth)

PR #19, opened 2018-04-11
Improvements for banks

goldoraf:
This PR adds a (temporary) special has-many association for Banks plus minor fine-tunings around OAuth.

@ptbrowne I also added an `autoHydrate` option, so that we can disable it in banks for the time being.


ptbrowne:
I did not take the time to review but I think it would be good to have tests for those functionalities

PR #21, opened 2018-04-24
Add a classic fetch in CozyStackClient

CPatchane:
Add a very low level fetch enhanced with the Cozy token and the Cozy URL and use it in the fetchJSON method.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -40,6 +40,10 @@ export default class CozyClient {
     return this.getOrCreateStackClient().collection(doctype)
   }
 
+  fetch(method, path, body, options = {}) {
+    return this.getOrCreateStackClient().fetch(method, path, body, options)
+  }
+
   all(doctype) {
     return new QueryDefinition({ doctype })
   }
--- a/packages/cozy-stack-client/src/AppCollection.js
+++ b/packages/cozy-stack-client/src/AppCollection.js
@@ -22,7 +22,7 @@ export default class AppCollection {
    */
   async all() {
     const path = uri`/apps/`
-    const resp = await this.client.fetch('GET', path)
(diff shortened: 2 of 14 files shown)


CPatchane:
We will have to use now `fetchJSON` for JSON behaviour instead of `fetch` so we may consider this change as a breaking change.

gregorylegarec:
@CPatchane @goldoraf If we plan to externalize the fetch methods to make them available in other lib (Yes I am thinking to cozy-interapp), shouldn't this method be outside CozyStackClient ? I don't feel that we need to ship the whole lib as we will be only interested in two methods.

CPatchane:
>  the actual fetch method is used in a lot of places, and merging right now without changing these calls to fetchJSON would break the client.

I think that at this early stage of development we shouldn't be afraid about breaking changes particularly if it's about concern that matter like here (we hide the fact that `fetch` is only about JSON and text, it can be confusing comparing to the browser `fetch` API spec).

But I ok about doing JSON by default (a `json` option `true` by default?) since it's the majority of usages in applications, it's a good point.
And completely agree about getting a very lightweight client with the fetch method, I am looking forward to see this code so :)

nono:
Any update on this PR?

ptbrowne:
> I would like to have that fetch method on the cozy-client level, so that you just have to instantiate the higher level client instead of using the lower level cozy-stack-client that should die basically. I have begun to work on this, I'll try to push a PR asap. Not sure I'm clear here, I hope it'll get clearer with some code to show.

@goldoraf any updates on this ? I agree that it would be easier from a user perspective not to know that that fetchJSON is using the stack link.

gregorylegarec:
@CPatchane @ptbrowne @goldoraf It should be ok now. I exposed `client.fetch()`, but not `client.fetchJSON()` for now.

CPatchane:
Great! Thanks a lot @gregorylegarec 🎉

PR #22, opened 2018-04-30
Fixed fetchMore on has-many-files association

goldoraf:


goldoraf:
[On ne peut afficher que les 50 premières photos d'un album](https://trello.com/c/rk7hhJQk/947-on-ne-peut-afficher-que-les-50-premi%C3%A8res-photos-dun-album)

packages/cozy-client/src/CozyClient.js
```diff
@@ -119,15 +119,15 @@ export default class CozyClient {
     ]
   }
 
-  static registerHook (doctype, name, fn) {
+  static registerHook(doctype, name, fn) {
     CozyClient.hooks = CozyClient.hooks || {}
-    const hooks = CozyClient.hooks[doctype] = CozyClient.hooks[doctype] || {}
+    const hooks = (CozyClient.hooks[doctype] = CozyClient.hooks[doctype] || {})
```
enguerran: That looks like a very hard to find bug.

goldoraf: No, it's only a Prettier reformatting, I should add a hook to avoid this

packages/cozy-client/src/CozyClient.js
```diff
@@ -136,7 +136,10 @@ export default class CozyClient {
 
   async destroy(document, mutationOptions = {}) {
     await this.triggerHook('before:destroy', document)
-    const res = await this.mutate(Mutations.deleteDocument(document), mutationOptions)
+    const res = await this.mutate(
+      Mutations.deleteDocument(document),
+      mutationOptions
+    )
     await this.triggerHook('after:destroy', document)
```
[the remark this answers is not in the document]
goldoraf: Hmmm I don't see how to avoid it, hooks must be executed in that order (I don't like this hook impl, but that's another problem). Do you have any idea?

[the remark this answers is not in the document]
enguerran: Unfortunately I don't have any ideas to propose, I saw this code in the PR, but I don't know what this code does, or how the code is structured, so it's hard to propose a solution to avoid this side effect, which is still a smell difficult to maintain.

packages/cozy-client/src/CozyClient.js
enguerran: Chaining variable assignments is discouraged by airbnb/javascript code style: https://github.com/airbnb/javascript#variables--no-chain-assignment

PR #25, opened 2018-05-26
Set the linter + run prettier + set babel preset

CPatchane:
- [x] : Set babel-preset-cozy-app
- [x] : Set eslint-config-cozy-app + add rules in webpack config
- [x] : Run prettier on every js/jsx files
- [x] : Fix all eslint issues

I fixed a lot of lint __issues about unused or undefined variables__ so please __check them again__.
There are __still one lint issue__, it's __about the [`updateDocument`](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/documents.js#L18) function which is never used__, so I kept it like that for now, I think I need more context.
Also some tests failed, but its exactly the same on the `master` branch, we should enable TravisCI on this project when the tests and linting will be good but that's another subject.


goldoraf:
Thanks a lot for this, it's greatly appreciated :heart: ! I'll take the time to review and fix the remaining issues next week :+1:

packages/cozy-client/src/CozyClient.js
```diff
@@ -53,6 +52,7 @@ export default class CozyClient {
     return new QueryDefinition({ doctype, id })
   }
 
+  // eslint-disable-next-line no-unused-vars
```
goldoraf: This disabling comment annoys me a bit, because I like that way of taking the fields that interest me and dismissing the rest using the spread operator: `{ _type, ...attributes }`. I suppose I could something like `{ _type, ..._ }` and remove the comment, but the code would loose a bit of readability. Do you see another way to fix the problem?

packages/cozy-client/src/CozyLink.js
```diff
@@ -5,6 +5,7 @@ export default class CozyLink {
     }
   }
 
+  // eslint-disable-next-line no-unused-vars
```
goldoraf: Arg! Just another reason to move to Typescript I guess... :smile:

CPatchane: Here I just put that because I guess the method is not implemented yet. If it will stay in this state we can remove all parameters here.

.babelrc
```diff
@@ -1,4 +1,3 @@
 {
-  "presets": ["env", "react"],
-  "plugins": ["transform-object-rest-spread", "transform-class-properties"]
-}
\ No newline at end of file
+  "presets": [["cozy-app", { "node": true }], "react"]
```
CPatchane: I had an issue using the app version with the `babel-plugin-transform-runtime` breaking the CSP.
Let's use the node preset for now until I figured out on my side.

packages/cozy-client/src/CozyClient.js
CPatchane: For here I think we can remove it because you then do: `const document = { _type: type, ...attributes }` to replace `_type`.
For some other cases I think we could just take the parameters and then doing like here `{toReplace: replacement, ...attributes}` or use `Object.assign` to keep just properties we want in a new Object. I know it's not ideal but it's also more readable IMHO.

nono:
@goldoraf @CPatchane ping, don't forget this PR!

.babelrc
gregorylegarec: Any news about this point ?

CPatchane: Nope :/, I completely forgot this point

PR #26, opened 2018-05-30
Support cozy-to-cozy files sharing

goldoraf:
This PR aims to support basic sharing features in Cozy apps. As the backend implementation is not complete right now (only files are fully supported for instance), this work shouldn't be considered as complete: there is still work to do, but we'll have to wait for the API to stabilize.

Regarding the changes:
 - initially, I chose to put sharing by link code into the `DocumentCollection`, mainly because sharing by link uses some special kind of permissions documents. After discussing the matter with @y-lohse, we decided it was better to implement this in a specific `PermissionCollection`.
 - sharing documents are some kind of a special beast too, therefore they needed their own `SharingCollection`.
 - I added some tests, but clearly we need more.

### DIFF
--- a/packages/cozy-stack-client/src/SharingCollection.js
+++ b/packages/cozy-stack-client/src/SharingCollection.js
@@ -1,0 +1,96 @@
+import DocumentCollection, { normalizeDoc } from './DocumentCollection'
+import { isFile, isDirectory } from './FileCollection'
+import { uri } from './utils'
+
+const normalizeSharing = sharing => normalizeDoc(sharing, 'io.cozy.sharings')
+
+export default class SharingCollection extends DocumentCollection {
+  async findByDoctype(doctype) {
+    const resp = await this.client.fetch(
+      'GET',
+      uri`/sharings/doctype/${doctype}`
+    )
+    return {
+      ...resp,
+      data: resp.data.map(normalizeSharing)
+    }
+  }
+
+  async share(document, recipients, sharingType, description) {
+    const resp = await this.client.fetch('POST', '/sharings/', {
+      data: {
+        type: 'io.cozy.sharings',
+        attributes: {
+          description,
+          rules: getSharingRules(document, sharingType)
+        },
+        relationships: {
+          recipients: {
+            data: recipients.map(({ _id, _type }) => ({ id: _id, type: _type }))
+          }
+        }
+      }
+    })
+    return { data: normalizeSharing(resp.data) }
+  }
+
+  revokeRecipient(sharing, recipientEmail) {
+    const memberIndex = sharing.attributes.members.findIndex(
+      m => m.email === recipientEmail
+    )
+    return this.client.fetch(
+      'DELETE',
+      uri`/sharings/${sharing._id}/recipients/${memberIndex}`
+    )
+  }
+
+  revokeSelf(sharing) {
+    return this.client.fetch(
+      'DELETE',
+      uri`/sharings/${sharing._id}/recipients/self`
+    )
+  }
+}
+
+// Rules determine the behavior of the sharing when changes are made to the shared document
+// See https://github.com/cozy/cozy-stack/blob/master/docs/sharing-design.md#description-of-a-sharing
+const getSharingRules = (document, sharingType) => {
+  const { _id, _type } = document
+  return isFile(document)
+    ? [
+        {
+          title: document.name,
+          doctype: 'io.cozy.files',
+          values: [_id],
+          ...getSharingPolicy(document, sharingType)
+        }
+      ]
+    : [
+        {
+          title: 'collection',
+          doctype: _type,
+          values: [_id],
+          ...getSharingPolicy(document, sharingType)
+        },
+        {
+          title: 'items',
+          doctype: 'io.cozy.files',
+          values: [`${_type}/${_id}`],
+          selector: 'referenced_by',
+          ...(sharingType === 'two-way'
+            ? { add: 'sync', update: 'sync', remove: 'sync' }
+            : { add: 'push', update: 'none', remove: 'push' })
+        }
+      ]
+}
+
+const getSharingPolicy = (document, sharingType) => {
+  if (isFile(document) && isDirectory(document)) {
+    return sharingType === 'two-way'
+      ? { add: 'sync', update: 'sync', remove: 'sync' }
+      : { add: 'none', update: 'none', remove: 'revoke' }
+  }
+  return sharingType === 'two-way'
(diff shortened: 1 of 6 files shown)


packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -0,0 +1,99 @@
...
+
+export default class SharingCollection extends DocumentCollection {
+  async findByDoctype(doctype) {
+    const resp = await this.client.fetch(
+      'GET',
+      uri`/sharings/doctype/${doctype}`
+    )
+    return {
+      ...resp,
+      data: resp.data.map(s => normalizeSharing(s))
```
enguerran:
```js
resp.data.map(s => normalizeSharing(s))
```
should be simplified by
```js
resp.data.map(normalizeSharing)
```

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -0,0 +1,99 @@
...
+      'DELETE',
+      uri`/sharings/${sharing._id}/recipients/self`
+    )
+  }
+}
+
+const isFile = ({ _type, type }) =>
+  _type === 'io.cozy.files' || type === 'directory' || type === 'file'
+
+const isDirectory = ({ type }) => type === 'directory'
```
enguerran: not sure `isDirectory` and `isFile` are specific to `SharingCollection`, may be placed in another file.

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -0,0 +1,99 @@
...
+
+const getBehaviorsFor = (document, sharingType) => {
```
enguerran: What does _behavior_ mean in the context of a sharing?

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -0,0 +1,99 @@
...
+  if (isFile(document) && isDirectory(document)) {
```
enguerran: Why do you check that it is both a file and a directory, even though the check is whether `type ==='directory'`?

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -0,0 +1,99 @@
...
+          title: 'items',
+          doctype: 'io.cozy.files',
+          values: [`${_type}/${_id}`],
+          selector: 'referenced_by',
+          ...(sharingType === 'two-way'
+            ? { add: 'sync', update: 'sync', remove: 'sync' }
+            : { add: 'push', update: 'none', remove: 'push' })
+        }
+      ]
+}
```
enguerran: You may refactor (divide the code into smaller functions) to enhance readability and so maintainance.

enguerran:
You divided your works into several commits, this is a very good practice. There are still a few tips I can share.
The comment for the commit https://github.com/cozy/cozy-client/pull/26/commits/0217dee546abd8203d09ae06ed8e0e362c4ab58b is too vague. Of two things one: either the commit adds something interesting and the message must reflect it, or it is a commit that amends another commit and in this case it must be amended to this previous commit.

enguerran:
I cannot understand how f8d9c1dfef8ef9a5ad851b951648244b40458846 is useful. There is a function `statById(id)` that does the same. If you want a more functional function name (`get` instead of obscure `statById`), then you must implement the function `get` and transform the function `statById` to call this `get` (with a warning of future depreciation).
For the moment we have two entries for the same thing, soon these two functions will diverge and the developers will not know which of the two functions they have to use for their own need.

packages/cozy-stack-client/src/SharingCollection.js
goldoraf: Well, with the previous sharing API, you had to define a bunch of permissions to characterize the sharing (RO/RW). Now, we have to define `rules` with properties that sets the "permissions", something like `{ add: 'sync', update: 'sync', remove: 'sync' }`. I called these properties a **behavior**, but I'd like to find a better name. Any idea?

packages/cozy-stack-client/src/SharingCollection.js
goldoraf: I agree, but I prefer to wait for the backend to implement sharing of other doctypes properly, because I don't know yet how this code will evolve.

goldoraf:
@enguerran: I'm sorry, I should have added a [WIP] label, this PR wasn't really ready for review :/

About the commits set: the commits are not atomic enough, and sometimes badly commented, I know that. To my defense, this work took place in a rather stretched period of time with a lot of interruptions, and I sometimes forgot what I was doing previously and didn't commit. 

About the `get` function: I wanted to have a `get` in order to mirror other collections. The simplest thing to do was to call the existing `statById`, but `statById` returns much more than I'd want to return in this `get` method. Not sure how to handle that right now, but thanks for your comment, I'll keep that in mind.

packages/cozy-stack-client/src/SharingCollection.js
enguerran: Let's do a little brainstorming here:

the configuration object for the sharing describes some __strategies__: synchronize, revoke, etc. At least that's an __aspect__ of the sharing.

So I vote for `getSharginAspect` or `getStrategies`. But I do not have much hindsight for sharing and its lexical field. Maybe `getBehaviors` is strong enough to express what this function does.

goldoraf: The stack doc talks about behavior here: https://docs.cozy.io/cozy-stack/sharing-design.html, but I guess they have the same lexical problem. Aspect looks too generic to me, strategy is better but not perfect.

ptbrowne: Policy ?

y-lohse: I found this confusing too, may I suggest adding a comment pointing to the doc mentioned above, or briefly describing what a *behavior* is here?

y-lohse:
Regarding `get` vs `statById `, `statById` never made much sense to me, so maybe we could get rid of that one instead?

packages/cozy-stack-client/src/SharingCollection.js
goldoraf: policy is great, thanks @ptbrowne!

goldoraf:
@enguerran We needed to move forward regarding to the sharing implementation, so we decided to merge this PR "as-is" (lumière). As our work on sharing is far from complete, we'll have opportunities to enhance that code, so feel free to review this PR when you have time, we'll take into account your feedback in future PRs. Thanks a lot!

PR #30, opened 2018-06-04
Let's make things green 🌴

y-lohse:
There are a few failing tests, but I don't know if there's a regression behind them or not. I'll link to the travis with more details log once it's created.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,14 @@ node_js:
 matrix:
   fast_finish: true
 cache:
-  yarn: true
   directories:
   - node_modules
+before_install:
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.7.0
+  - export PATH=$HOME/.yarn/bin:$PATH
+  - yarn --version
+before_script:
+  - lerna run build
 script:
 - yarn lint
 - yarn test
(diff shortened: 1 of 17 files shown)


y-lohse:
Here we go:

- [A problem with CouchLink](https://travis-ci.org/cozy/cozy-client/builds/387825162#L482) — I don't even have the same locally. Probably something for @ptbrowne ?
- [A new throw](https://travis-ci.org/cozy/cozy-client/builds/387825162#L516) when a document has no `_type`. [This is the offending line](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/__tests__/ObservableQuery.spec.js#L33), I suspect the result of `receiveQueryResult ` has changed? Probably for @goldoraf 
- [A `queryId` has changed](https://travis-ci.org/cozy/cozy-client/builds/387825162#L545) — it used to be a name, now it's a number. Again, not sure if it's a bug or not.

goldoraf:
@y-lohse I fixed the remaining failing tests, including unhandled promises rejections that didn't make tests fail, but were annoying and smelly. It was only a matter of mocking the fetch with a (dummy) response instead of no response: https://github.com/cozy/cozy-client/pull/30/files#diff-b3071f3043c1b750abded70f159b3d75R30

Thanks a lot for starting this!

PR #31, opened 2018-06-07
Added lib to detect file mimetypes from the extension

goldoraf:
When trying to upload a directory in Firefox, there is a bug that prevents to get the directory files' mimetypes. Consequently, it's necessary to try to guess the mimetype from the extension so that files are not uploaded with a `application/octet-stream` mimetype. I chose to use a rather simple and light lib to do that.


goldoraf:
[(2) Transfert de fichiers & et de dossiers par drag and drop](https://trello.com/c/FQm5RH0Y/956-2-transfert-de-fichiers-et-de-dossiers-par-drag-and-drop)

PR #36, opened 2018-06-19
Fix Travis

goldoraf:


gregorylegarec:
We should keep ESLint on Cozy Client, shouldn't we ?

If a rule is really blocking integration we may maybe disable it instead ?

What do you think ?

goldoraf:
@gregorylegarec don't worry, I didn't completely removed it, but its run during webpack build made Travis fail, no idea why yet. I investigate.

PR #40, opened 2018-07-05
Add package name in package.json

gregorylegarec:
This makes easier to `yarn link`

Also, package.json was indented with tabs, so this PR also:
* updates .editorconfig to match cozy-collect one
* replaces tabs with spaces in package.json

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,2 +1,24 @@
-[*.js]
-indent_size=2
+# editorconfig.org
+# editorconfig is a unified configuration file that all editors can take
+# into account
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+
+[*.coffee]
+indent_size = 4
+
+[*.jade]
+trim_trailing_whitespace = false
+
+[*.pug]
+trim_trailing_whitespace = false
+
+[*.styl]
+indent_size = 4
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -1,2 +1,24 @@
...
+[*]
+charset = utf-8
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+
+[*.coffee]
+indent_size = 4
```
enguerran: That's a good idea to have a more general configuration (that's configure the editor to manage `.json`).
Very great work!
But why do you copy/paste some irrelevent `.editorconfig` instead of using a simple more accurate one?

package.json
```diff
@@ -1,47 +1,48 @@
...
-	"jest": {
-		"setupFiles": [
-			"<rootDir>/packages/cozy-client/src/__tests__/setup.js"
-		],
-		"setupTestFrameworkScriptFile": "<rootDir>/packages/cozy-stack-client/src/__tests__/setup.js",
-		"testMatch": [
-			"**/__tests__/**/(*.)(spec|test).js?(x)"
-		]
-	}
+  "name": "cozy-client",
```
enguerran: This name is used in `./packages/cozy-client/package.json` file: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/package.json#L2

ptbrowne:
This PR is without activity since 2 weeks. I am closing. Feel free to reopen @gregorylegarec if need be.

PR #41, opened 2018-07-06
Add AppCollection

gregorylegarec:
As apps are retrieved with a specific endpoint, they need a custom implementation into an AppCollection class.

At this time, the only method needed is `all()`

### DIFF
--- a/packages/cozy-stack-client/src/AppCollection.js
+++ b/packages/cozy-stack-client/src/AppCollection.js
@@ -1,0 +1,33 @@
+import { uri } from './utils'
+import { normalizeDoc } from './DocumentCollection'
+
+export const APP_DOCTYPE = 'io.cozy.apps'
+
+/**
+ * Abstracts a collection of documents of the same doctype, providing
+ * CRUD methods and other helpers.
+ */
+export default class AppCollection {
+  constructor(client) {
+    this.client = client
+  }
+
+  /**
+   * Lists all documents of the collection, without filters.
+   *
+   * The returned documents are not paginated by the stack.
+   *
+   * @return {{data, meta, skip, next}} The JSON API conformant response.
+   * @throws {FetchError}
+   */
+  async all() {
+    const path = uri`/apps/`
+    const resp = await this.client.fetch('GET', path)
+    return {
+      data: resp.rows.map(row => normalizeDoc(row.doc, APP_DOCTYPE)),
+      meta: { count: resp.total_rows },
+      skip: 0,
+      next: false
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/AppCollection.js
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Lists all documents of the collection, without filters.
+   *
+   * The returned documents are paginated by the stack.
+   *
+   * @param  {{limit, skip}} options The pagination options.
+   * @return {{data, meta, skip, next}} The JSON API conformant response.
+   * @throws {FetchError}
+   */
+  async all(options = {}) {
```
enguerran: It seems that this parameter is never used.

enguerran:
Oh, I just see an extra __"s"__ in the commit message, use the infinitive form.

packages/cozy-stack-client/src/AppCollection.js
```diff
@@ -0,0 +1,36 @@
+import { uri } from './utils'
+import { normalizeDoc } from './DocumentCollection'
+
+/**
+ * Abstracts a collection of documents of the same doctype, providing
+ * CRUD methods and other helpers.
+ */
+export default class AppCollection {
+  constructor(doctype, client) {
+    this.doctype = doctype
```
enguerran: The doctype for such a collection is always `io.cozy.apps`, it must be simplified with a optional and default valued parameter or without the parameter `doctype` at all.

packages/cozy-stack-client/src/AppCollection.js
```diff
@@ -0,0 +1,36 @@
...
+    try {
+      const resp = await this.client.fetch('GET', path)
+      return {
+        data: resp.rows.map(row => normalizeDoc(row.doc, this.doctype)),
+        meta: { count: resp.total_rows },
+        skip: 0,
+        next: false
+      }
+    } catch (error) {
+      throw error
```
y-lohse: I think we can remove the `try/catch` block here, no?

packages/cozy-stack-client/src/AppCollection.js
kosssi: I choose to remove doctype parameter.

packages/cozy-stack-client/src/AppCollection.js
kosssi: I remove it

enguerran: cf. https://repl.it/@enguerran/Throw-Catched-Error

PR #43, opened 2018-07-06
fix: generate query string without URLSearchParams

y-lohse:
I couldn't reproduce the initial problem so I don't have many details, but it appears `URLSearchParams.toString()` sometimes returns `[Object object]` — at least recently. Using a plain object seems to solve the problem.


y-lohse:
[🐛📱Erreur lors de la connexion a un Cozy sur Android](https://trello.com/c/2fWqxmk4/1039-erreur-lors-de-la-connexion-a-un-cozy-sur-android)

PR #47, opened 2018-07-18
feat: Move token refresh callback to a separate option

y-lohse:
Mixed in with the oauth options, the `onTokenRefresh ` callback was occasionally getting eaten up.


y-lohse:
[🐛 Corriger le problème de non-rafraichissement du token OAuth](https://trello.com/c/QrphxR16/1042-%F0%9F%90%9B-corriger-le-probl%C3%A8me-de-non-rafraichissement-du-token-oauth)

PR #48, opened 2018-07-18
chore: Auto release using travis

y-lohse:
Adresses #20 .

Warning — this is experimental as fuck.

If everything goes according to plan, travis will generate a `.npmrc` file containing a token that lerna will use to publish to npm. This should happen on git tags only and the git tag will be used as version number.

The main problem that I don't know how to fix just yet is that the new version number won't be commited. The real version number will be in the git tags, and we should try to keep the code up to date.
If this PR works as intended, I'll try to work out a solution to that. Maybe semantic release or something like that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,17 @@ cache:
   directories:
   - node_modules
 before_install:
-  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.7.0
-  - export PATH=$HOME/.yarn/bin:$PATH
-  - yarn --version
+- curl -o- -L https://yarnpkg.com/install.sh | bash -s
+- export PATH=$HOME/.yarn/bin:$PATH
+- echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
 before_script:
-  - lerna run build
+- lerna run build
 script:
 - yarn lint
 - yarn test --runInBand
+after_success:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,11 +7,18 @@ cache:
   directories:
   - node_modules
 before_install:
-  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.7.0
```
CPatchane: Maybe we can just the yarn from Travis, now it's correctly updated with the last version.

.travis.yml
```diff
@@ -7,11 +7,18 @@ cache:
...
-  - export PATH=$HOME/.yarn/bin:$PATH
-  - yarn --version
```
CPatchane: We don't need that, each time we run yarn it will output its version automatically.

.travis.yml
```diff
@@ -7,11 +7,18 @@ cache:
...
 before_script:
-  - lerna run build
+- lerna run build
 script:
 - yarn lint
-- yarn test --runInBand
\ No newline at end of file
+- yarn test --runInBand
+after_success:
+- test $TRAVIS_BRANCH = "master" && test $TRAVIS_REPO_SLUG = "cozy/cozy-client" && $TRAVIS_TAG &&
```
CPatchane: You may need to check if we are in a pull requestion or not?

PR #53, opened 2018-07-27
fix: mutation example

ptbrowne:
The second parameter passed to the children callback of `Query` is an object containing all the mutations.


ptbrowne:
thx

PR #74, opened 2018-08-03
Fix AppCollection

gregorylegarec:
In #41 I added AppCollection but it was a little bit too soon, because it appeared later that it was not working as expected in Collect.

So there is a fix, before merging I wish to:

* [x] Add tests
* [x] throw errors when not implemented or not available methods like `find()`, `get()`, etc... are called.

### DIFF
--- a/packages/cozy-stack-client/src/AppCollection.js
+++ b/packages/cozy-stack-client/src/AppCollection.js
@@ -32,4 +32,24 @@ export default class AppCollection {
       next: false
     }
   }
+
+  async find() {
+    throw new Error('find() method is not yet implemented')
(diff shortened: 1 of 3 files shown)


gregorylegarec:
@ptbrowne @y-lohse  Ready to be reviewed and to merge !

ptbrowne:
lint please

gregorylegarec:
@ptbrowne My bad the file was not saved on my machine. 👌

CPatchane:
Thanks @gregorylegarec 💛

y-lohse:
(Late to the party, but 👍 )

PR #76, opened 2018-08-06
Fix lerna watch

gregorylegarec:
Also, upgrade to latest version.


ptbrowne:
Thanks @gregorylegarec , I have updated the first post to reflect the why of this change.

PR #88, opened 2018-08-13
fix: Watch command 🐛

kosssi:
This script is used by the watch command "bin/build --watch". To work, it needs to forward the arguments.


y-lohse:
I'm confused, what does this do ?

ptbrowne:
@y-lohse it forwards the passed arguments to the underlying command.

kosssi:
@y-lohse sorry, thanks @ptbrowne to add description ;)

PR #89, opened 2018-08-13
docs: Update mutation docs

### LINKED ISSUE #54: Mutating example is not up-to-date
On the README.md https://github.com/cozy/cozy-client#mutating-data, the example is misleading as the createMutations function is not passed the `mutate` function but the query.

y-lohse:
Fixes #54 , but I'm a bit worried by this — we don't *actually* receive an instance of the client, we receive an [`ObservaleQuery`](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/ObservableQuery.js) which is mostly a proxy for the client, but there are some extra methods, and some are missing.

I still used `client` to avoid adding an extra term and because [that's what it's pretending to be](https://github.com/cozy/cozy-drive/blob/edd7fe1952989ba5b117861ce893774bd6d9462a/src/photos/ducks/timeline/index.js#L22), but maybe there's a better way to handle things here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,9 +133,8 @@ import { Query } from 'cozy-client'
 
 const query = client => client.find('io.cozy.todos').where({ checked: false })
 
-const createMutations = (mutate, ownProps) => ({
-  addTodo: label =>
-    mutate(client => client.create('io.cozy.todos', { label }))
+const createMutations = (client, ownProps) => ({
+  addTodo: label => client.create('io.cozy.todos', { label })
 }) 
 
 const App = () => (


ptbrowne:
I understand your point.

1. I think it received the query so that any mutations would directly work on the query, so that it would be easier to know which query to *touch/update*, and therefore which `Query` to `forceUpdate`.

This is indeed an implementation detail. Receiving the `client` seems OK to me. The work I am doing on the auto-update of queries should make 1. irrelevant since every query that would contain the doc would be touched.

PR #92, opened 2018-08-16
feat: auto update related queries

### LINKED ISSUE #68: Improve auto-update queries when a new document arrives
When a new document is created, all queries that should contain this document should be touched. For queries with no filtering, this seems easy : we only have to check the _type of the new document and touch all queries with the same type. But for queries that implements some kind of filtering (`.where`), we want to check that the new document belongs to the collections. The definition of the query could be converted to a filtering function and applied to the document to know if it can be inserted in a query. Conversely, when updating a document, we have to check if queries containing the old document should contain the new document, maybe the attribute on which the filtering is done has changed.

ptbrowne:
The goal of this PR is that when receiving results that concern several queries, every queries that can receive those results will be updated. It currently works by filtering
on the doctype and a limited subset of the mango operators (normal value, $gt and $lt) are supported.

- [x] Document creation (add document to every relevant query)
- [x] Document update (removal from queries that no longer contain the document)

Fix #68


packages/cozy-client/src/__tests__/store.spec.js
```diff
@@ -266,24 +268,51 @@ describe('Store', () => {
...
-        const result = { data: { ...TODO_1, done: true, id: TODO_1._id } }
+    describe('auto update', () => {
+      it('should update all the queries that can contain the document (creation)', async () => {
+        const result = { data: { ...TODO_1, done: true, _id: '1337', id: '1337' } }
         await store.dispatch(receiveMutationResult('foo', result))
-        const query1 = getQueryFromState(store.getState(), 'allTodos')
-        const query2 = getQueryFromState(store.getState(), 'allTodos2')
+        const query1 = getQueryFromStore(store, 'allTodos')
+        const query2 = getQueryFromStore(store, 'allTodos2')
+        expect(query1.data.find(doc => doc._id == '1337')).not.toBe(undefined)
```
kosssi: You can used `.not.toBeUndefined()`

ptbrowne:
fix https://github.com/cozy/cozy-client/issues/68

packages/cozy-client/src/store/setUtils.js
```diff
@@ -0,0 +1,11 @@
+export const union = (set1, set2) => {
```
kosssi: Ca n'existe pas dans lodash ?

packages/cozy-client/src/store/queries.js
```diff
@@ -79,80 +82,107 @@ const query = (state = queryInitialState, action) => {
...
-    Object.entries(queries).filter(([queryId, query]) =>
-      queryContains(query, id)
-    )
-  )
+const filterForQuery = query => {
+  const qdoctype = query.definition.doctype
+  const selectorFilterFn = query.definition.selector
+    ? selectorFilter(query.definition.selector)
+    : null
+  return datum => {
```
y-lohse: I don't want to nitpick about variable names, but what does "datum" mean in this context ?

packages/cozy-client/src/store/queries.js
```diff
@@ -79,80 +82,107 @@ const query = (state = queryInitialState, action) => {
...
-const touchQueries = queries => mapValues(queries, touchQuery)
+const mergeSets = (set1, set2) => {
+  const res = {}
+  for (let item of set1) {
+    res[item] = true
+  }
+  for (let item of set2) {
+    res[item] = true
+  }
+  return Object.keys(res)
```
y-lohse: If we're returning an array here, can't we do something like:

```
const res = []
  for (let item of set1) {
    res.push(item)
  }
  for (let item of set2) {
    res.push(item)
  }

return res
```

?

packages/cozy-client/src/store/queries.js
```diff
@@ -79,80 +82,107 @@ const query = (state = queryInitialState, action) => {
...
-          ...state[action.contextQueryId],
-          lastUpdate: Date.now()
-        }
+  if (isQueryAction(action)) {
+    const updater = autoQueryUpdater(action)
+    return mapValues(state, queryState => {
+      if (queryState.id == action.queryId) {
+        return query(queryState, action)
+      } else {
+        return updater(queryState)
```
y-lohse: 👍

packages/cozy-client/src/store/queries.js
ptbrowne: Just one doc. d3 uses the same convention. One piece of data is a datum. No problem about nitpicking about variable names.

Issue #50, opened 2018-07-20
Ability to connect a Query to a component without fetching ?


ptbrowne:
I think for now the Query cannot be used only to retrieve a document from the store (and observe its changes) without fetching the document again. When we know that the document is already in the db, it can be useful. I propose to have it as an option in Query ?

```
<Query nofetch query={...} />
```

drazik:
Isn't it the goal of the dedup link ? 

That's what I understand from https://github.com/cozy/cozy-client/blob/master/README.md#network-layer-cozy-link

ptbrowne:
The Dedup link here would deduplicate multiple queries. Here the goal is not to fire at all the query and only react to store changes.

ptbrowne:
Fixed via `fetchPolicy='cache-only'`

PR #93, opened 2018-08-17
feat: add prop not for Query not to fetch on mount

ptbrowne:
With this change, we can control whether or not the Query
will fire a request on mount. This can be useful if we
now that the data is being loaded in another manner, but we
still want the observable ability of the Query.


ptbrowne:
The Query has two roles : react to changes of the underlying data and fetch the data. The idea here is to decouple the fetching from the observing since there are times when you know that the data has been fetched.

I think it could be done in a cleaner way if you could put as an option to the Query the caching strategy, but I think this can be a first step.

packages/cozy-client/src/Query.spec.jsx
```diff
@@ -0,0 +1,40 @@
...
+    expect(client.query).toHaveBeenCalled()
+  })
+
+  it('should not fire a query fetch when mounted with initialFetch=false', () => {
+    const uut = mount(<Query query={queryDef} children={() => null} initialFetch={false}/>, { context })
+    expect(client.query).not.toHaveBeenCalled()
+  })
+
+  it('should be refreshed when relevant results are updated', () => {
+
```
kosssi: this test is empty ?

ptbrowne: Well spotted, fixing this ASAP

goldoraf:
@ptbrowne OK, so you want to display cached data without (network-)fetching it again if I understand well, and thus it's akin to Apollo's [fetch policies](https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-config-options-fetchPolicy). I understand it's a first step, but don't you think it'd be a good idea to try to plan a bit the next steps, and to settle on an API that would allow to expand capabilities without breaking existing code? That's the thing I found interesting with policies: a single prop allows to change easily the behavior of data retrieval.

ptbrowne:
> don't you think it'd be a good idea to try to plan a bit the next steps

Yes, this is why I opened an issue, to possibly discuss about the APIs before implementing : https://github.com/cozy/cozy-client/issues/50

Since I needed to have something working, I went with the simplest / decoupled thing.

> to settle on an API that would allow to expand capabilities

Yes. It seems the initialFetch={false} can be reworked into fetchPolicy='cache-only'.

> without breaking existing code

IMHO, there is no problem in at some point deprecating a property, without breaking downstream code. For example here, the `initialFetch` could have been translated to `fetchPolicy='cache-only'` in code and a codemod can be done. IHMO, the hardest refactoring are when things are too much coupled.

The reason why I did not go with the policy way right-away was because I don't want to wait to have the whole policy machinery ready before being able to use it. However, I think now that we can go with only understanding the `cache-only` value for now and improve later ? What do you think @y-lohse ? @kosssi ?

goldoraf:
> Yes. It seems the initialFetch={false} can be reworked into fetchPolicy='cache-only'.

That's exactly what I had in mind, sorry if I wasn't clear enough about that. I understand that you wanted something working fast, I just wanted to avoid prop name change.

ptbrowne:
I have updated with the `fetchPolicy` prop, thanks Raph for the reminder :)

Issue #57, opened 2018-07-27
Improving documentation


ptbrowne:
Watching : https://www.youtube.com/watch?v=azf6yzuJt54 and reading https://www.divio.com/blog/documentation/, there is the advice that documentation should be organized according to 4 different categories : 

- Tutorial (beginner wanting to have a first exposure with the project)
- References (diving deep into the nitty-gritty details)
- How-to guides (I know already about the project, and just want to achieve a task, goal oriented)
- Discussions (technical decisions, architecture etc...)

At first look, it seems a great way to divide things and to approach the good documentation problem. I propose that we use that.

At first for the references : JSDoc could be used to facilitate method documentation ?

[Pad](https://pad.cozycloud.cc/p/ideas-documentation-cozy-client)

y-lohse:
WIP on that in #96 

What is still missing is the automatic update of the doc, and I'm not sure how to best do this.

- Option 1 is that the user does this somehow. But I tried it with a `pre-commit` hook and it didn't seem to work, I still have to investigate why.
- Option 2 would be that travis does this. We could probably make this a part of the publishing process since lerna will make new commits anyway?

Any other ideas on how to best approach this?

ptbrowne:
Option 2 seems good. An approach (untested) :

```bash
if [[ $GIT_BRANCH == "master" ]]; then
	yarn docs
	git status --porcelain # exits with 0 if nothing has changed, 1 otherwise
	
	if [[ ! -z $? ]]; then
	  git add docs/
	  git commit -m 'chore: Update docs'
	  git push origin HEAD
	fi
fi
```

kosssi:
On npm we don't have documentation on packages: https://www.npmjs.com/package/cozy-client

y-lohse:
Right, I guess we have to add a REAME [here](https://github.com/cozy/cozy-client/tree/master/packages/cozy-client).

y-lohse:
This issue is taken over by #105.

PR #96, opened 2018-08-21
Generate docs with jsdoc2md

y-lohse:
There's now a `yarn doc` command that will generate an API documentation based on jsdoc comments. i have updated the existing jsdoc where needed so that the generated doc makes sense.

I've also moved the existing README to the docs folder and added a `toc.yml` file so we can publish this doc on docs.cozy.io

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -1,12 +1,3 @@
-# cozy-client
-
-A simple and declarative way of managing [cozy-stack](https://github.com/cozy/cozy-stack) API calls and resulting data.
-
-## Introduction
-
-`cozy-client` is a convenient yet powerful way to bind `cozy-stack` queries to your (p)React components — but you can still benefit of it if you're not using React!
-
-<<<<<<< HEAD
 If you don't use React, you can still benefit of it, see below.
 
 ## Setup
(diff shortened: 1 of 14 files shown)


docs/guide.md
```diff
@@ -0,0 +1,445 @@
+## Setup
```
ptbrowne: This file is big and it is hard to understand the organization of the guide.

I think either splitting it or adding a TOC would be good.

In Sublime text, there is a plugin that can generate a TOC for a markdown file automatically ;)

package.json
```diff
@@ -50,5 +50,10 @@
     "extends": [
       "cozy"
     ]
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "yarn doc"
```
ptbrowne: Is it fast enough ? Maybe we could do that in prepush ? And fail if the staging area has something ?

y-lohse: Shit that wasn't supposed to be part of this PR, especially because it doesn't work :'). BRB removing it.

docs/guide.md
y-lohse: Yes, definitely a next step as mentioned in #57 .

PR #97, opened 2018-08-22
fix: Sleep function was not imported

y-lohse:
We were missing an import, causing a bug when an index is created for the first time. I

- fixed it
- added test coverage
- completed the docs


ptbrowne:
Thanks !

PR #100, opened 2018-08-23
fix: Use only one sorting order

y-lohse:
Mango queries can't have opposing sorting orders (ie. one field `asc` and the other `desc`). But we had a default value of `desc`, so sorting by `asc` was impossible.

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -1,4 +1,5 @@
 import { uri, attempt, sleep } from './utils'
+import uniq from 'lodash/uniq'
 
 export const FETCH_LIMIT = 50
 
@@ -161,9 +162,15 @@ export default class DocumentCollection {
     const indexFields = this.getIndexFields({ ...options, selector })
     const indexId = options.indexId || (await this.getIndexId(indexFields))
     const { fields, skip = 0, limit = FETCH_LIMIT } = options
-    // Mango wants an array of single-property-objects...
+    // Mango wants an array of single-property-objects..
+    const sortOrders = options.sort
+      ? uniq(Object.values(options.sort))
+      : ['asc']
+    if (sortOrders.length > 1)
+      throw new Error('Mango sort can only use a single order (asc or desc).')
+    const sortOrder = sortOrders[0]
     const sort = options.sort
-      ? indexFields.map(f => ({ [f]: options.sort[f] || 'desc' }))
+      ? indexFields.map(f => ({ [f]: sortOrder }))
       : undefined
 
     return {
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -162,8 +162,9 @@ export default class DocumentCollection {
     const indexId = options.indexId || (await this.getIndexId(indexFields))
     const { fields, skip = 0, limit = FETCH_LIMIT } = options
     // Mango wants an array of single-property-objects...
+    const sortOrder = options.sort ? Object.values(options.sort)[0] : 'desc'
```
ptbrowne: Where does the [0] come from ? The order of the values is not guaranteed so this may sort in 1 direction sometime and in another another time, right ?

I think we should just throw in case of double sort by the user.

packages/cozy-stack-client/src/__tests__/DocumentCollection.spec.js
```diff
@@ -243,6 +243,22 @@ describe('DocumentCollection', () => {
...
+      const collection = new DocumentCollection('io.cozy.todos', client)
+      await collection.find({ done: false }, { sort: { label: 'asc', _id: 'desc' } })
+      expect(client.fetchJSON).toHaveBeenLastCalledWith(
+        'POST',
+        '/data/io.cozy.todos/_find',
+        {
+          limit: 50,
+          skip: 0,
+          selector: { done: false },
+          sort: [{ _id: 'asc' }, { done: 'asc' }, { label: 'asc' }],
```
ptbrowne: IMHO, this is very surprising from the point of view of the user. They tell the library to get results in a certain order, and the library does something behind her back to change that ?

y-lohse:
Yeah fair enough comments, I realized it would do that after writing the tests. I'll find a better wya to handle this.

y-lohse:
Here we go!

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -161,9 +162,15 @@ export default class DocumentCollection {
     const indexFields = this.getIndexFields({ ...options, selector })
     const indexId = options.indexId || (await this.getIndexId(indexFields))
     const { fields, skip = 0, limit = FETCH_LIMIT } = options
-    // Mango wants an array of single-property-objects...
+    // Mango wants an array of single-property-objects..
+    const sortOrders = options.sort
+      ? uniq(Object.values(options.sort))
+      : ['asc']
+    if (sortOrders.length > 1)
+      throw new Error('Mango sort can only use a single order')
```
kosssi: J'ajouterai bien 'asc' et 'desc' dans le descriptif de l'erreur

PR #102, opened 2018-08-23
Improve associations

drazik:
There was some flaws in the associations. I worked with @ptbrowne to improve it.


packages/cozy-client/src/CozyClient.js
```diff
@@ -1,5 +1,5 @@
 import StackLink from './StackLink'
-import Association from './associations'
+import createAssociation from './associations'
```
ptbrowne: Minor: 

I would not `export default createAssociation` inside './associations'.

I would `export { create }` and then here `import { create as createAssociation } from './associations'`

drazik: :+1:

Issue #105, opened 2018-08-24
Better documentation


y-lohse:
The docs improvements are scattered all over the place, so I'm creating one issue to track them. Here's what we want to do about the documentation:

- [ ] [Auto-update the docs](https://github.com/cozy/cozy-client/issues/57#issuecomment-414651363)
- [ ] [Adding a README for npm](https://github.com/cozy/cozy-client/issues/57#issuecomment-415313344)
- [ ] Reworking the guide. [A general idea here](https://github.com/cozy/cozy-client/issues/57#issue-345196387), the (pad is here](https://pad.cozycloud.cc/p/ideas-documentation-cozy-client). 
- [ ] Exhaustive jsdoc for `Query` (#49)
- [ ] Docs on `withMutations` (#56)
- [ ] Generally adding to the jsdocs

I'll keep this issue updated.

ptbrowne:
We spoke about " Generally adding to the jsdocs" with the Gangsters and we think it is a good idea to add documentation everytime we touch to a file. We should watch for this when reviewing. At some point we'll gain the habit ;)

PR #106, opened 2018-08-24
chore: Added readme for npm, see #105

y-lohse:


kosssi:
You add README.md only on `cozy-client` package, there are also `cozy-pouch-link`

PR #108, opened 2018-08-24
chore(Query): Remove unused vars

drazik:
ESLint was complaining about these variables not being used.


kosssi:
hum ESLint does not do its own job... why

PR #110, opened 2018-08-24
chore: Lint jsx files

ptbrowne:
JSX were not linted before. Now they are.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "packages/*"
   ],
   "scripts": {
-    "lint": "eslint 'packages/*/src/**/*.js'",
+    "lint": "eslint 'packages/*/src/**/*.{js,jsx}'",
     "test": "jest",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "packages/*"
   ],
   "scripts": {
-    "lint": "eslint 'packages/*/src/**/*.js'",
+    "lint": "eslint 'packages/*/src/**/*.js' 'packages/*/src/**/*.jsx'",
```
drazik: I suggest `'packages/*/src/**/*.{js,jsx}'`

kosssi: ou `eslint 'packages/*/src' --ext .jsx,.js` ?

ptbrowne: Ca a pas marché avec ta commande simon :(

Merci drazik

PR #116, opened 2018-08-28
Fix empty relationships

y-lohse:
When creating a document with an empty array as associations, an error was generated.

### DIFF
--- a/packages/cozy-client/src/CozyClient.spec.js
+++ b/packages/cozy-client/src/CozyClient.spec.js
@@ -1,10 +1,10 @@
 import configureStore from 'redux-mock-store'
 
-import { TODO_SCHEMA, TODO_1, TODO_2, TODO_3 } from './fixtures'
+import { TODO_SCHEMA, TODO_1, TODO_2, TODO_3 } from './__tests__/fixtures'
 
-import CozyClient from '../CozyClient'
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -151,6 +151,22 @@ describe('CozyClient', () => {
...
+        label: 'Buy RAM',
+        attachments: [{ _id: 12345, _type: 'io.cozy.files' }]
+      }
+      const mutation = client.getDocumentSavePlan(NEW_TODO, {
+        attachments: []
+      })
+      expect(Array.isArray(mutation)).toBe(false)
+      expect(mutation).toEqual({
+        mutationType: 'CREATE_DOCUMENT',
+        document: NEW_TODO
```
ptbrowne: I do not understand, here it means that attachments would not be removed from the TODO even if we tell `getDocumentSavePlan` to do it ?

y-lohse: No, AFAIK, `getDocumentSavePlan` is only used to add new relationships / attachments, not to remove them. I'm not sure how that would be done at the moment, but it doesn't seem to go through `getDocumentSavePlan`.

PR #120, opened 2018-08-29
feat: remove default limit

ptbrowne:
* remove FETCH_LIMIT, use the server limit by default


drazik:
Some linting issues and LGTM :+1:

PR #122, opened 2018-08-30
feat(pouchlink): Throw if can't get replicationurl

drazik:
I had a error because I was trying to call `getReplicationUrl` while my client doesn't have a token yet. So I added a throw with a clear message for that case.


ptbrowne:
OK thanks, I'm refactoring this part, I will keep the throw :)

y-lohse:
Do we still want this PR, or is it now part of #121 ?

ptbrowne:
It's part of https://github.com/cozy/cozy-client/pull/123

PR #123, opened 2018-08-30
feat: periodic replication of pouchdb databases

ptbrowne:
- Extracted the code that manage PouchDB databases
- Added replication


packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -1,79 +1,49 @@
...
-  if (!obj._id && obj.id) {
-    obj.id = obj._id
-  }
-  return obj
+const parseMutationResult = (original, res) => {
+  return { ...original, ...omit(res, 'ok') }
 }
 
-const addDoctype = (arr, doctype) => arr.map(x => ({ ...x, _type: doctype }))
+const doNothing = () => {}
```
enguerran: What is it used for?

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -1,79 +1,49 @@
...
 
-const pipe = fn => res => {
-  try {
-    fn(res)
-  } catch (e) {
-    console.warn('Error during pipe', e)
-  }
-  return res
-}
+const sanitized = doc => omit(doc, '_type')
```
enguerran: ES6 new syntax let us do this without the need of a third-party library:

```js
const sanitized = ({_type, ...doc}) => doc
```

Is there a particular reason for using this third-party library?

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -1,79 +1,49 @@
...
-  return { ...original, ...omit(res, 'ok') }
+export const getReplicationURL = (uri, token, doctype) => {
+  const basicAuth = token.toBasicAuth()
+  const authenticatedURL = addBasicAuth(uri, basicAuth)
+  return `${authenticatedURL}/data/${doctype}`
 }
 
-const doNothing = () => {}
+export const isExpiredTokenError = pouchError => {
+  return pouchError.error == 'code=400, message=Expired token'
```
enguerran: It is recommended to never use the evil twins, aka `==` and `!=` by Crockford.
[More details on StackOverflow](https://stackoverflow.com/a/359509/178575).

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -83,80 +53,44 @@ export default class PouchLink extends CozyLink {
...
-    const client = this.client
-    const basicAuth = client.token.toBasicAuth()
-    return (client.uri + '/data/' + doctype).replace('//', `//${basicAuth}`)
+  async onSyncError(error) {
+    if (isExpiredTokenError(error)) {
+      try {
+        await this.client.renewAuthorization()
+        this.pouches.startReplicationLoop()
+      } catch (err) {
+        console.warn('Could not refresh token, replication has stopped', err)
```
enguerran: Are we sure we don't want to throw the exception anyway?

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -0,0 +1,95 @@
+/** Handles the lifecycle of several pouches */
+
+import PouchDB from 'pouchdb'
+import fromPairs from 'lodash/fromPairs'
+import map from 'lodash/map'
+import repeatWithDelay from './repeat'
+
+const DEFAULT_DELAY = 30 * 1000
```
enguerran: :+1: this is easy to understand we are talking about 30 seconds.

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * - Creates/Destroys the pouches
+ * - Replicates periodically
+ */
+export default class PouchManager {
+  constructor(doctypes, options) {
+    this.pouches = fromPairs(
+      doctypes.map(doctype => [doctype, new PouchDB(doctype)])
+    )
```
enguerran: `_.fromPairs` could be replaced by `Array.prototype.reduce` (and the third-party library will not be required anymore):

```js
doctypes.reduce((pouches, doctype) => ({ ...pouches, [doctype]: new PouchDB(doctype) }), {})
```

See a [demo](https://codesandbox.io/s/5xwqxl7xkl)

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -0,0 +1,95 @@
...
+    this.options = options
+    this.getReplicationURL = options.getReplicationURL
+  }
+
+  destroy() {
+    this.stopReplicationLoop()
+    map(this.pouches, pouch => pouch.destroy())
```
enguerran: Here we can use `Object.values` (that is [well supported by browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values), expect from IE):

```js
Object.values(this.pouches).forEach(pouch => pouch.destroy())
```

The function `map` is often use to return an array. And the anonymous arrow function `pouch => pouch.destroy()` returns the result of `pouch.destroy()`. When reading this whole line, a confusion can be made: we build an array of result of `pouch.destroy()`, but this value is not stored in a variable. Is that something missing?

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -1,79 +1,49 @@
...
- * that it supports and knows how to respond to queries
- * and mutations.
+ * Link to be passed to cozy-client to support CouchDB. It instantiates
+ * PouchDB collections for each doctype that it supports and knows how
+ * to respond to queries and mutations.
  */
 export default class PouchLink extends CozyLink {
-  constructor(options = {}) {
+  constructor(opts = {}) {
+    const options = defaults({}, opts, DEFAULT_OPTIONS)
```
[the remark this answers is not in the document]
ptbrowne: `defaults` will not overwrite any value already set in `opts`

```
{...DEFAULT_OPTIONS, ...opts}
```

would do the job yes 👍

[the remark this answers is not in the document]
ptbrowne: The devil is in the details here.

`defaults` would  not overwrite the default options if the option provided is undefined.

```
> const nativeDefaults = (opts, defaults) => ({ ...defaults, ...opts })
> const DEFAULT_OPTIONS = { replicationDelay: 1000 }
> const options = { replicationDelay: undefined }
> nativeDefaults(options, DEFAULT_OPTIONS)
{ replicationDelay: undefined }
> _.defaults(options, DEFAULT_OPTIONS)
{ replicationDelay: 1000 }
```

packages/cozy-pouch-link/src/CozyPouchLink.js
ptbrowne: I find the intention clearer with `omit` here and `omit` has saner defaults IMHO.

packages/cozy-pouch-link/src/CozyPouchLink.js
ptbrowne: Thanks !

packages/cozy-pouch-link/src/CozyPouchLink.js
ptbrowne: To ensure that the forward option default value is to do nothing

packages/cozy-pouch-link/src/CozyPouchLink.js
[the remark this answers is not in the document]
enguerran: TIL

packages/cozy-pouch-link/src/CozyPouchLink.js
ptbrowne: Yes. It is an error but it should not break the application.

There will be code that will try to restart the replication when mobile gets the focus so I think it should not that much of a problem.

PR #124, opened 2018-08-31
Add login method on client

drazik:
Add a `login` method that is the opposite of `logout`: it (re)initializes things. Actually for now it is a public interface to `registerClientOnLinks`, which have an internal purpose.

I also refactored a little the links initialization.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -74,12 +74,14 @@ export default class CozyClient {
   constructor({ link, links, schema = {}, ...options }) {
     this.options = options
     this.idCounter = 1
-    this.link = link || links || new StackLink()
+    const lnk = link || links || new StackLink()
 
     this.createClient()
-    this.registerClientOnLinks(this.link)
-    if (Array.isArray(this.link)) {
-      this.link = chain(this.link)
+    this.registerClientOnLinks(lnk)
+    if (Array.isArray(lnk)) {
+      this.chain = chain(lnk)
+    } else {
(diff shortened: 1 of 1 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -71,21 +71,31 @@ const allValues = async x => {
...
-    if (Array.isArray(this.link)) {
-      this.link = chain(this.link)
+
+    this.links = Array.isArray(lnk) ? lnk : [lnk]
+    this.registerClientOnLinks()
+
+    if (Array.isArray(lnk)) {
+      this.chain = chain(lnk)
+    } else {
+      this.chain = lnk
```
ptbrowne: `this.chain = chain(this.links)` would remove the need for the if/else ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -71,21 +71,31 @@ const allValues = async x => {
...
     this.idCounter = 1
-    this.link = link || links || new StackLink()
+    const lnk = link || links || new StackLink()
 
     this.createClient()
-    this.registerClientOnLinks(this.link)
-    if (Array.isArray(this.link)) {
-      this.link = chain(this.link)
+
+    this.links = Array.isArray(lnk) ? lnk : [lnk]
```
ptbrowne:
```js
const ensureArray = arr => Array.isArray(arr) ? arr : [arr]
this.links = ensureArray(link || links || new StackLink())
```

? No `lnk` variable this way.

PR #125, opened 2018-09-03
Fix mobile authentication

nono:


packages/cozy-client/src/authentication/mobile.js
```diff
@@ -25,8 +25,8 @@ export const authenticateWithCordova = url => {
       }
 
       const onLoadStart = ({ url }) => {
-        const accessCode = /\?code=(.+)$/.test(url)
-        const state = /\?state=(.+)$/.test(url)
+        const accessCode = /\bcode=/.test(url)
+        const state = /\bstate=/.test(url)
```
ptbrowne: It's been several times that we have trouble parsing URLs with regexes. I think it's better if we use a dedicated url parser. It  will also be more readable.

nono: :+1: it would be a lot better to parse URLs with a dedicated URL parser. It's just that this code is in cordova and I don't know what APIs are available here, or how to include an external module.

ptbrowne: OK I will open a dedicated issue for this.

ptbrowne:
LGTM

drazik:
@nono asked me to try this PR this morning but it was not working. That's why we reverted #118

ptbrowne:
So we close ?

PR #130, opened 2018-09-03
feat(pouch): Handle errors for pouches destroy

drazik:
If the databases were already destroyed, this throws an error.


drazik:
When login after logout. In `CozyPouchLink::registerClient` we destroy databases. But in that case they were already destroyed by `CozyPouchLink::reset` when logout.

PR #134, opened 2018-09-04
fix(oauth): Don't forget data in updateInformation

drazik:
When we do a PUT on `/auth/register/:clientID`, the stack don't return all the fields that were sent. So we need to merge the current data with what the stack returns. Otherwise there's a potential loss of data. For example on banks, we lost the `registrationAccessToken` key.


drazik:
See https://github.com/cozy/cozy-client/issues/135

PR #136, opened 2018-09-04
docs: add runnable example for relationships

ptbrowne:
Runnable examples can serve as end to end test and also can be learned from.


packages/cozy-client/src/CozyClient.js
```diff
@@ -400,6 +401,7 @@ export default class CozyClient {
     switch (type) {
       case 'has-many':
         if (query) {
+          console.log('query for ', document)
```
drazik: Do we want to keep this console.log ?

PR #139, opened 2018-09-05
fix(client): Don't check skip property

drazik:
This leaded to an infinite loop creating `INIT_QUERY` actions endlessly

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -288,7 +288,7 @@ export default class CozyClient {
     this.ensureStore()
     const existingQuery = getQueryFromState(this.store.getState(), queryId)
     // Don't trigger the INIT_QUERY for fetchMore() calls
-    if (existingQuery.fetchStatus !== 'loaded' || !queryDefinition.skip) {
+    if (existingQuery.fetchStatus !== 'loaded') {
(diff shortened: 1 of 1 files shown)


ptbrowne:
@goldoraf we had an infinite loop here. Was the intention of initQuery also to set the query definition in the store ? In this case do you think it would be good to check the equality of the queryDefinition with the one in the store and then dispatch a "setQuery" action ?

goldoraf:
The intention behind [this check](https://github.com/cozy/cozy-client/pull/139/commits/b8f21a25b2c9049568bc04b0209129416b331ce3#diff-405c53e28840707e47a25b26d2ed602bL291) was to ensure we don't trigger many fetches for the same query and that we don't reinit the query when we just want to fetch more files and to include them into the (paginated) results of the query already in store. So I think this check is necessary right nowif we don't want to break things. I choose to store the query definition with initQuery so that we could eventually dispense ourselves of providing a query name (and instead rely on comparison of the definitions).

But I don't see how it could result in an infinite loop. What does the calling code looks like? Can you give me a bit more context?

drazik:
Sorry @goldoraf, it seems we missed something. I tried to use the original condition to show you the problem in a video, but everything went well. I propose to close this PR, and I will try to see what tests I can add and open a new PR.

PR #140, opened 2018-09-05
fix: jsonapi needs to extract the doc from the rows

ptbrowne:
- allDocs sends back rows with doc inside

Some refactors along the way


drazik:
@ptbrowne tests are still failing :thinking:

ptbrowne:
Good tests 🐶 We had a problem when doing mutations, this is fixed.

PR #145, opened 2018-09-06
feat(client): Add setData method

drazik:
When using PouchLink replication, we need something to pass the result of the replication to the client. The idea is to provide a `setData` method that we can call when the pouch synced successfuly. The link passes the data synced during the replication process to `setData` and the internal store is updated accordingly.

### DIFF
--- a/packages/cozy-client/src/CozyClient.spec.js
+++ b/packages/cozy-client/src/CozyClient.spec.js
@@ -151,6 +151,20 @@ describe('CozyClient', () => {
     })
   })
 
+  describe('setData', () => {
+    it('should dispatch RECEIVE_QUERY_RESULT actions', () => {
+      const data = {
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -151,6 +151,20 @@ describe('CozyClient', () => {
...
+    it('should dispatch RECEIVE_QUERY_RESULT actions', () => {
+      const data = {
+        'io.cozy.todos': [{ id: 1, done: true }, { id: 2, done: false }],
+        'io.cozy.people': [{ id: 1, name: 'Alice' }, { id: 2, name: 'Bob' }]
+      }
+
+      client.setData(data)
+
+      expect(store.getActions().length).toBe(Object.keys(data).length)
+      expect(store.getActions()[0].type).toBe('RECEIVE_QUERY_RESULT')
```
ptbrowne: TIL getActions

PR #150, opened 2018-09-10
Sharing previews

y-lohse:
This PR adds the necessary routes to handle preview pages in sharings:

- Providing a `preview_path` to the `share` function
- Fetching the current set of permissions

The rest of the changes are docs and tests.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -18,6 +18,12 @@ files associated to a specific document</p>
 </dd>
 <dt><a href="#module_OAuthClient">OAuthClient</a></dt>
 <dd></dd>
+<dt><a href="#module_PermissionCollection">PermissionCollection</a></dt>
+<dd><p>Interact with permissions</p>
+</dd>
+<dt><a href="#module_SharingCollection">SharingCollection</a></dt>
+<dd><p>Interact with sharing doctypes</p>
+</dd>
 </dl>
 
 ## Constants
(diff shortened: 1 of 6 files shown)


y-lohse:
[(3) En tant que destinataire d'un partage, je peux visualiser le contenu avant d'accepter le partage](https://trello.com/c/9vomUaDf/1026-3-en-tant-que-destinataire-dun-partage-je-peux-visualiser-le-contenu-avant-daccepter-le-partage)

docs/api.md
```diff
@@ -153,6 +160,19 @@ Renews the token if, for instance, new permissions are required.
 
 **Kind**: instance method of [<code>CozyClient</code>](#module_CozyClient)  
 **Returns**: <code>object</code> - Contains the fetched token and the client information.  
+<a name="module_CozyClient+setData"></a>
+
+### cozyClient.setData(data)
```
goldoraf: I missed the PR that introduced this method, but it's a bad idea imho: its name is too generic and doesn't reflect its use. You can be sure that it will be used for bad things. And moreover, I don't get in which cases the PouchDB link needs to update the store itself. What's the precise use case?

goldoraf: Yeah I saw it (late), but it doesn't add much info

drazik: We need to update the store to notify a Query component that it's query has been updated after a pouch replication completed. Maybe we can do it in another way ?

PR #155, opened 2018-09-12
fix(stack-client): Map redirect_uris to camelcase

drazik:
This property was not mapped in `camelCaseOAuthData`, but we need it to be present in `register`, otherwise we send `undefined` to the stack, and we receive an error telling us that the `redirect_uris` parameter is mandatory.

### DIFF
--- a/packages/cozy-stack-client/src/OAuthClient.js
+++ b/packages/cozy-stack-client/src/OAuthClient.js
@@ -88,7 +88,8 @@ export default class OAuthClient extends CozyStackClient {
       client_name: 'clientName',
       client_secret: 'clientSecret',
       registration_access_token: 'registrationAccessToken',
-      software_id: 'softwareID'
+      software_id: 'softwareID',
+      redirect_uris: 'redirectURI'
     }
 
     const result = {}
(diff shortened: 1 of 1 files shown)


packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -111,6 +112,18 @@ export default class OAuthClient extends CozyStackClient {
   async register() {
     if (this.isRegistered()) throw new Error('Client already registered')
 
+    const mandatoryFields = ['redirectUR']
```
ptbrowne: redirectURI ?

drazik: yep, don't know why I remove a char before commiting :thinking: thanks

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -111,6 +112,18 @@ export default class OAuthClient extends CozyStackClient {
   async register() {
     if (this.isRegistered()) throw new Error('Client already registered')
 
+    const mandatoryFields = ['redirectURI']
+    const fields = Object.keys(this.oauthOptions)
+    const missingMandatoryFields = difference(mandatoryFields, fields).join(
+      ', '
+    )
```
ptbrowne: Alternative name and method :

```
const missingFields = mandatoryFields.filter(fieldName => this.oauthOptions[fieldName])
```

I would join in the error since it's joined only for formatting.

PR #158, opened 2018-09-14
feat: Relationships and schema improvements

ptbrowne:
- Schema related functionality has been extracted to the Schema class

Two benefits : 

1. less functionality in Cozy Client and clearer where are the responsibilityies.
2. Since the data is arranged for our need inside the schema class 
(relationships have their name inside, relationships are indexed
by doctype), there is much less work to do when hydrating (less
objects are created and passed around).
3. BREAKING CHANGE: Each doctype accessed in Cozy Client must have its doctype
declared in the schema

- Relationships improvements

It is now possible to declare a custom relationships on the application side.
If the application has specific needs (ids of the relation declared in a custom
way, custom querying), it is possible to declare `query` inside a custom
relationship.

Now it is clearer that this is the responsibility of the relationship class to
know how to query.

- Two new relationships

  + HasOneInPlace
  + HasManyInPlace

They represent relationships where the info on the related documents is stored directly under the attribute and not in the `relationships` attribute.

---

The terms relationships and associations are used interchangeably in the codebase.
Relationship is the term used by JSONAPI, I would personnally in favor of
removing association since this can led to confusion (are relationship and associations
the same ? is one the instanciation of the other ? is one in the schema and
the other in the document ?)


packages/cozy-client/src/associations/index.js
```diff
@@ -27,14 +43,9 @@ export const create = (
...
-      return doctype === 'io.cozy.files'
-        ? new HasManyFilesAssociation(target, name, doctype, accessors)
-        : new HasManyAssociation(target, name, doctype, accessors)
-    case 'has-many-UNSAFE':
-      return new HasManyUNSAFEAssociation(target, name, doctype, accessors)
-    default:
-      throw new Error(`Can't handle ${type} associations`)
+  const cls = getClass(doctype, type)
+  if (!cls) {
+    throw new Error(`Can't handle ${type} associations`)
```
drazik: Is it really possible to enter this condition, since we already `throw` in `getClass` if we don't return anything ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -164,37 +159,11 @@ export default class CozyClient {
   }
 
   async save(document, mutationOptions = {}) {
-    const ret = await this.validate(document)
+    const ret = await this.schema.validate(document)
     if (ret !== true) throw new Error('Validation failed')
     return this.mutate(this.getDocumentSavePlan(document), mutationOptions)
   }
 
-  async validate(document) {
```
y-lohse: Just a head's up that this will also be a breaking change (cozy-photo uses `client.validate`)

packages/cozy-client/src/associations/HasManyInPlace.js
```diff
@@ -29,6 +29,27 @@ export default class HasManyInPlace extends Association {
...
 
+  addById(id) {
+    const rel = this.getRelationship()
+    rel.push(id)
+  }
+
+  removeById(id) {
+    const rel = this.getRelationship()
+    const index = rel.indexOf(id)
+    rel.splice(index, 1)
```
y-lohse: Maybe you want to test the existence first? Splicing on `-1` has [undesirable effects](https://repl.it/@y_lohse/PrizeDentalProcedures).

packages/cozy-client/src/associations/helpers.js
```diff
@@ -1,33 +1,60 @@
...
+    meta: response.meta,
+    next: response.next,
+    skip: response.skip
   })
-  return dehydrated
+
+const aliases = {
+  'io.cozy.files:has-many': HasManyFiles,
+  'has-many': HasMany,
+  'belongs-to-in-place': HasOneInPlace,
```
y-lohse: I this just a convenience alias for `has-one-in-place` or am I missing something?

packages/cozy-client/src/CozyClient.js
ptbrowne: Done in https://github.com/cozy/cozy-client/pull/158/commits/9c9d061a5adb9be8cd363795f748cdcf633827db

packages/cozy-client/src/associations/HasManyInPlace.js
ptbrowne: Rookie mistake :/ Thanks, well spotted !

Fixed in https://github.com/cozy/cozy-client/pull/158/commits/52b169de8fdde28c908e3cd54df877525e3ceea9

packages/cozy-client/src/CozyClient.js
y-lohse: Cool, thanks!

PR #184, opened 2018-09-25
Split parts of the docs

y-lohse:
Takes out some parts of the main guide and puts them in secondary guides.

The part about links will need some rewording, and we have to add links to the various How-to's. I think both can be done in subsequent PRs.

### DIFF
--- a/docs/how-to/existing-store.md
+++ b/docs/how-to/existing-store.md
@@ -1,0 +1,26 @@
+# Integrating with an existing redux store
+
+`cozy-client` uses redux internally to centralize the statuses of the various fetches and replications triggered by the library, and to store locally the data in a normalized way. If you already have a redux store in your app, you can configure `cozy-client` to use this existing store:
+
+```js
+import CozyClient from 'cozy-client'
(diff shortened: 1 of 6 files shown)


docs/how-to/existing-store.md
```diff
@@ -0,0 +1,26 @@
+# Integrating with an existing redux store
+
+`cozy-client` uses redux internally to centralize the statuses of the various fetches and replications triggered by the library, and to store locally the data in a normalized way. If you already have a redux store in your app, you can configure `cozy-client` to use this existing store:
```
drazik: In the code below, I understand that the app's store uses the `cozy-client` internal store, not that `cozy-client` uses the app's store. Am I wrong ? I find it confusing.

(it may not be the purpose of this PR, since you just splitted the doc in smaller pieces)

y-lohse: Mmh to me, "you can configure `cozy-client` to use this existing store:" means that cozy-client will use the app's store, which is the actual behavior. So not sure how we could further clarify things. Do you know what part is confusing you?

PR #185, opened 2018-09-25
Fix offline on pouch link

kosssi:
When an application is offline you must be able to query Pouch
A limitation has been put in place to avoid launching replications when device are in offline mode


packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -114,10 +114,13 @@ export default class PouchManager {
     }
     console.info('Start replication loop')
     delay = delay || this.options.replicationDelay || DEFAULT_DELAY
-    this._stopReplicationLoop = promises.setInterval(
-      () => this.replicateOnce(),
-      delay
-    )
+    this._stopReplicationLoop = promises.setInterval(() => {
+      if (window.navigator.onLine) {
```
drazik: :+1:

PR #202, opened 2018-10-01
fix: Unregistering client

Crash--:
Unregistering client when registration process is aborted

The inAppBrowser can be closed during the registration process. In that case, we need to `unregistered` the client. 

Repro : 
- Open a login session 
- Write your cozy address 
- When you're asked to put your password, close your window 
- Try to write again your cozy address 

Fix Drive's issue : `Client is already registered`


packages/cozy-client/src/CozyClient.js
```diff
@@ -477,20 +478,26 @@ export default class CozyClient {
...
+        token,
+        infos: client.oauthOptions,
+        client: client.oauthOptions // for compat with Authentication comp reasons
+      }
+    } catch (error) {
+      /* if REGISTRATION_ABORT is emited, we have to unregister the client. */
+      if (error.message === REGISTRATION_ABORT) {
+        const client = this.getClient()
+        client.unregister()
+      }
```
y-lohse: Should we throw the error after the unregistration? If the app needs to do something itself this could be useful, and it wouldn't change the existing behavior (in Drive we'd to a catch with presumably nothing in it?)

Crash--: Done

y-lohse: Actually I was thinking about all error cases, not just the abort, sorry if that wasn't very clear.
If there is an error in any of the other function calls, it will be silently absorbed here and we'll have to way to detect it.

Crash--: Good catch! Is it ok now?

packages/cozy-client/src/CozyClient.js
```diff
@@ -477,20 +478,28 @@ export default class CozyClient {
...
+        token,
+        infos: client.oauthOptions,
+        client: client.oauthOptions // for compat with Authentication comp reasons
+      }
+    } catch (error) {
+      /* if REGISTRATION_ABORT is emited, we have to unregister the client. */
+      if (error.message === REGISTRATION_ABORT) {
+        const client = this.getClient()
+        client.unregister()
+        throw new Error(REGISTRATION_ABORT)
```
kosssi: this line is not mandatory ;)

packages/cozy-client/src/CozyClient.js
```diff
@@ -477,20 +478,28 @@ export default class CozyClient {
...
+      }
+      throw new Error(error)
```
kosssi: error is already an Error no ?

packages/cozy-client/src/CozyClient.js
Crash--: I worked on a cozy-error lib but didn't push it in this PR. 

The idea is to be able to throw some custom errors. With those customs errors, we can configure Sentry to not send it... Our (drive) Sentry is full of "non error", and this `Registration_abort` is part of it. 

I'll remove it for now.

PR #210, opened 2018-10-09
Breaking change: Respecting the sort order

y-lohse:
Until now, sorting options where passed as a single object, like `{ zebras: 'asc', abyssinian: 'asc' }`. The problem is that the order of the sort usually matters, but with an object the order of keys is not guaranteed. Typically the example above would end up sorting on `abyssinian` first and `zebra` second.

The new format is the same as what mango expects: `[{ zebras: 'asc' }, { 'abyssinian': 'asc' }]`. 

The old format is *still* supported, although you'll get a deprecation warning. 

The breaking change comes from the usage of selectors combined with sorting order. A query with `{ selector: { done: true }, sort: { name: 'asc' } }` would previously resolve to a sorting order of `[{ done: 'asc' }, { name: 'asc' }]`, but now it becomes `[{ name: 'asc' }, { done: 'asc' }]`. This is *probably* what you wanted anyway, but still, it's a breaking change.

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -207,7 +207,7 @@ export default class DocumentCollection {
 
       for (const field of indexedFields) {
         if (!sort.find(sortOption => head(Object.keys(sortOption)) === field))
-          sort.unshift({ [field]: sortOrder })
+          sort.push({ [field]: sortOrder })
       }
     }
 
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -1,5 +1,7 @@
 import { uri, attempt, sleep } from './utils'
 import uniq from 'lodash/uniq'
+import transform from 'lodash/transform'
+import head from 'lodash/head'
 import * as querystring from './querystring'
 
 export const normalizeDoc = (doc, doctype) => {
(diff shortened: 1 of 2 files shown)


y-lohse:
[(8) Déplacer un fichier ou un dossier](https://trello.com/c/OSfXpP1q/847-8-d%C3%A9placer-un-fichier-ou-un-dossier)

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -178,19 +180,39 @@ export default class DocumentCollection {
...
+        'Passing an object to the "sort" is deprecated, please use an array instead.'
+      )
+      sort = transform(
+        sort,
+        (acc, order, field) => {
+          acc.push({ [field]: order })
+          return acc
+        },
+        []
+      )
```
ptbrowne:
```js
sort = Object.entries(sort).map(([field, order]) => ({ [field]: order }))
```
?

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -178,19 +180,39 @@ export default class DocumentCollection {
...
+    }
+
+    const indexedFields = this.getIndexFields({ sort, selector })
+    const indexId = options.indexId || (await this.getIndexId(indexedFields))
+
+    if (sort) {
+      const sortOrders = uniq(
+        sort.map(sortOption => head(Object.values(sortOption)))
```
ptbrowne: Any reason to use head instead of [0] ?

packages/cozy-stack-client/src/DocumentCollection.js
y-lohse: Yeah I don't mind either way. Not knowing how a lodash function works is more or less true of all lodash functions though :) But happy to change it if you think the native version makes the code clearer!

PR #218, opened 2018-10-18
fix(ReferencedFiles): Properly detect next pages

y-lohse:
Responses to the `/relationships/references` don't have a `rows` attribute, but they have a `next` property in the `links` when there is more content, as per the JSONAPi spec.


y-lohse:
[Album photo - Problème de pagination](https://trello.com/c/BYtj4B16/1305-album-photo-probl%C3%A8me-de-pagination)

ptbrowne:
Cool, can you add a test please ?

PR #221, opened 2018-10-23
feat: Add addById / removeById to HasMany

Crash--:
Some interrogations: 
- Do I need to use `exists` before doing both action ? 
- Do I need to replace `_id` by the whole  document ? (cf `exists(document)` ?)

### DIFF
--- a/packages/cozy-client/src/associations/HasMany.js
+++ b/packages/cozy-client/src/associations/HasMany.js
@@ -65,10 +65,30 @@ export default class HasMany extends Association {
   }
 
   exists(document) {
+    return this.existsById(document._id)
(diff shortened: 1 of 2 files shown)


ptbrowne:
> Do I need to use exists before doing both action ?

Mmh,  if you add twice the _id, it will a problem since an association needs to be a set. There is no meaning having twice an id in HasMany. So yes, for me addById should have exists and be idempotent. Same with removeById.

> Do I need to replace _id by the whole document ? (cf exists(document) ?)

I don't understand :/

Crash--:
@ptbrowne Updated with tests

PR #222, opened 2018-10-24
fix: When creating a document we create also its relationships

Crash--:
Let's say I have the following schema : 
```jsx
 schema: {
      contacts: {
        doctype: 'io.cozy.contacts',
        relationships: {
          groups: {
            type: 'has-many',
            doctype: 'io.cozy.contacts.groups'
          }
        }
      },
      groups: { doctype: 'io.cozy.contacts.groups' }
    }
```

When I create a `contact` without `groups` :
```jsx 
createContact: attributes =>
    client.create(
      'io.cozy.contacts',
      attributes
    ),
```
It doesn't create  empty `groups relationships`. But this empty relationship is mandatory if I want to add a relationships. 

I saw the third argument of `create` but I do not want to put useless/fake relation in it.


packages/cozy-client/src/associations/HasMany.js
```diff
@@ -39,6 +39,15 @@ const empty = () => ({
  * ```
  */
 export default class HasMany extends Association {
+  constructor(target, name, doctype, { get, query, mutate, save }) {
```
drazik:
```suggestion
  constructor(target, name, doctype, options) {
```

packages/cozy-client/src/associations/HasMany.js
```diff
@@ -39,6 +39,15 @@ const empty = () => ({
...
+    super(target, name, doctype, { get, query, mutate, save })
```
drazik:
```suggestion
    super(target, name, doctype, options)
```

packages/cozy-client/src/associations/HasMany.js
```diff
@@ -39,6 +39,15 @@ const empty = () => ({
...
+    if (!this.target.relationships) this.target.relationships = {}
+    if (!this.target.relationships[this.name]) {
+      this.target.relationships[this.name] = { data: [] }
+    }
+  }
+
+  createEmptyRelationships(relationshipsName, doctype) {}
```
drazik: What is the purpose of this method ?

packages/cozy-client/src/associations/HasMany.js
```diff
@@ -11,6 +11,11 @@ const empty = () => ({
  * Related documents are stored in the relationships attribute of the object,
  * following the JSON API spec.
  *
+ * Responsible for
+ *
+ * - creating relationship
```
ptbrowne:
```suggestion
 * - Creating relationship
  * - Adding / removing relationship
```

PR #223, opened 2018-10-24
Optimize id queries

ptbrowne:
When fetching has-many relationships, we often end up with a lot of id queries (query for a single documents, or for several documents whose id are known): 

```
[
  new QueryDefinition({ id: 1, doctype: 'io.cozy.bills' }),
  new QueryDefinition({ id: 2, doctype: 'io.cozy.bills' }),
  new QueryDefinition({ id: 3, doctype: 'io.cozy.bills' })
]
```

This result in many requests being made, which increase latencies and decrease performances.

Here the queries are optimized before being sent resulting in: 

```
[
  new QueryDefinition({ ids: [1, 2, 3], doctype: 'io.cozy.bills' })
]
```


packages/cozy-client/src/store/documents.js
```diff
@@ -7,9 +7,15 @@ import isEqual from 'lodash/isEqual'
...
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn('Document without _type', document)
+    }
+    throw new Error('Document without _type')
   }
   if (!document._id) {
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn('Document without _id', document)
+    }
     throw new Error('Document without _id', document)
```
kosssi: You want to delete document here ? like [here](https://github.com/cozy/cozy-client/pull/223/files#diff-1b915ff4945478dc0700c73893b71ae0R13)

ptbrowne:
```suggestion
    throw new Error('Document without _id')
```

PR #226, opened 2018-10-24
docs: Add method docstrings

ptbrowne:
Hopefully, these will help future contributors ;)


packages/cozy-client/src/CozyClient.js
```diff
@@ -405,19 +405,27 @@ export default class CozyClient {
   }
 
   hydrateRelationships(document, relationships) {
-    const methods = this.getRelationshipStoreAccessors()
+    const methods = this.getRelationshipcessors()
```
drazik: I think you didn't want to remove these characters ;)

packages/cozy-client/src/CozyClient.js
```diff
@@ -426,6 +434,14 @@ export default class CozyClient {
     )
   }
 
+  /**
+   * Returns the accesors that are given to the relationships for them
```
y-lohse:
```suggestion
   * Returns the accessors that are given to the relationships for them
```

packages/cozy-client/src/CozyClient.js
ptbrowne: Eagle eye ! 🦅

PR #228, opened 2018-10-25
feat: Add Sift for mango selectors

Crash--:
No need to create our own implementation of mango ;)


packages/cozy-client/src/store/mango.js
```diff
@@ -1,33 +1,6 @@
...
-        return val !== selValue
-      } else {
-        throw new Error(`Unknown operator ${$operator}`)
-      }
-    })
-    return and(attrConditions)(doc)
-  })
-  return and(filters)
+export const selectorFilter = selector => doc => {
+  const test = sift(selector, [doc])
```
ptbrowne: It would be good if we did not have to wrap in an array.

ptbrowne: Looking at https://github.com/crcn/sift.js/tree/master#api, I think we could just have

```
export const selectorFilter = selector => sift(selector)
```

> array - sifts against target array. Without this, a function is returned

```js
var siftExists = sift({$exists:true});

siftExists('craig'); //true
siftExists(null); //false
['craig',null].filter(siftExists); //['craig']
```

packages/cozy-client/src/store/mango.spec.js
```diff
@@ -45,4 +45,25 @@ describe('mango selector', () => {
...
+            $in: ['a@a.fr']
+          }
+        }
+      }
+    })
+    expect(
+      filter({ phone: { number: 0 }, email: { address: ['a@a.fr'] } })
+    ).toBe(true)
+    expect(filter({ phone: 10 })).toBe(false)
+  })
```
ptbrowne: Let's remove the tests since here we would be testing sift ?

packages/cozy-client/src/store/mango.js
ptbrowne: I think we could remove this file and just use sift in `queries`.

PR #232, opened 2018-10-28
fix: Repair HasMany files relationships

ptbrowne:
In https://github.com/cozy/cozy-client/commit/a9fa010ed15c8b3c28fab426e78a47059d71968b, relationships filling was broken : the relationships attribute was no longer filled at fetch time. This was particularly important for documents whose relationship data was "on the other side" of the relation.

For example, for photo albums, relationship data for the photos, is on the file side of the relationship and this is returned in the `referencedBy` queries. This data needs to be stored on the album side, in the `relationships` attribute, for it to be easily managed on the client.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -10,6 +10,10 @@ import { REGISTRATION_ABORT } from './const'
 
 import StackLink from './StackLink'
 import { create as createAssociation } from './associations'
+import {
+  responseToRelationship,
+  attachRelationships
+} from './associations/helpers'
 import { dehydrate } from './helpers'
 import { QueryDefinition, Mutations } from './queries/dsl'
 import CozyStackClient, { OAuthClient } from 'cozy-stack-client'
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -306,38 +309,73 @@ export default class CozyClient {
...
 
-    // Relationships can yield documents, ready for use, or definitions needing
-    // to be executed
+    // Relationships can yield "queries" that are already resolved documents.
+    // These do not need to go through the usual link request mechanism.
     const { documents = [], definitions = [] } = groupBy(
-      definitionOrDocuments,
+      docIdRelQueries.map(triplet => triplet[2]),
       d => (d instanceof QueryDefinition ? 'definitions' : 'documents')
     )
```
Crash--: Are we mapping three times on the "same" content? Is there any space for optimization here?

ptbrowne: Yes. https://github.com/cozy/cozy-client/pull/232/commits/5ec2e73494970db53892c7c7ed8f9d989239ca7c

ptbrowne:
Travis's push is failing because of time sensitive tests. These are not really failures.

PR #237, opened 2018-10-29
fix(pouch): Wait end of replication before triggering another one

drazik:
We had an issue because the promise was not returned. It means that the replication loop was not waiting for the current iteration to be finished before triggering the next one.

Now, the behavior is the one intended: when we start the loop, the first replication is done, then we wait for a delay (30 seconds by default), then the second replication is done, and so on.


drazik:
Merging since we saw that directly with @ptbrowne

PR #241, opened 2018-10-31
Custom requests

y-lohse:
### Problem

I needed to make a change to the [`upload`](https://github.com/cozy/cozy-client/blob/custom-requests/packages/cozy-client/src/CozyClient.js#L235) mutation function, but after consideration, I think it shouldn't be there at all.

Mutations like create/read/update/delete make perfect sense, and even adding/removing references is ok, as those are common across doctypes. But some doctypes have special behaviors — files, apps, settings, etc. If we want to support all these behaviors, we will need a lot of custom functions in the main CozyClient, and we will have [to integrate every one of them in all our Links](https://github.com/cozy/cozy-client/blob/custom-requests/packages/cozy-client/src/StackLink.js#L69). I think this will make cozy-client bloated and it's maintenance complicated. So I want us to be able to execute a mutation function outside of cozy-client.

### Solution

[The first change I made](https://github.com/cozy/cozy-client/commit/235e12fa866cb0f5f465000a2dbc4b3fb69c0798) is that Links don't have to know about all mutations anymore. If they don't know how to handle one, they should just forward it to the next link (instead of throwing an error).

Instead, there is now a default "last link" that will be called if all previous links forward a request or mutation. This link does the final error-throwing when all hope is lost.

In the [second commit](https://github.com/cozy/cozy-client/commit/a32be8f4ebe4a2dcec92a0f1ce70ea0a26095162), I added a new feature: mutations and requests can have a `execute` property which acts as a callback, and the result of that function will be returned by the default link.

### Downsides

The callback function can return anything, whereas before we had more control over what it did. I don't think it's a big problem though — most of the time you'll want to call a collection from cozy-client anyway, and if the function returns non-sense… the queries and documents will ignore it.

I'm not super happy with the `execute` name, as I find it too generic.

Links handle `requests` and `mutations` but doesn't differenciate much between the 2. I mostly wanted the ability to execute custom mutations, but as a side effect it also becomes possible for regular requests. I think it's ok, but in the longer term having a stronger separation between queries and mutations might make sense.

### Next ?

What do you think about this change? If it's ok for you, I'll add tests and docs before merging the PR.

### DIFF
--- a/packages/cozy-client/src/CozyClient.spec.js
+++ b/packages/cozy-client/src/CozyClient.spec.js
@@ -311,7 +311,7 @@ describe('CozyClient', () => {
     it('should call the link with the query', async () => {
       await client.query(query)
       expect(requestHandler).toHaveBeenCalledTimes(1)
-      expect(requestHandler).toHaveBeenCalledWith(query)
+      expect(requestHandler.mock.calls[0][0]).toBe(query)
     })
 
     it('should handle queries with includes', async () => {
@@ -394,7 +394,7 @@ describe('CozyClient', () => {
 
     it('should call the link with the mutation', async () => {
       await client.mutate(mutation)
-      expect(requestHandler).toHaveBeenCalledWith(mutation)
+      expect(requestHandler.mock.calls[0][0]).toBe(mutation)
     })
 
     it('should then dispatch a RECEIVE_MUTATION_RESULT action', async () => {
--- a/packages/cozy-client/src/CozyLink.js
+++ b/packages/cozy-client/src/CozyLink.js
@@ -15,6 +15,7 @@ const toLink = handler =>
 
 const defaultLinkHandler = (operation, result) => {
   if (result) return result
+  else if (operation.execute) return operation.execute()
   else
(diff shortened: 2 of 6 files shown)


ptbrowne:
Can you show an example of how you use this new functionality to implement the `upload` ?

y-lohse:

y-lohse:
Aaah yes, I meant to do that.

Code before:

```
const TIMELINE_MUTATIONS = (query) => ({
  uploadPhoto: async (file, dirPath) => {
    return query.client.upload(file, dirPath)
  },
})
```

But now I want to pass a `dirId`, not `dirPath`:

```
const TIMELINE_MUTATIONS = (query) => ({
  uploadPhoto: async (file, dirId) => {
    return query.client.mutate({
      mutationType: 'UPLOAD_PHOTO',
      execute: () => query.client.collection('io.cozy.files').createFile(file, { dirId })
    })
  }
})
```

ptbrowne:
Why is the upload not a mutation supported by the stack client ?

PR #243, opened 2018-10-31
docs: Reorganize docs

ptbrowne:
Removed duplication
Extracted a mobile guide
Added table of contents through MarkdownTOC (SublimeText plugin)


README.md
```diff
@@ -1,13 +1,24 @@
+> This is the documentation of the current version of cozy-client. If you want to check the old version, [go to the old version](http://github.com/cozy/cozy-client-js) 👵👴.
```
y-lohse: Ok we need to make up our mind about where that disclaimer goes :D https://github.com/cozy/cozy-client/commit/5ae721319b1532ac8008ef4838132d8f4ccf35fb#diff-04c6e90faac2675aa89e2176d2eec7d8

I think it's better at the end — the readme is short and it's still visible at the end, but it's a bit off-putting to see this as the first line.

Crash--: Yep, better to have it at the end. First reflex here is to click on the link. This is what happened to some of our prospects :(

ptbrowne: You're both right :) Thanks

PR #248, opened 2018-11-01
feat: Links receive CozyClient

ptbrowne:
This PR supersedes https://github.com/cozy/cozy-client/pull/247.

The goal of this PR was for a link to be able to inject data in the store of CozyClient. This is necessary for the PouchLink that receives data asynchronously after a synchronization.

Previously, in Banks, the onSync handler was used to manually `setData`.
https://github.com/cozy/cozy-banks/blob/697ac7f8bb328cbc7d83f34145b811f0c7d34831/src/ducks/client/mobile.js#L80

This was done as a temporary solution. Now, this solution is backported into CozyClient. The link receives a reference to the CozyClient at registration time, saves the reference, and call `setData` when it receives new data asynchronously.

The links already received a `StackClient` at registration time but it was wrongly named `client` which caused confusion between CozyClient and CozyStackClient. To clear this confusion, this PR renames all places where a StackClient is used under the name `client` into `stackClient`. This addresses the concerns in https://github.com/cozy/cozy-client/issues/246.

## Discussions

I am not very happy with the name `setData`, I have the feeling that a link should
not directly call a method like this. I could be called `requestSetData`, or `onReceiveData`.

Should the links receive a full blown client or just a subset of its API (`getToken`, `getURL`, `requestSetData`, `refreshToken`) in order to respect the Law of Demeter ? I am not very happy that the links know that there is a stack client in the client they receive.

Should the links instantiate their own stack client when they need it instead of using the one from the cozy client ? This would make the coupling between the links and the client less strong.

### DIFF
--- a/packages/cozy-stack-client/src/__tests__/setup.js
+++ b/packages/cozy-stack-client/src/__tests__/setup.js
@@ -35,6 +35,17 @@ expect.extend({
   }
 })
 
+beforeEach(() => {
+  jest.spyOn(global.console, 'log').mockImplementation(() => {})
+  jest.spyOn(global.console, 'info').mockImplementation(() => {})
+})
+
+afterEach(() => {
+  global.console.log.mockRestore && global.console.log.mockRestore()
+  global.console.info.mockRestore && global.console.info.mockRestore()
+})
+
+
 // In Node v7 unhandled promise rejections will terminate the process
 if (!process.env.LISTENING_TO_UNHANDLED_REJECTION) {
   process.on('unhandledRejection', reason => {
--- a/packages/cozy-stack-client/src/__tests__/setup.js
+++ b/packages/cozy-stack-client/src/__tests__/setup.js
@@ -45,7 +45,6 @@ afterEach(() => {
   global.console.info.mockRestore && global.console.info.mockRestore()
 })
 
-
 // In Node v7 unhandled promise rejections will terminate the process
 if (!process.env.LISTENING_TO_UNHANDLED_REJECTION) {
   process.on('unhandledRejection', reason => {
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -43,6 +43,14 @@ import forEach from 'lodash/forEach'
 
 const ensureArray = arr => (Array.isArray(arr) ? arr : [arr])
 
+const deprecatedHandler = msg => ({
+  get(target, prop) {
+    console.warn(msg);
+    return target[prop];
+  },
+});
+
+
 /**
  * @module CozyClient
  */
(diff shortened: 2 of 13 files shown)


packages/cozy-pouch-link/src/jsonapi.js
```diff
@@ -1,6 +1,7 @@
 export const normalizeDoc = (doc, doctype) => {
   const id = doc._id || doc.id
-  return { id, _id: id, _type: doctype, ...doc }
+  const rev = doc._rev || doc.rev
```
kosssi: When you update a document pouch response include `rev` but the document have already old `_rev` so it's indispensable to change it:
```suggestion
  const rev = doc.rev || doc._rev
```

packages/cozy-pouch-link/src/jsonapi.js
```diff
@@ -1,6 +1,7 @@
...
+  return { id, _id: id, rev, _rev: rev, _type: doctype, ...doc }
```
kosssi: I suggeste to not create the `rev` key but use always `_rev`:
```suggestion
  const normalized = { ...doc, id, _id: id, _rev: rev, _type: doctype }
  delete normalized.rev

  return normalized
```

packages/cozy-pouch-link/src/jsonapi.js
```diff
@@ -1,7 +1,22 @@
 export const normalizeDoc = (doc, doctype) => {
   const id = doc._id || doc.id
-  const rev = doc._rev || doc.rev
-  return { id, _id: id, rev, _rev: rev, _type: doctype, ...doc }
+
+  // PouchDB sends back .rev attribute but we do not want to
+  // keep it on the server.
+  // The rev can be higher than _rev, we keep the highest not
+  // to have any document conflict.
+  const _rev = doc.rev > doc._rev ? doc.rev : doc._rev
```
kosssi: If .rev or ._rev is not defined this throw an error.

packages/cozy-pouch-link/src/jsonapi.js
ptbrowne: Thanks !

packages/cozy-pouch-link/src/jsonapi.js
ptbrowne: Done !

packages/cozy-stack-client/src/__tests__/setup.js
```diff
@@ -35,6 +35,17 @@ expect.extend({
   }
 })
 
+beforeEach(() => {
+  jest.spyOn(global.console, 'log').mockImplementation(() => {})
+  jest.spyOn(global.console, 'info').mockImplementation(() => {})
+})
```
y-lohse: What is the reason behind this change?

ptbrowne: So many console.log polluting the test output. It's better to display them on a test by test basis, with a `console.log.mockRestore()`.

y-lohse: Meh, ok, but we should eventually review why we are writing so much to the console :)

ptbrowne: When I write to the console, it is to have useful debugging imformation. It is wrapped in a `NODE_ENV !== 'production'`.

PR #252, opened 2018-11-07
feat: Add pouch parameters

kosssi:
You can instanciate CozyPouchLink with pouchPlugins and pouchOptions

Example:
```js
const cozyPouchLink = new CozyPouchLink({
  doctypes: offlineDoctypes,
  pouch: {
    plugins: [require('pouchdb-adapter-cordova-sqlite')],
    options: { adapter: 'cordova-sqlite', location: 'default' }
  }
})
```

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -90,6 +90,7 @@ export default class PouchLink extends CozyLink {
       }
     }
     this.pouches = new PouchManager(this.doctypes, {
+      pouch: this.options.pouch,
       getReplicationURL: this.getReplicationURL.bind(this),
       onError: err => this.onSyncError(err),
       onSync: this.handleOnSync.bind(this)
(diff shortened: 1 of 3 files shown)


packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -61,6 +63,10 @@ export default class PouchLink extends CozyLink {
     this.synced = window.localStorage.getItem(LOCALSTORAGE_SYNCED_KEY) || false
   }
 
+  getPouchOptions() {
+    return pickBy(this.options, (value, key) => startsWith(key, 'pouch'))
```
ptbrowne: Instead of prefixing with `pouch`, wouldn't be simpler to do:

```
{
  pouch: {
    options: {},
    plugins: [] 
  }
}
```

kosssi: I change it 👍

PR #253, opened 2018-11-07
fix: ID case typo

y-lohse:
`createDirectory` was called with a `dirID` but was expecting a `dirId`.


y-lohse:

PR #254, opened 2018-11-07
URl encode parameters in pathes

y-lohse:
The problem might be more pervasive than just the `FileCollection`, but at least there we didn't encode some strings in our urls.


y-lohse:

y-lohse:
Just updated the code — the `uri` template literal does the job, it just wasn't called everywhere. I added the missing calls.
I removed the extra tests i had added as I don't think they make much sense.

PR #255, opened 2018-11-07
feat(pouch): Manage sync status by doctype

drazik:
The synchronization status was managed globally (eg set to true when all the doctypes have been synchronized). This lead to issues where some doctypes already synchronized couldn't be returned from Pouch because other doctypes are still synchronizing.

Now, the status is managed for each doctype. So if transactions are synchronizing and accounts already synchronized and we query the accounts, they are fetched from Pouch.

### DIFF
--- a/packages/cozy-pouch-link/package.json
+++ b/packages/cozy-pouch-link/package.json
@@ -18,6 +18,7 @@
   },
   "devDependencies": {
     "babel-cli": "6.26.0",
+    "jest-localstorage-mock": "2.3.0",
     "parcel": "1.10.3",
     "react": "16.4.2",
     "react-dom": "16.4.2"
--- a/packages/cozy-pouch-link/src/PouchManager.js
+++ b/packages/cozy-pouch-link/src/PouchManager.js
@@ -72,6 +72,8 @@ const startReplication = (pouch, getReplicationURL) => {
   return promise
 }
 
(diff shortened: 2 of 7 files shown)


packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -273,4 +279,37 @@ export default class PouchManager {
   getPouch(doctype) {
     return this.pouches[doctype]
   }
+
+  getSyncedStatus() {
+    const item = window.localStorage.getItem(LOCALSTORAGE_SYNCED_KEY)
+
+    // We check if the item in local storage is an array because we previously stored a boolean
+    if (!item || !Array.isArray(item)) {
```
kosssi:
```suggestion
    if (!item || !Array.isArray(JSON.parse(item))) {
```

packages/cozy-pouch-link/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "devDependencies": {
     "babel-cli": "6.26.0",
+    "jest-localstorage-mock": "^2.3.0",
```
kosssi: si tu peux fixé la version sinon c'est renovate qui fera une PR ;)
```suggestion
    "jest-localstorage-mock": "2.3.0",
```

PR #259, opened 2018-11-12
Add DocumentCollection method to index fields

Gara64:
This PR adds a `indexFields` method to force the indexing of arbitrary fields.

Imagine the following set of documents: 
`[{ planet: "jupiter"}, { planet: "earth", life: true}]`

We want to use the following mango query: `selector: {life: { $exists: false }}`. With the current behaviour, this would try to use an index on `life`, which cannot work as none document with a missing `life` field would enter the index.
This PR make it possible by using `.indexFields(['planet'])` to force an index on another field that the ones queried.

It also: 
* Add a test
* Add docs
* Fix doc for subdirectories (see package.json)


packages/cozy-client/src/queries/dsl.js
```diff
@@ -19,21 +38,51 @@ export class QueryDefinition {
...
     this.sort = sort
     this.includes = includes
     this.referenced = referenced
     this.limit = limit
     this.skip = skip
   }
 
+  /**
+   * Query documents with a [mango selector](http://docs.couchdb.org/en/latest/api/database/find.html#find-selectors).
+   * Each field passed in the selector will be indexed, except if the indexField is used.
```
y-lohse:
```suggestion
   * Each field passed in the selector will be indexed, except if the indexField option is used.
```

packages/cozy-stack-client/src/DocumentCollection.spec.js
```diff
@@ -329,6 +329,24 @@ describe('DocumentCollection', () => {
...
+
+    it('should index the specified fields', async () => {
+      const collection = new DocumentCollection('io.cozy.todos', client)
+      await collection.find(
+        { done: { $exists: true } },
+        { indexedFields: ['label'], sort: [{ label: 'desc' }] }
+      )
+      expect(client.fetchJSON).toHaveBeenLastCalledWith(
+        'POST',
+        '/data/io.cozy.todos/_find',
```
y-lohse: I don't think this test really tests that the proper fields are indexed. Shouldn't we check for a call to `'/data/io.cozy.todos/_index'`?

packages/cozy-client/src/queries/dsl.js
```diff
@@ -19,21 +38,51 @@ export class QueryDefinition {
...
+   *
+   * @param {Array} fields The fields to return.
+   * @return {QueryDefinition}  The QueryDefinition object.
+   */
   select(fields) {
     return new QueryDefinition({ ...this.toDefinition(), fields })
   }
 
+  /**
+   * Specify which fields should be indexed. This prevent the automatic indexing of the mango fields.
```
ptbrowne:
```suggestion
   * Specify which fields should be indexed. This prevent the automatic indexing of mango fields.
```

packages/cozy-client/src/queries/optimize.js
```diff
@@ -9,8 +9,8 @@ const isIdQuery = query => query.id || query.ids
 /**
  * Optimize queries on a single doctype
  *
- * @param  {[QueryDefinition]} queries - Queries of a same doctype
- * @return {[QueryDefinition]} Optimized queries
+ * @param  {Array.QueryDefinition} queries - Queries of a same doctype
+ * @return {Array.QueryDefinition} Optimized queries
```
ptbrowne: I didn't know this syntax (Array.Type) and I cannot find it on the JSX doc. I think it should be `QueryDefinition[]`.

http://usejsdoc.org/tags-param.html#multiple-types-and-repeatable-parameters

ptbrowne:
LGTM

packages/cozy-client/src/queries/optimize.js
Gara64: I got confused by this doc : http://usejsdoc.org/tags-type.html, which is for the type definition. It seems your version is correct

packages/cozy-stack-client/src/DocumentCollection.spec.js
Gara64: Yes, actually both are necessary. The `_find` call to ensure the sort doesn't include the label, and the `_index` call to make sure the index is created.

PR #261, opened 2018-11-14
feat: Add authenticate with SafariViewController on mobile

kosssi:
On iOS it's better to use SafariViewController to open authenticate page.

This code comes from https://github.com/cozy/cozy-drive/blob/b033f88209c83c7af7575119634b5e05429d57bc/src/lib/cozy-client/authentication/mobile.js

### DIFF
--- a/packages/cozy-client/src/authentication/mobile.js
+++ b/packages/cozy-client/src/authentication/mobile.js
@@ -1,48 +1,79 @@
 /* global prompt */
 import { REGISTRATION_ABORT } from '../const'
+import { hasInAppBrowserPlugin, hasSafariPlugin } from 'cozy-device-helper'
 
-const hasCordovaPlugin = () => {
-  return (
-    window.cordova !== undefined && window.cordova.InAppBrowser !== undefined
-  )
+const authenticateWithSafari = url => {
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/authentication/mobile.js
```diff
@@ -1,48 +1,83 @@
...
+    }
+
+    inAppBrowser.addEventListener('loadstart', onLoadStart)
+    inAppBrowser.addEventListener('exit', onExit)
+  })
+}
+
+export const authenticateWithCordova = async url => {
+  if (isIOSApp() && (await hasSafariPlugin())) {
+    return authenticateWithSafari(url)
```
Crash--: Why don't we use this plugin also on Android?

kosssi: I copied the code from the old cozy-client https://github.com/cozy/cozy-drive/blob/master/src/lib/cozy-client/authentication/mobile.js#L31 ;)

You think we can activate it ?

kosssi: After discussing on mattermost we decided to activate it on Android

PR #262, opened 2018-11-14
feat(DocCollection): Use _normal_docs route

y-lohse:
When fetching all the documents of a certain type without any criteria, we were using the `_all_docs` route. This route also returns design documents, which we had to remove from the results and would mess up the pagination.

Instead, there's now a new route called `_normal_docs`, which doesn't take into account design docs - so we're using that one instead.
Unfortunately, it doesn't support the `keys` option and the response is slightly different than the `_all_docs` response, but these difference can be abstracted away.


y-lohse:
[(1) Pagination des albums](https://trello.com/c/Ah0aKivw/1138-1-pagination-des-albums)

kosssi:
👍

PR #263, opened 2018-11-14
feat(pouch): Destroy databases when login on another instance

drazik:
We destroy all pouch dbs when the user logouts. But it happens asynchronously and can take several seconds. So during this time, the user can kill the app, aborting this process.

To avoid data leak between instances, this PR introduce two things:

* The Pouch databases are now named after the instance they correspond to
* If we still have references to some databases and we login on another instance, we destroy them

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.spec.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.spec.js
@@ -15,32 +15,38 @@ const mockClient = {
 
 const TODO_DOCTYPE = SCHEMA.todos.doctype
 
-describe('CozyPouchLink', () => {
-  let client, link
+let client, link
 
-  beforeEach(() => {
-    link = new CozyPouchLink({ doctypes: [TODO_DOCTYPE] })
-    client = new CozyClient({
-      ...mockClient,
-      links: [link],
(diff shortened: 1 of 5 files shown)


packages/cozy-pouch-link/src/CozyPouchLink.spec.js
```diff
@@ -203,28 +216,24 @@ describe('CozyPouchLink', () => {
...
     afterEach(() => {
       spy.mockRestore()
     })
 
-    it('should start the replication loop if `options.initialSync` is true', () => {
-      link.options.initialSync = true
-      link.onLogin()
+    it('should start the replication loop if `options.initialSync` is true', async () => {
+      await setup({ initialSync: true }, () => {
+        spy = jest.spyOn(link, 'startReplication').mockReturnValue(jest.fn())
```
ptbrowne: Minor: maybe we don't need linkSetup ?

```
jest.spyOn(CozyPouchLink.prototype, 'startReplciation').mockReturnValue(jest.fn())
await setup({ initialSync: true })
```

drazik: Thanks @ptbrowne.

PR #264, opened 2018-11-16
fix(DocCollection): Correct pagination limits

y-lohse:
I don't understand how this hasn't been found before, but I think I've double checked everything and this really appears to be a bug.

If there are 10 rows in total, and we have a skip of `0` and no limit, we will get 10 rows in the result. This will evaluate to `0 + 10 <= 10`, which is `true`… but of course there are no next pages.


y-lohse:
[(1) Pagination des albums](https://trello.com/c/Ah0aKivw/1138-1-pagination-des-albums)

PR #266, opened 2018-11-19
Docs on the schema

ptbrowne:
Now that the schema is required for each doctype accessed by cozy-client, an update to the docs was required.


ptbrowne:
> Do you think that default doctypes into Cozy-Client is a good idea ?

Yes :)

PR #270, opened 2018-11-21
feat: Save state of login to not start replication twice

kosssi:
In banks, we run `cozyClient.login()` several times without doing it on purpose, I think that cozyClient must know how to manage its login state.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -67,6 +67,7 @@ export default class CozyClient {
 
     this.options = options
     this.idCounter = 1
+    this.isLogged = false
 
     this.createClient()
 
@@ -91,6 +92,12 @@ export default class CozyClient {
   }
 
   async login() {
+    if (this.isLogged) {
+      console.warn(`CozyClient is already logged.`)
+      return
+    }
+
+    this.isLogged = true
     this.registerClientOnLinks()
 
     for (const link of this.links) {
@@ -105,6 +112,12 @@ export default class CozyClient {
   }
 
   async logout() {
+    if (!this.isLogged) {
+      console.warn(`CozyClient isn't logged.`)
+      return
+    }
+
+    this.isLogged = false
     // unregister client on stack
     if (
       this.stackClient.unregister &&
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -91,35 +92,41 @@ export default class CozyClient {
...
   async login() {
-    this.registerClientOnLinks()
-
-    for (const link of this.links) {
-      if (link.onLogin) {
-        try {
-          await link.onLogin()
-        } catch (e) {
-          console.warn(e)
+    if (!this.isLogged) {
```
ptbrowne: nit: `if (this.isLogged) { return }` is clearer IMHO and this way there is no indentation to add.

packages/cozy-client/src/CozyClient.js
```diff
@@ -91,35 +92,41 @@ export default class CozyClient {
...
 
   async logout() {
-    // unregister client on stack
-    if (
-      this.stackClient.unregister &&
-      (!this.stackClient.isRegistered || this.stackClient.isRegistered())
-    ) {
-      await this.stackClient.unregister()
-    }
+    if (this.isLogged) {
```
ptbrowne: Same here.

packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -139,6 +146,11 @@ describe('CozyClient login', () => {
 
     expect(links[0].onLogin).toHaveBeenCalledTimes(1)
     expect(links[2].onLogin).toHaveBeenCalledTimes(1)
+
+    await client.login()
```
ptbrowne: Maybe add a comment explaining why we login twice.

Crash--:
Why this is to CozyClient to manage this use case? It's your responsibility, in your app, to know why the login is called several times, nop?

packages/cozy-client/src/CozyClient.js
ptbrowne: I think we should `console.warn`, it is wrong for the application to try to login twice.

ptbrowne:
> It's your responsibility, in your app, to know why the login is called several times, nop?

I agree.

> Why this is to CozyClient to manage this use case?

At least the framework should not be too dumb and warn the user if there is something wrong.

kosssi:
@Crash-- I agree with you, but the problems that it causes are too important to leave the responsibility to the app. With the warning we do not hide the fact that there is a problem on app ;)

PR #273, opened 2018-11-27
feat(stack-client): Always use all_docs when fetching all docs

drazik:
The usage of `_normal_docs` is very confusing. It doesn't act like a `_all_docs` but like a `_find` (see https://docs.cozy.io/en/cozy-stack/data-system/#list-all-the-documents-alternative). Thus, it is subject to the 100 documents limit from the stack. So we didn't fetch **all** docs when using `_normal_docs`. I think always using `_all_docs` and filter design docs by hand is better than fetching multiple `_normal_docs` to get all docs, because it makes it less confusing (using `all` method fetches all documents).

That's potentially a breaking change and maybe we need another method that uses `_normal_docs` to keep the paginated results ?

Tests are broken, but I want your feedbacks on the proposition itself before fixing it.

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -32,9 +32,7 @@ export default class DocumentCollection {
   async all(options = {}) {
     const { limit, skip = 0, keys } = options
 
-    const isUsingAllDocsRoute = !!keys
-    const route = isUsingAllDocsRoute ? '_all_docs' : '_normal_docs'
-    const url = uri`/data/${this.doctype}/${route}`
+    const url = uri`/data/${this.doctype}/_all_docs`
     const params = {
       include_docs: true,
       limit,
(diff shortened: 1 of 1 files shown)


y-lohse:
I spent some more time investigating : `_all_docs` is indeed not paginated by default, although you can use pagination on it. I wasn't aware this was possible (and tbh I'm still not sure it's desirable?), so apologies for the inconvenience.

> I think always using _all_docs and filter design docs by hand is better

The problem is that this stops working as soon as you use pagination on that route, because the total number of rows reported by the stack will be wrong (as it counts design documents that we would filter out).

I personally don't mind that `all` returns paginated results. It's not less intuitive than for a method like `find`, and at least it's consistent. To be honest I wouldn't even be too surprised of the stack starts paginating these results, as it's surprising that we can essentially load an arbitrary number of documents in memory.

______

I think the best thing to do would be to repair the [`UNSAFE_all`](https://github.com/cozy/cozy-client/blob/deeb5be14dcd53be0487633bf55ced1ca1f0f6ae/packages/cozy-client/src/queries/dsl.js#L119) method — so everything would be paginated by default, but there's a way to explicitly bypass it.

An alternative would be to use the `_all_docs` route if there are no `limit` or `skip` parameters. But then `all` would sometimes return paginated results, and sometimes not, which would be a surprising behaviour.

If we re-instate the `all` method as currently proposed in this PR, we need to prevent using pagination on it, as that is broken. But yeah I don't think that would be a very good solution.

drazik:
Thanks for your feedback @y-lohse. I'm OK for a `UNSAFE_all` method that always uses `_all_docs`, so we don't mess with `all`, at least for now since we need it very quickly to fix an important bug in banks. I close this PR and will open a new one.

y-lohse:
Alright! Sorry again that my previous change cause a bug on your end, I really thought `_all_docs` results were paginated as well. Turns out we just always add limits on drive/photos/contacts :|

PR #274, opened 2018-11-28
fix(DocumentCollection): Return all docs if limit is null

drazik:
There is a `UNSAFE_noLimit` method in cozy-client. It sets the `limit` to `null` on a query definition. The goal is to be able to fetch all documents of a collection at once. But `DocumentCollection::all` was not handling this case properly. It was making a request to `_normal_docs` route, which has a hard limit of 100 set by the stack.

This PR fixes this behaviour by making a request to `_all_docs` if `limit` is intentionnally `null` (keeping the previous behaviour if it is `undefined`, so not breaking). It also filters design docs, which are present only when using `_all_docs` route.

We talked about adding a `UNSAFE_all` previously. But I think this behaviour is intended to be right inside `all`.

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -2,6 +2,7 @@ import { uri, attempt, sleep } from './utils'
 import uniq from 'lodash/uniq'
 import transform from 'lodash/transform'
 import head from 'lodash/head'
+import startsWith from 'lodash/startsWith'
 import * as querystring from './querystring'
 
 export const normalizeDoc = (doc, doctype) => {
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -54,11 +55,16 @@ export default class DocumentCollection {
...
-    return {
-      data: resp.rows.map(row =>
+
+    const data = resp.rows
+      .map(row =>
         normalizeDoc(isUsingAllDocsRoute ? row.doc : row, this.doctype)
-      ),
-      meta: { count: resp.total_rows },
+      )
+      .filter(doc => doc._id && doc._id.indexOf('_design') !== 0)
```
kosssi: c'est un détail mais je n'aime pas `.indexOf !== 0` je préfère utiliser `startsWith` de `lodash` qui est plus compréhensible ;)

kosssi:
```suggestion
      .filter(doc => doc._id && !startsWith(doc._id, '_design'))
```

Crash--: Plus besoin de lodash pour ça :D. Y a aussi le `includes` si besoin :D

drazik: Thanks

PR #275, opened 2018-11-29
feat: Add fetch changes

ptbrowne:
We use this API in our services in Banks.

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.spec.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.spec.js
@@ -449,6 +449,10 @@ describe('DocumentCollection', () => {
       )
     })
 
+    afterEach(() => {
+      client.fetchJSON.mockRestore()
+    })
+
     it('should call the right route with the right payload', async () => {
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/DocumentCollection.spec.js
```diff
@@ -462,4 +466,33 @@ describe('DocumentCollection', () => {
...
+      expect(client.fetchJSON).toHaveBeenCalledWith(
+        'GET',
+        '/data/io.cozy.todos/_changes?include_docs=true&since=my-seq'
+      )
+      expect(changes).toEqual({
+        newLastSeq: 'new-seq',
+        documents: [{ _id: '1', done: false, label: 'Fetch changes' }]
+      })
+    })
+  })
```
Crash--: Can we also add a test for `includeDesign` & `includeDeleted` ?

ptbrowne: Thanks Quentin, I was just lazy ><

PR #278, opened 2018-12-03
Has many query

y-lohse:
When using the `include()` option on a `HasMany` relationship, we use `HasMany.query()` to get a QueryDefinition that allows us to fetch the related documents (for example the tasks related to a todo).

At least in our case, this was not working properly — when constructing the QueryDefinition, we would have an array of ids that was looking like `[{ _id: '123', _type: 'io.cozy.tasks' }]`, instead of a plain `['123']` as we need to execute the query.

Note that we were also looking at `myTodo.tasks` instead of `myTodo.relationships.tasks.data` which iI think is odd too.

Was anyone else using `HasMany` relationships with includes? Or are we just the first to find this bug?

### DIFF
--- a/packages/cozy-client/src/associations/HasMany.js
+++ b/packages/cozy-client/src/associations/HasMany.js
@@ -1,3 +1,4 @@
+import get from 'lodash/get'
 import Association from './Association'
 import { QueryDefinition } from '../queries/dsl'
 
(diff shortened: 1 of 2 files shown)


y-lohse:
[(3) Groupes - supprimer](https://trello.com/c/AUW3POQY/1404-3-groupes-supprimer)

ptbrowne:
> Or are we just the first to find this bug?

Contacts is the first application to really use the HasMany association. In Banks, we use HasManyInPlace, so I am not really surprised to find this bug.

LGTM 🎉 !

y-lohse:
> Since you're adding get() I think you can use it also for addById ?

Well I considered it, but in the case of `addById` we're also recreating the chain as we go along, and `get`doesn't really help with that…

PR #295, opened 2018-12-07
Create with fixed id

gregorylegarec:
Needed to create an agregator account. Agregator account is a single account used by several konnectors to connect on a remote API. An agregator accounts must be created with a fixed `_id`. Example : Budget Insight.

This PR :

* Now rely on _rev existence to determine if the action is an update or a creation
* Call the expected endpoint when created document has fixed _id attribute

__BREAKING CHANGE:__ To be updated, documents now requires a `_rev` attribute.

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -162,11 +162,13 @@ export default class DocumentCollection {
   }
 
   async create({ _id, _type, ...document }) {
-    const resp = await this.stackClient.fetchJSON(
-      'POST',
-      uri`/data/${this.doctype}/`,
-      document
-    )
+    // In case of a fixed id, let's use the dedicated creation endpoint
+    // https://github.com/cozy/cozy-stack/blob/master/docs/data-system.md#create-a-document-with-a-fixed-id
+    const hasFixedId = !!_id
+    const method = hasFixedId ? 'PUT' : 'POST'
+    const endpoint = uri`/data/${this.doctype}/${hasFixedId ? _id : ''}`
+    const resp = await this.stackClient.fetchJSON(method, endpoint, document)
+
     return {
       data: normalizeDoc(resp.data, this.doctype)
     }
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -201,7 +201,7 @@ export default class CozyClient {
    * client.getDocumentSavePlan(baseDoc, relationships)
    */
   getDocumentSavePlan(document, relationships) {
-    const newDocument = !document._id
+    const newDocument = !document._rev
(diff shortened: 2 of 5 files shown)


packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -162,11 +162,11 @@ export default class DocumentCollection {
...
   async create({ _id, _type, ...document }) {
-    const resp = await this.stackClient.fetchJSON(
-      'POST',
-      uri`/data/${this.doctype}/`,
-      document
-    )
+    const hasFixedId = !!_id
+    const method = hasFixedId ? 'PUT' : 'POST'
+    const endpoint = uri`/data/${this.doctype}/${hasFixedId ? _id : ''}`
+    const resp = await this.stackClient.fetchJSON(method, endpoint, document)
```
y-lohse: I'm not convinced by this part of the PR — we already have an `update` method that `PUT`s a document, and making `create` dual-purpose seems risky.
Would it be possible to determine whether we want to create or update a document *before* calling this function? I haven't looked but probably in the StackLink?

gregorylegarec: You're right, I did not realize first but the endpoint is the same than for an update: https://github.com/cozy/cozy-stack/blob/master/docs/data-system.md#create-a-document-with-a-fixed-id

I need to dive into it a little bit again.

gregorylegarec: Ok, so I got the difference now.

[On update](https://github.com/cozy/cozy-stack/blob/master/docs/data-system.md#update-an-existing-document), a document is PUT to the endpoint /data/:type/:id, and the document contains the `_id`.
[On create with forced `_id`](https://github.com/cozy/cozy-stack/blob/master/docs/data-system.md#create-a-document-with-a-fixed-id), the document is not supposed to contain the `_id`. The update method on CozyClient provides the `_id`. But it the document does not already exist, [the stack return an error](https://github.com/cozy/cozy-stack/blob/master/web/data/data.go#L192). So we need to unset the `_id` from the document to create it.

So our solution are:
* Keep checking on `_id` and not `_rev` do check for update or creation, and explicitly unset the `_id` attributes from updated/created document to avoid stack error. The creation with forced id now uses an update of an not existing document.
* Create a third method createWithForcedId into DocumentCollection and manage to call it [here](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/StackLink.js#L57-L58).
* Keep the forced id logic into the create method, like it is done in this PR.

Actually after a little thinking I am not particularly offended by a `create()` method which check if the `_id` is already defined. The fact that in this case the same endpoint as `update()` method is called does not bother me. Maybe this endpoint will change later and we can then juste replace it there.

What do you think ?

PR #296, opened 2018-12-07
Fix Query's mutations prop

gregorylegarec:
It seems that the client parameter was missing and that the code and behaviour were not consistent with the documentation.


gregorylegarec:
@y-lohse I added some test.

I checked on Cozy Banks and Cozy Contacts if this prop was used and it never was. Actually it was not working, or not with the client passed as parameter. So I am not sure if it deserves a breaking change.

PR #305, opened 2018-12-19
Queries with id selector

y-lohse:
So far, we only use the `selector` field of the DSL to compare results with queries, but there are other fields that can contain filtering information, such as `id`, `ids`, `referenced`, and possibly more.

I think some bigger changes are needed in the DSL, but in the meantime this PR fixes the problem for queries with an `id` field.

### DIFF
--- a/packages/cozy-client/src/store/queries.spec.js
+++ b/packages/cozy-client/src/store/queries.spec.js
@@ -39,7 +39,7 @@ describe('queries reducer', () => {
         )
       )
     })
-    it('should update correctly', () => {
+    it('should correctly update', () => {
       applyAction(
         receiveQueryResult('a', {
           data: [TODO_1, TODO_2]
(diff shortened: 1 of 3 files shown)


y-lohse:
[Album photo - pagination](https://trello.com/c/0Z6W0sEN/1511-album-photo-pagination)

packages/cozy-client/src/store/queries.spec.js
```diff
@@ -89,5 +89,25 @@ describe('queries reducer', () => {
...
+          data: [TODO_2]
+        })
+      )
+      expect(state).toMatchSnapshot()
+      applyAction(
+        receiveQueryResult('a', {
+          data: [TODO_3]
+        })
+      )
+      expect(state).toMatchSnapshot()
```
ptbrowne: The inconvenient with `toMatchSnapshot` is that it is hard to understand what is tested. A comment like `b query should be updated now` would help I think.

PR #307, opened 2018-12-19
Add add() method to PermissionCollection

gregorylegarec:
The add method allows to add a permission for a given permissions document, app or konnector.

This method tests the type of the first parameter and call the related endpoint.

It throws an error if this type is unexpected.

Needed in https://github.com/cozy/cozy-home/pull/1041.

### DIFF
--- a/packages/cozy-stack-client/src/PermissionCollection.spec.js
+++ b/packages/cozy-stack-client/src/PermissionCollection.spec.js
@@ -3,6 +3,15 @@ jest.mock('./CozyStackClient')
 import CozyStackClient from './CozyStackClient'
 import PermissionCollection from './PermissionCollection'
 
+const fixtures = {
+  permission: {
+    test: {
+      type: 'io.cozy.todos',
+      verbs: ['GET', 'PUT', 'DELETE']
+    }
+  }
+}
+
 describe('PermissionCollection', () => {
   const client = new CozyStackClient()
   const collection = new PermissionCollection('io.cozy.permissions', client)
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/PermissionCollection.spec.js
```diff
@@ -17,5 +26,84 @@ describe('PermissionCollection', () => {
       await collection.getOwnPermissions()
       expect(client.fetchJSON).toHaveBeenCalledWith('GET', '/permissions/self')
     })
+
+    describe('add', () => {
+      it('fetches expected permissions endpoint', async () => {
```
ptbrowne:
```suggestion
      it('uses the correct endpoint and method', async () => {
```

ptbrowne: J'avais cru que fetch c etait pour dire GET mais en fait c etait juste pour dire fetchJSON

packages/cozy-stack-client/src/PermissionCollection.spec.js
```diff
@@ -17,5 +26,84 @@ describe('PermissionCollection', () => {
...
+              type: 'io.cozy.permissions',
+              attributes: {
+                permissions: fixtures.permission
+              }
+            }
+          }
+        )
+      })
+
+      it('throws an error if document do not have expected doctype', () => {
```
ptbrowne:
```suggestion
      it('throws an error if document does not have expected doctype', () => {
```

PR #308, opened 2018-12-20
fix: Guard against rows containing errors

y-lohse:
If we use the `keys` option on the `_all_docs` route and specify an invalid key, the row has an `error` field but no `id` or `doc`. This caused the normalisation function to crash.

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -56,13 +56,18 @@ export default class DocumentCollection {
       throw error
     }
     let data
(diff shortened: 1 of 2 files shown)


y-lohse:
[(0.5j) Partages - erreur connexion internet (bis)](https://trello.com/c/3FddJAoU/1499-05j-partages-erreur-connexion-internet-bis)

Crash--:
So we can remove this check ` doc.doc !== null`. Right?

y-lohse:
No, I'm pretty sure that was a different case, where there was no `error`.

PR #312, opened 2018-12-21
docs: Added general schema

y-lohse:
Here's a preview : https://whimsical.co/6AJnnS7v5ePRcuKZysPAF5

graph


docs/schema.md
```diff
@@ -0,0 +1,10 @@
+### cozy-client schema
+
+The schema can also be accessed [online here](https://whimsical.co/6AJnnS7v5ePRcuKZysPAF5).
+It's created on https://whimsical.co/ — if you need to edit it, ask for an invite from one of the projects maintainers.
```
ptbrowne: I think architecture instead of schema would be better. It would be good to insert the image also so that it is visible. Is there a link to here somewhere ?

ptbrowne:
:shipit:

nono:
@y-lohse I think you can merge this PR

PR #324, opened 2019-01-09
feat(cozy-client): Deprecate connect HOC

### LINKED ISSUE #317: [cozy-client] Should we drop connect HOC in favor of queryConnect?
We have two HOCs to connect a react component to a query: `connect` and `queryConnect`. The later can connect multiple queries to a single react component, where the former can connect only one query. So `queryConnect` looks like a enhanced version of `connect`. I think we should keep this one and deprecate then drop `connect`. What do you think?

drazik:
Add a deprecation message and update docs.

Closes #317

### DIFF
--- a/packages/cozy-client/src/connect.jsx
+++ b/packages/cozy-client/src/connect.jsx
@@ -4,49 +4,55 @@ import PropTypes from 'prop-types'
 
 import { getQueryFromState } from './store'
 
-const connect = (query, options = {}) => WrappedComponent => {
-  const wrappedDisplayName =
-    WrappedComponent.displayName || WrappedComponent.name || 'Component'
-
-  const mapStateToProps = (state, ownProps) => ({
-    ...getQueryFromState(state, ownProps.queryId)
-  })
-
-  const ConnectedWrappedComponent = reduxConnect(mapStateToProps)(
-    WrappedComponent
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/connect.jsx
```diff
@@ -4,49 +4,57 @@ import PropTypes from 'prop-types'
...
-  )
+    const mapStateToProps = (state, ownProps) => ({
+      ...getQueryFromState(state, ownProps.queryId)
+    })
 
-  class Wrapper extends Component {
-    static contextTypes = {
-      client: PropTypes.object
-    }
+    console.log('before')
```
Crash--: to be removed?

packages/cozy-client/src/connect.jsx
```diff
@@ -4,49 +4,57 @@ import PropTypes from 'prop-types'
...
+    const ConnectedWrappedComponent = reduxConnect(mapStateToProps)(
+      WrappedComponent
+    )
+    console.log('after')
```
Crash--: to be removed?

packages/cozy-client/src/connect.jsx
drazik: Yep, test remainder :) thanks

PR #330, opened 2019-01-14
fix(CozyPouchLink): Re-add allDocs function

kosssi:
When I added the default query `{_id: {$ exists: true}}` I thought it
worked for allDocs but it was not. Now we can use both `allDocs` or
`find` functions.


packages/cozy-pouch-link/src/helpers.spec.js
```diff
@@ -10,53 +10,88 @@ PouchDB.plugin(adapter)
...
+    it('should update total rows number', () => {
+      const filteredResponse = withoutDesignDocuments(response)
+      expect(filteredResponse.total_rows).toEqual(1)
     })
-
-    it('should find 20 docs with limit', async () => {
-      await insertData(db, 200)
-      const data = await find(db, { limit: 20 })
-      expect(data.docs).toHaveLength(20)
+    it('should not mute response', () => {
```
ptbrowne:
```suggestion
    it('should not mutate response', () => {
```

PR #341, opened 2019-01-23
fix(docs): Broken link to DocumentCollection

cedricmessiant:
`DocumentCollection` has moved to `cozy-stack-client.md`


cedricmessiant:
In fact, it seems that I have to fix the link in `CozyClient.js`: `{@link DocumentCollection}`

cedricmessiant:
@ptbrowne do you know if it's possible to add a link to `cozy-stack-client/src/DocumentCollection.js` from here?

PR #342, opened 2019-01-23
fix: HasManyFiles relationships

y-lohse:
This is an alternative solution to #337 — we make sure the `relationships` attribute is kept in documents in the redux store when they are updated.

### DIFF
--- a/packages/cozy-client/src/associations/helpers.js
+++ b/packages/cozy-client/src/associations/helpers.js
@@ -1,5 +1,6 @@
 import pick from 'lodash/pick'
 import pickBy from 'lodash/pickBy'
+import isEqual from 'lodash/isEqual'
 import Association from './Association'
 import HasOne from './HasOne'
 import HasOneInPlace from './HasOneInPlace'
@@ -21,7 +22,10 @@ export const responseToRelationship = response =>
   })
 
 const attachRelationship = (doc, relationships) => {
-  if (doc.relationships) {
+  if (
+    doc.relationships &&
+    isEqual(Object.keys(doc.relationships), Object.keys(relationships))
+  ) {
     return doc
   }
   return {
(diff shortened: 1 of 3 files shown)


y-lohse:
[(1) Relations de fichiers dans cozy-client](https://trello.com/c/7IkEB6u4/1545-1-relations-de-fichiers-dans-cozy-client)

packages/cozy-client/src/associations/helpers.js
```diff
@@ -21,7 +22,10 @@ export const responseToRelationship = response =>
   })
 
 const attachRelationship = (doc, relationships) => {
-  if (doc.relationships) {
+  if (
+    doc.relationships &&
+    isEqual(Object.keys(doc.relationships), Object.keys(relationships))
```
ptbrowne: I don't understand this ?

y-lohse: With the new dehydratation, some documents look like this:

```
{
  _id: 1,
  title: lol,
  relationships: {}
}
```

And we try to attach relationships, to it, for example:

```
{
  tasks: { data: […] }
}
```

This fails, because the original doc already has a `relationships` property… but it's incomplete.

I can add tests to this function, but I'm not even sure why the condition is there in the first place…

ptbrowne:
> I don't know if I prefer to create a special case like in this PR where relationships is empty or having to store in the DB the relationships :/

To me, storing the relationships in the DB because they disappear during a mutation is cleary a bug from the library, and I do not think it is a special case of the files. The relationship attribute is to me a `reserved` attribute of the library, it is normal to have special cases for it. WDYT ?

y-lohse:
Merging this, since it's getting a bit urgent on our side and we seem to agree this is the most acceptable solution.

PR #351, opened 2019-01-28
Attaching relationships to objects

y-lohse:
When a `query` uses `includes(['some_relationship'])`, we will fetch the included data and inject it into the main response. This is especially useful for `HasManyFiles` relationships where the main document doesn't contain the relationship info.
But we also have to be careful not to destroy any existing data in the `relationships` field.

This PR makes sure we don't accidentally override some relationship data, and adds tests for all situations I know of.

### DIFF
--- a/packages/cozy-client/src/associations/helpers.js
+++ b/packages/cozy-client/src/associations/helpers.js
@@ -22,16 +22,12 @@ export const responseToRelationship = response =>
   })
 
 const attachRelationship = (doc, relationships) => {
-  if (
-    doc.relationships &&
-    relationships &&
-    isEqual(Object.keys(doc.relationships), Object.keys(relationships))
-  ) {
-    return doc
-  }
   return {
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/associations/helpers.js
```diff
@@ -22,15 +22,12 @@ export const responseToRelationship = response =>
...
-    isEqual(Object.keys(doc.relationships), Object.keys(relationships))
-  ) {
-    return doc
-  }
   return {
     ...doc,
-    relationships: relationships || {}
+    relationships: {
+      ...doc.relationships,
+      ...relationships
```
ptbrowne: When destructuring, please make sure that the object exists. A test for the case where `relationships` is undefined should catch the failure ;)

ptbrowne:
Thanks @y-lohse for the test :tada

PR #352, opened 2019-01-28
fix(client): Handle undefined relationships

drazik:
We were trying to apply `Object.keys` on something that could be undefined. I just added a check to avoid a `can't convert undefined to object` error.

### DIFF
--- a/packages/cozy-client/src/associations/helpers.js
+++ b/packages/cozy-client/src/associations/helpers.js
@@ -24,6 +24,7 @@ export const responseToRelationship = response =>
 const attachRelationship = (doc, relationships) => {
   if (
     doc.relationships &&
+    relationships &&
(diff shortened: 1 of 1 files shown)


drazik:
Merging, as we saw it live on Zoom with @ptbrowne

Crash--:
Hmm too fast ? ^^ https://github.com/cozy/cozy-client/pull/351

y-lohse:
Yep, the problem with this fix is that if the document already has some data in `relationships`, it will be overridden. That's the case for `io.cozy.files` for example.

ptbrowne:
> Yep, the problem with this fix is that if the document already has some data in relationships, it will be overridden. 

This is not the same fix as yours. Here the application was crashing because `relationships` was undefined and `Object.keys` did not work on it.

PR #356, opened 2019-02-04
refactor(client): Use properId in all store slices

drazik:
We had a bug caused by a Cozy on which some tests were made using ACH, resulting in documents having different `_id` but same `id`. This can't happen on real instances, but it showed and inconsistency in cozy client: sometimes we were accessing document id via `._id`, and sometimes with a `properId` helper. In the previous case, we had some ids in the `queries` slice that we didn't have in the `document` slice, so the query returned some `null` documents.

I think that even if the bug is not a "real" one, we should be consistent, so I used the `properId` helper in both slices.

This PR is also here for discussion. I propose to use `properId` everywhere but if you think there is a better way, please tell me.


y-lohse:
Personally, i think using `id` instead of `_id` was a mistake that has confusion, extra complexity in our code and in the worst cases, actual bugs. All of this for no actual benefits, as far as I can tell.

For me, `_id` is where document ids are stored, period. I (kinda strongly) feel like we should gradually move away from using `id` instead of `_id`.

(👍 on making things consistent in this PR though)

ptbrowne:
I agree with Yannick. I think it's a good first step to use properId everywhere, it will simplify migrations. I think the next step is to deactivate .id at build time with an environment variable ?

```
const properId = doc => {
  if (process.env.LEGACY_ID) {
    return doc.id || doc._id
  } else {
    return doc._id
  }
}
```

WDYT ?

y-lohse:
I would think that using `properId` where we were using `id` is a good step and will simplify migrations. But i'm not sure it's really useful in places where we're already using `_id` — isn't that taking a step back, only to eventually end up with the same code?

ptbrowne:
It leads to bug currently when some documents have a id different from their _id.

y-lohse:
But can't we fix that bug by changing the code that uses `id` instead?

ptbrowne:
> But can't we fix that bug by changing the code that uses id instead?

We could but I think it would be potentially a breaking change. I would prefer to go slowly on this one, I fear the id/_id differences are subtle and would sometimes break unexpectedly. This is why I would personally prefer to go the slow way, and first fix the bug, then deprecate the id.

y-lohse:
Alright! Clearly I don't have a good understanding of that bug, so if you think the safest way to go forward is to use `properId` I'm ok with it 👍

packages/cozy-client/src/store/helpers.js
```diff
@@ -0,0 +1,3 @@
+export const properId = doc => {
+  return doc.id || doc._id
```
kosssi: il ne faut pas écrire l'inverse ? return doc._id qui est le cas le plus commun ?

drazik: Le but là est juste d'harmoniser les endroits où on accède à l'ID d'un document quand on manipule le store redux, afin de ne pas nous retrouver avec des différences entre les ids stockés dans les queries et ceux des documents. Je suis plutôt pour garder le helper qui fonctionnait déjà tel quel, du coup.

PR #359, opened 2019-02-05
Allow multiple mutations as parameters for withMutations()

gregorylegarec:
Useful for passing mutations from different modules.

### DIFF
--- a/packages/cozy-client/src/withMutations.jsx
+++ b/packages/cozy-client/src/withMutations.jsx
@@ -1,7 +1,17 @@
+import merge from 'lodash/merge'
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
-const withMutations = (mutations = {}) => WrappedComponent => {
+/**
+ * @function
+ * @description HOC to provide mutations to components. Needs client in context
+ * or as prop.
+ * @param  {function} mutations One ore more mutations, which are function
+ * taking CozyClient as parameter and returning an object containing one or
+ * more mutations as attributes.
+ * @return {function} - Component that will receive mutations as props
+ */
+const withMutations = (...mutations) => WrappedComponent => {
   const wrappedDisplayName =
     WrappedComponent.displayName || WrappedComponent.name || 'Component'
 
@@ -24,9 +34,14 @@ const withMutations = (mutations = {}) => WrappedComponent => {
           saveDocument: client.save.bind(client),
           deleteDocument: client.destroy.bind(client)
         },
-        ...(typeof mutations === 'function'
-          ? mutations(client, props)
-          : mutations)
+        ...merge(
+          ...mutations.map(
+            mutations =>
+              typeof mutations === 'function'
+                ? mutations(client, props)
+                : mutations
+          )
+        )
       }
     }
 
--- a/packages/cozy-client/src/withMutations.spec.jsx
+++ b/packages/cozy-client/src/withMutations.spec.jsx
@@ -1,16 +1,15 @@
 import React from 'react'
 import { shallow } from 'enzyme'
 
-import CozyClient from '../CozyClient'
-import CozyLink from '../CozyLink'
-import withMutations from '../withMutations'
+import withMutations from './withMutations'
 
 describe('withMutations', () => {
   const clientMock = {
     create: jest.fn(),
     destroy: jest.fn(),
     save: jest.fn(),
-    mutate: jest.fn()
+    mutate: jest.fn(),
+    mutateAsWell: jest.fn()
   }
 
   afterEach(() => {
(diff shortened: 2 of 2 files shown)


packages/cozy-client/src/withMutations.jsx
```diff
@@ -24,9 +33,16 @@ const withMutations = (mutations = {}) => WrappedComponent => {
...
+        ...reduce(
+          mutations.map(
+            mutations =>
+              typeof mutations === 'function'
+                ? mutations(client, props)
+                : mutations
+          ),
+          merge,
+          {}
+        )
```
ptbrowne: It seems merge can be used directly here.

```
merge(...mutations.map(
            mutations =>
              typeof mutations === 'function'
                ? mutations(client, props)
                : mutations
          ))
```

---

On another level, I wonder if this merging should be the responsibility of the library ? There can be the risk of overwriting :/ It will not be understandable by the user :/

packages/cozy-client/src/withMutations.jsx
```diff
@@ -1,7 +1,16 @@
...
+import reduce from 'lodash/reduce'
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
-const withMutations = (mutations = {}) => WrappedComponent => {
+/**
+ * @function
+ * @description HOC to provide mutations to components. Needs client in context
+ * or as prop.
+ * @param  {Component} Component - wrapped component
```
y-lohse: I think the interesting param to document here is `mutations` and not `WrappedComponent `, no?

packages/cozy-client/src/withMutations.jsx
```diff
@@ -24,9 +33,16 @@ const withMutations = (mutations = {}) => WrappedComponent => {
           saveDocument: client.save.bind(client),
           deleteDocument: client.destroy.bind(client)
         },
-        ...(typeof mutations === 'function'
-          ? mutations(client, props)
-          : mutations)
+        ...reduce(
+          mutations.map(
```
y-lohse: Since passing an object to `withMutations` would no longer work, isn't this a breaking change? Or am I missing something?

packages/cozy-client/src/withMutations.jsx
y-lohse: I'd be ok with having this in the library. While I haven't had the need to merge mutations, it seems quite legitimate and I wouldn't be surprised to need this as well some day. And as a user, if I specify multiple mutations, it seems expected that mutations with the same name would en up overriding each other in the `props`.

packages/cozy-client/src/withMutations.jsx
ptbrowne: I thought the same but in fact, with the ... syntax, it will work since mutations here is all the mutations and they are merge, you are not passing an array to withMutations, only allowing to pass multiple mutations to be passed as parameters.

packages/cozy-client/src/__tests__/withMutations.spec.jsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
```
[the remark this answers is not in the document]
gregorylegarec: I am not sure. All tests in Cozy-Client are under the `/packages/*/src/__tests__/` folders.

packages/cozy-client/src/withMutations.jsx
gregorylegarec: Oops you're totally right 👍

packages/cozy-client/src/__tests__/withMutations.spec.jsx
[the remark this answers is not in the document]
ptbrowne: Not all tests are under the `__tests__`, there's actually a mixture :

```
./src/helpers.spec.js
./src/associations/helpers.spec.js
./src/associations/HasMany.spec.js
./src/associations/HasOne.spec.js
./src/associations/HasManyFiles.spec.js
./src/associations/HasManyInPlace.spec.js
./src/CozyLink.spec.js
./src/queries/optimize.spec.js
./src/queries/dsl.spec.js
./src/Query.spec.jsx
./src/CozyClient.spec.js
./src/hoc.spec.js
./src/Schema.spec.js
./src/store/documents.spec.js
./src/store/queries.spec.js
./src/__tests__/connect.spec.jsx
./src/__tests__/Provider.spec.jsx
./src/__tests__/StackLink.spec.js
./src/__tests__/associations.spec.js
./src/__tests__/withMutation.spec.jsx
./src/__tests__/ObservableQuery.spec.js
./src/__tests__/store.spec.js
```

I prefer the spec next to the file way as it is easier to check the spec along the file and you do not have to recreate the folder structure, it is also easier when you are moving file since you do not have to remember to move the fixture.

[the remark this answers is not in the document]
gregorylegarec: Ok for me. The flow we recenlty identified with @CPatchane on Cozy-Harvest-Lib is that test file are also copyied in the `dist` folder during build. In Cozy-Harvest-Lib, we are using the `--copy-files` options to copy the locales files into the `dist` folder.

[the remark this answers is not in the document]
gregorylegarec: Done 👌

PR #364, opened 2019-02-15
feat: getIcon from stack or registry for apps and konnectors

ptbrowne:
- Get icon content (need to XHR to send cookies on mobile)
- Return URL with URL.createObjectURL

If icon type cannot be determined or is not available (ios10), the app manifest is downloaded and the mime type is determined from the icon filename.

### Usage

```
stackClient.getIcon({ type: 'konnector', slug: 'caissedepargne1' })
stackClient.getIcon({ type: 'app', slug: 'drive' })
stackClient.getIcon({ type: 'app', slug: 'drive', priority: 'registry' })
```

### DIFF
--- a/packages/cozy-stack-client/src/getIconURL.js
+++ b/packages/cozy-stack-client/src/getIconURL.js
@@ -1,0 +1,115 @@
+import memoize from './memoize'
+
+const mimeTypes = {
+  gif: 'image/gif',
+  ico: 'image/vnd.microsoft.icon',
+  jpeg: 'image/jpeg',
+  jpg: 'image/jpeg',
+  png: 'image/png',
+  svg: 'image/svg+xml'
+}
+
+const fetchOptions = stackClient => {
+  return {
+    headers: {
+      Authorization: 'Bearer ' + stackClient.token
+    },
+    credentials: true
+  }
+}
+
+const getIconExtensionFromApp = app => {
+  if (!app.icon) {
+    throw new Error(
+      `${app.name}: Cannot detect icon mime type since app has no icon`
+    )
+  }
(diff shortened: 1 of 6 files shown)


packages/cozy-stack-client/src/getIconURL.js
```diff
@@ -0,0 +1,126 @@
...
+    } catch (e) {
+      err = e
+    }
+  }
+  throw err
+}
+
+const _getIconURL = async (stackClient, type, slug, appData) => {
+  const fetchOpts = fetchOptions(stackClient)
+  const resp = await fallbacks(
```
gregorylegarec: If I understand well here we are always trying to get the icon from an installed app first. It means in the Cozy-Store, that we will have more than 100 404 errors at every load.

As we have `appData` as a parameter and as thanks to it we are able to differenciate installed apps from registry, I think we should use `appData` if it is passed.

Another way could be to fetch first registry icons, but the icons can differ between registry and stack (when a newer registry version embed a different icon for example)

ptbrowne: > e are able to differenciate installed apps from registry, I think we should use `appData` if it is passed.

How do you differentiate between registry app and installed app ?

ptbrowne: I added a priority option (so I changed to a option parameters) so that it is possible to indicate that you prefer the registry first.

ptbrowne: I will not add this to AppCollection at this point. I find that instantiating a Collection to get this method is a bit cumbersome.

Issue #363, opened 2019-02-12
Adding cozy metadata automatically


ptbrowne:
Need: each document should have the name of the app creating /updating the document. The schema could store general metadata : 

- name of the app creating/updating the document

`CozyClient` seems to be the candidate to add this metadata to the mutation since this metadata needs to be provided to all the links. I would check `requestMutation` to see if it is possible to add cozyMetadata.

ptbrowne:
Fixed by #365 🎉

PR #365, opened 2019-02-18
Automatic metadata

y-lohse:
Implements #363 — automatically inserting and updating `cozyMetadata`. I will add the relevant documentation after the first round of review.

Basically: if the app provides a `cozyMetadata` object in the schema, the keys will be added / updated to the documents. The schema also contains the values to use and when to update the various metadata fields.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -32,6 +32,8 @@ import flatten from 'lodash/flatten'
 import uniqBy from 'lodash/uniqBy'
 import zip from 'lodash/zip'
 import forEach from 'lodash/forEach'
+import get from 'lodash/get'
+import set from 'lodash/set'
 
 const ensureArray = arr => (Array.isArray(arr) ? arr : [arr])
 
@@ -41,6 +43,9 @@ const deprecatedHandler = msg => ({
     return target[prop]
   }
 })
+
+const TRIGGER_CREATION = 'creation'
+const TRIGGER_UPDATE = 'update'
 
 /**
  * Responsible for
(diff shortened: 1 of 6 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -187,6 +189,49 @@ class CozyClient {
...
+      ...document
+    }
+    const metadata = Object.entries(schema.cozyMetadata)
+
+    for (const [
+      metadataName,
+      { trigger, value, useCurrentDate = false }
+    ] of metadata) {
+      const shouldSetMetadata =
+        options.event === 'creation' ||
```
cedricmessiant: maybe we could use constants for event identifiers ?

Crash--:
Can we add a note somewhere in our documentation to explain how to use it in our apps?

packages/cozy-client/src/CozyClient.js
```diff
@@ -187,6 +192,49 @@ class CozyClient {
...
+      if (shouldAccumulateValues) {
+        const prevValue = get(document, `cozyMetadata.${metadataName}`, [])
+        newValue = uniqBy([...prevValue, ...value])
+      } else if (useCurrentDate) {
+        newValue = new Date().toISOString()
+      } else {
+        newValue = get(document, `cozyMetadata.${metadataName}`, value)
+      }
+
+      set(enriched, `cozyMetadata.${metadataName}`, newValue)
```
drazik: Nit: since this string is used 3 times, you can store it in a `const path = `cozyMetadata.${metadataName}` for example.

PR #370, opened 2019-02-21
fix: Support $gt null in query selectors

y-lohse:
Ok, so:

- A common practice when working with couchdb is to include a `{ $gt: null }` selector — telling couch to use that field as an index, but not using it as a real filter, since almost any value passes the `{ $gt: null }` filter. We do this every now and then in our queries and [it's all over their docs](http://docs.couchdb.org/en/latest/api/database/find.html#find-selectors).
- When documents in the cozy-client store change, we run the updated documents against the selectors of the `queries` to know which ones to update.
- We use [sift.js](https://github.com/crcn/sift.js) to run the selectors, but contrary to couchdb, sift returns no documents when using `{ $gt: null }` .
- Sift is mirroring mongodb behaviour, so it's not a bug on their side. [It might be a bug in mongo](https://stackoverflow.com/questions/29835829/mongodb-comparison-operators-with-null), but better not hold our breath on that.

Sucks for us. Since I don't want to fork / reimplement sift, I looked for a workaround.

[Sift supports custom operators](https://github.com/crcn/sift.js#custom-expressions). We can't override `$gt`, so I created a new `$gtnull` operator that behaves like couchdb.  
Before running the selector function, I convert all `$gt: null` to `$gtnull: null` in the selector, so the user doesn't need to worry about this quirk.

Last thing to note, I actually *downgraded* sift to v6.0.0:

- sift v6 supports custom operators through `sift.use`
- sift v7 (our current version) does not support custom operators at all
- sift v8 (currently in beta) has a new API for custom operators as mentioned in their readme, but has [a change we really don't want](https://github.com/crcn/sift.js/issues/117)

So I'm afraid we'll be stuck with sift v6 for the foreseeable future.

### DIFF
--- a/packages/cozy-client/package.json
+++ b/packages/cozy-client/package.json
@@ -19,7 +19,7 @@
     "react-redux": "5.0.7",
     "redux": "3.7.2",
     "redux-thunk": "2.3.0",
-    "sift": "7.0.1"
+    "sift": "6.0.0"
   },
   "scripts": {
     "build": "../../bin/build",
--- a/packages/cozy-client/src/store/__snapshots__/queries.spec.js.snap
+++ b/packages/cozy-client/src/store/__snapshots__/queries.spec.js.snap
@@ -59,6 +59,66 @@ Object {
 }
 `;
 
+exports[`queries reducer updates should correctly update a query with a $gt: null selector 1`] = `
+Object {
+  "a": Object {
+    "count": 1,
+    "data": Array [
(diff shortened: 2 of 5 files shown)


y-lohse:
[(2) Synchro Google - suppression(s) Cozy->Google [2 pers.]](https://trello.com/c/7IXzLsRe/1535-2-synchro-google-suppressions-cozy-google-2-pers)

PR #371, opened 2019-02-21
feat: Saner internal API for stackClient/OAuthClient

### LINKED ISSUE #368: [stackClient] Error prone credentials parameter in fetch
The credentials option of `fetch` is error prone IMHO. You might think it is equivalent to the `credentials` option of `fetch` (the global) and that you have to pass `"include"` to send the cookies but no, this option is passed as the value of the Authorization header. So if you do 

`stackClient.fetch('GET', 'route', { credentials: 'include' })` you end up with `"include"` in your `Authorization` header.

I was bitten by this and the fix was simply to remove the options https://github.com/cozy/cozy-client/pull/367 but I think it is very confusing.

I want to remove this part of the code but want to check if this behavior is relied on by anybody.

Relevant part of the code :

https://github.com/cozy/cozy-client/blob/da41842a3becc8efc36af59332d4d0218c5aecb5/packages/cozy-stack-client/src/CozyStackClient.js#L72-L78

ptbrowne:
Fix #368

- Deprecate CozyStackClient::setCredentials in favor of setToken
- Deprecate CozyStackClient::getCredentials in favor of getToken
- Put credentials='include' as `fetch` option even if the Authorization header
is not set.
- Override the default Authorization header via headers.Authorization

### DIFF
--- a/packages/cozy-stack-client/src/CozyStackClient.js
+++ b/packages/cozy-stack-client/src/CozyStackClient.js
@@ -6,6 +6,7 @@ import SharingCollection from './SharingCollection'
 import PermissionCollection from './PermissionCollection'
 import TriggerCollection, { TRIGGERS_DOCTYPE } from './TriggerCollection'
 import getIconURL from './getIconURL'
+import logDeprecate from './logDeprecate'
 
 const normalizeUri = uri => {
   if (uri === null) return null
@@ -22,7 +23,7 @@ const normalizeUri = uri => {
 class CozyStackClient {
   constructor({ token, uri = '' }) {
     this.setUri(uri)
-    this.setCredentials(token)
+    this.setToken(token)
   }
 
   /**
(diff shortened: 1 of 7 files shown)


packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -65,8 +65,9 @@ class CozyStackClient {
    */
   async fetch(method, path, body, opts = {}) {
     const options = { ...opts }
+    const headers = { ...(opts.headers || {}) }
```
drazik: Nit: spreading `undefined` works, so doing

```js
const headers = { ...opt.headers }
```

works even if `opt.headers` is undefined, in which case `headers` will be an empty object:

```
const opts = {}
undefined
const headers = { ...opts.headers }
undefined
headers
Object {  }
```

packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -63,22 +64,22 @@ class CozyStackClient {
...
-      // the option credentials:include tells fetch to include the cookies in the
-      // request even for cross-origin requests
-      options.credentials = 'include'
+    if (!headers.Authorization) {
+      headers.Authorization = this.getAuthorizationHeader()
     }
 
+    // the option credentials:include tells fetch to include the cookies in the
+    // request even for cross-origin requests
+    options.credentials = 'include'
```
edas: Do we need this ? as we are using the Authorization header, the cookie is probably not necessary anyways.

ptbrowne: Some API from the stack necessitate to have a session cookie. Until recently it was for example necessary for the registry icons. I agree that this may not be suitable but we would need to do an audit of the stack routes to know this. Maybe it is needed for the auth routes.

I think we should tackle this in another issue.

Issue #109, opened 2018-08-24
[stack-client] The database used for mime-types is very big


ptbrowne:
Do we need such a big mime type database ?

@goldoraf ?

![image](https://user-images.githubusercontent.com/465582/44577536-ae3b0480-a791-11e8-8c4d-19f5c6ca4958.png)

goldoraf:
Well, yes and no. I had to add this mimetype db because some files' mimetypes are not correctly detected by browsers, especially when drag n' dropping a folder for upload. That leaves us having to guess the mimetype from the file extension, which is the purpose of this lib. But I agree this db is way too large, we could probably build our own with only the more common file types (it'll take some time though). And maybe `cozy-client` isn't the right place to deal with this, we could do it in Drive instead.

ptbrowne:
Mozilla has a pretty good list of mime-types https://developer.mozilla.org/fr/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types

y-lohse:
Hard to say — it looks good, but off the top of my head it lacks HEIC and HEIF types for example.

I don't mind giving this a shot as long as we don't have to do much maintenance on it. If in a couple of month we realize we have to frequently update this list then a reversal might be in order. Does that sound ok ?

ptbrowne:
Yes it sounds OK ;)

ptbrowne:
I am closing and we will reopen if we have problems.

PR #373, opened 2019-02-26
feat: Use a lighter mime library

nono:
Mime-types depends of mime-db, which is mostly a 160kb JSON file. It is heavy, as it is embedded in most of our webapps. Mime is a lighter alternative: https://www.npmjs.com/package/mime.

Please note that I have made this change without testing anything! Be careful before merging it.

### DIFF
--- a/packages/cozy-stack-client/package.json
+++ b/packages/cozy-stack-client/package.json
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "detect-node": "2.0.4",
-    "mime-types": "2.1.24",
+    "mime": "2.4.0",
     "qs": "6.7.0"
   },
   "scripts": {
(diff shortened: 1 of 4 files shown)


Crash--:
For reviewers see https://github.com/cozy/cozy-client/issues/109 

We need to test it a lot on drive specially when drag and dropping folders 

@nono this commit shouldn’t be “chore” since it changes the behavior of the released code. (Also sementic release will not create a new release). Fix / feat or refactor sounds preferable.

ptbrowne:
In case this DB is not sufficent for Drive, there should be a way to override the default mime library used in cozy-client with a bigger one (used only in Drive).

ptbrowne:
We need input from Drive developers. @y-lohse @Crash-- ? In the meantime : https://github.com/cozy/cozy-client/pull/383

y-lohse:
Before merging this, we want to add some tests the `getFileTypeFromName` function to make sure it works for the main formats we support in drive (most notably the problematic ones like HEIC or Open Office).

ptbrowne:
Ping :)

Crash--:
@nono Can you rebase and merge this PR? :)

PR #376, opened 2019-02-28
feat(OAuthClient): Add resetClient

Crash--:
With the new onboarding process (with deeplink, access_code and state), we're not using "standard OAuth" since the client (aka our app) is not registering the OAuth Client anymore. This work is done by the stack. 

In order to manage the logout process correctly with this new onboarding, we have to reset manually the OAuth Client. 

We can do that manually in each app, or we can provide a small method in CozyClient.

### DIFF
--- a/packages/cozy-stack-client/src/OAuthClient.js
+++ b/packages/cozy-stack-client/src/OAuthClient.js
@@ -389,6 +389,15 @@ class OAuthClient extends CozyStackClient {
     this.oauthOptions.clientID = ''
   }
 
+  /**
+   * Reset the current OAuth client
+   */
+  resetClient() {
+    this.resetClientId()
+    this.setUri(null)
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -389,6 +389,15 @@ class OAuthClient extends CozyStackClient {
     this.oauthOptions.clientID = ''
   }
 
+  /**
+   * Reset the current OAuth client
+   */
+  resetClient() {
+    this.resetClientId()
+    this.setUri('')
```
ptbrowne: Nit: I prefer null in this cases.

Crash--: I had to edit `normalizeUri`. Is it better now?

PR #380, opened 2019-03-06
feat: Don't use client.schema to add/update cozyMetadata, 

cedricmessiant:
- don't use client.schema to add/update cozyMetadata. Metadata are updated directly in `ensureCozyMetadata` using `client.metadata` for data that have to be passed by the app
- new format for updatedByApps (see https://github.com/cozy/cozy-libs/pull/302 )

I still need to update the docs.

### DIFF
--- a/docs/how-tos.md
+++ b/docs/how-tos.md
@@ -210,13 +210,36 @@ Custom relationships are useful if the relationship data is not stored in a buil
 
 ### Metadata
 
-cozy-client will also automatically insert and update the standard document metadata if you provide a `cozyMetadata` object to the schema.
+cozy-client will also automatically insert and update the standard document metadata. You may provide some info when you initialize the client:
 
-Each metadata field can be configured with the following options:
+- in the `appMetadata` parameter:
+  - `slug`: the app's slug
+  - `sourceAccount`: in case of the app is a connector, the `io.cozy.accounts` (defaults to `null`)
+  - `version`: the app's version
+- in the `schema`, for each doctype:
+  - `doctypeVersion`: the version of the doctype
 
-- `trigger`: Can be `creation` or `update`, and determines whether to update this value every time the document changes, or only on creation.
-- `value`: The value to use when setting the metadata field. If `value` is an array, the values will be appended to the array on each update.
-- `useCurrentDate`: Instead of a fixed value, use the current execution date for this field.
+For example, to initialize the client for google connector that deals with `io.cozy.contacts` and `io.cozy.contacts.accounts`:
+
+```javascript
+const client = new CozyClient({
+  schema: {
+    contacts: {
+      doctype: 'io.cozy.contacts',
+      doctypeVersion: 2
+    },
+    contactsAccounts: {
+      doctype: 'io.cozy.contacts.accounts',
+      doctypeVersion: 1
+    }
+  },
+  appMetadata: {
+    slug: 'konnector-google',
+    sourceAccount: 'xxx', // id of the io.cozy.accounts object
+    version: 3
+  }
+})
+```
 
 ### Validation
 
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -33,7 +33,6 @@ import uniqBy from 'lodash/uniqBy'
 import zip from 'lodash/zip'
 import forEach from 'lodash/forEach'
 import get from 'lodash/get'
-import set from 'lodash/set'
 
 const ensureArray = arr => (Array.isArray(arr) ? arr : [arr])
 
@@ -61,12 +60,14 @@ class CozyClient {
    * @param  {Link}         options.link   - Backward compatibility
    * @param  {Array.Link}   options.links  - List of links
    * @param  {Object}       options.schema - Schema description for each doctypes
+   * @param  {Object}       options.appMetadata - Metadata about the application that will be used in ensureCozyMetadata
    */
-  constructor({ link, links, schema = {}, ...options }) {
+  constructor({ link, links, schema = {}, appMetadata = {}, ...options }) {
     if (link) {
       console.warn('`link` is deprecated, use `links`')
     }
 
+    this.appMetadata = appMetadata
     this.options = options
     this.idCounter = 1
     this.isLogged = false
@@ -198,42 +199,54 @@ class CozyClient {
       event: TRIGGER_CREATION
     }
   ) {
-    if (!document._type) return document
-
-    const schema = this.schema.getDoctypeSchema(document._type)
-    if (!schema.cozyMetadata) return document
-
-    const enriched = {
-      ...document
+    if (this.appMetadata === undefined) return document
+    let doctypeVersion
+    if (document._type) {
+      const schema = this.schema.getDoctypeSchema(document._type)
+      doctypeVersion = get(schema, 'doctypeVersion')
     }
-    const metadata = Object.entries(schema.cozyMetadata)
 
-    for (const [
-      metadataName,
-      { trigger, value, useCurrentDate = false }
-    ] of metadata) {
-      const shouldSetMetadata =
-        options.event === TRIGGER_CREATION ||
-        (options.event === TRIGGER_UPDATE && trigger === TRIGGER_UPDATE)
-      if (!shouldSetMetadata) continue
+    const { slug, sourceAccount, version } = this.appMetadata
 
-      let newValue
-      const metadataPath = `cozyMetadata.${metadataName}`
+    const now = new Date().toISOString()
 
-      const shouldAccumulateValues = Array.isArray(value)
-      if (shouldAccumulateValues) {
-        const prevValue = get(document, metadataPath, [])
-        newValue = uniqBy([...prevValue, ...value])
-      } else if (useCurrentDate) {
-        newValue = new Date().toISOString()
-      } else {
-        newValue = get(document, metadataPath, value)
+    let cozyMetadata = get(document, 'cozyMetadata', {})
+    if (options.event === TRIGGER_CREATION) {
+      cozyMetadata = {
+        doctypeVersion,
+        createdByApp: slug,
+        sourceAccount,
+        createdAt: now,
+        createdByAppVersion: version,
+        updatedAt: now,
+        updatedByApps: slug
+          ? [
+              {
+                date: now,
+                slug,
+                version
+              }
+            ]
+          : [],
+        ...cozyMetadata // custom metadata that are set by the app
       }
-
-      set(enriched, metadataPath, newValue)
+    } else if (options.event === TRIGGER_UPDATE) {
+      cozyMetadata = {
+        ...cozyMetadata,
+        updatedAt: now,
+        updatedByApps: [
+          { date: now, slug, version },
+          ...get(document, 'cozyMetadata.updatedByApps', []).filter(
+            info => info.slug !== slug
+          )
+        ]
+      }
     }
 
-    return enriched
+    return {
+      ...document,
+      cozyMetadata
+    }
   }
 
   /**
--- a/docs/api/cozy-client.md
+++ b/docs/api/cozy-client.md
@@ -439,6 +439,7 @@ Responsible for
 | options.link | <code>Link</code> | Backward compatibility |
 | options.links | <code>Array.Link</code> | List of links |
 | options.schema | <code>Object</code> | Schema description for each doctypes |
+| options.appMetadata | <code>Object</code> | Metadata about the application that will be used in ensureCozyMetadata |
 
 <a name="CozyClient+collection"></a>
 
--- a/packages/cozy-client/src/CozyClient.spec.js
+++ b/packages/cozy-client/src/CozyClient.spec.js
@@ -184,7 +184,15 @@ describe('CozyClient', () => {
 
   let client
   beforeEach(() => {
-    client = new CozyClient({ links: [link], schema: SCHEMA })
+    client = new CozyClient({
+      links: [link],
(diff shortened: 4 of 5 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -62,11 +61,12 @@ class CozyClient {
    * @param  {Array.Link}   options.links  - List of links
    * @param  {Object}       options.schema - Schema description for each doctypes
    */
-  constructor({ link, links, schema = {}, ...options }) {
+  constructor({ link, links, schema = {}, metadata = {}, ...options }) {
```
edas: Why not having our metadata directly in the `options`?

packages/cozy-client/src/CozyClient.js
```diff
@@ -198,42 +198,54 @@ class CozyClient {
...
-
-      const shouldAccumulateValues = Array.isArray(value)
-      if (shouldAccumulateValues) {
-        const prevValue = get(document, metadataPath, [])
-        newValue = uniqBy([...prevValue, ...value])
-      } else if (useCurrentDate) {
-        newValue = new Date().toISOString()
-      } else {
-        newValue = get(document, metadataPath, value)
+    if (!document._type || this.metadata === undefined) return document
```
edas: Not sure about why we filter based on `_type` now that we do not need the schema anymore.

packages/cozy-client/src/CozyClient.js
```diff
@@ -198,42 +198,54 @@ class CozyClient {
...
+
+    const {
+      doctypeVersion,
```
edas: Hum… `this.metadata` is from the client initialization. If you client should manage multiple doctypes, you can't define a unique `doctypeVersion` there. This is the one that still should be from the schema

packages/cozy-client/src/CozyClient.js
```diff
@@ -198,42 +198,54 @@ class CozyClient {
...
+            date: now,
+            slug
+          }
+        ]
+      }
+    } else if (options.event === TRIGGER_UPDATE) {
+      cozyMetadata = {
+        // add doctypeVersion and sourceAccount if they don't exist
+        doctypeVersion,
+        sourceAccount,
```
edas: Not sure we should define a `sourceAccount`. If my connector is updating pre-existing data, we should probably not say that these updated documents come from the connector account.

packages/cozy-client/src/CozyClient.js
```diff
@@ -198,42 +198,54 @@ class CozyClient {
...
+          {
+            date: now,
+            slug
+          }
+        ]
+      }
+    } else if (options.event === TRIGGER_UPDATE) {
+      cozyMetadata = {
+        // add doctypeVersion and sourceAccount if they don't exist
+        doctypeVersion,
```
edas: This one seems dangerous. We assume that updating an existing document will always respect the new schema. Will it?

I need comments from whoever has one.

packages/cozy-client/src/CozyClient.js
cedricmessiant: That's what I did first and then decided to make a clear distinction (as it holds 4 fields and it is used only in `ensureCozyMetadata`). It also allows us to initialize it with an empty object.

packages/cozy-client/src/CozyClient.js
cedricmessiant: you're right it's not needed anymore

packages/cozy-client/src/CozyClient.js
cedricmessiant: :+1:

packages/cozy-client/src/CozyClient.js
y-lohse: Yeah I wouldn't automatically update that. We currently try very hard to avoid having multiple versions of the same doctype in the wild, but still.

packages/cozy-client/src/CozyClient.js
```diff
@@ -198,42 +198,54 @@ class CozyClient {
...
+        sourceAccount,
+        ...cozyMetadata,
+        updatedAt: now,
+        updatedByApps: [
+          ...get(document, 'cozyMetadata.updatedByApps', []).filter(
+            info => info.slug !== slug
+          ),
+          { date: now, slug }
```
y-lohse: > By convention, the first app (slug) listed in updatedByApps is the one which did the last update.

… so we should move the `slug` in front of the spread.

packages/cozy-client/src/CozyClient.js
edas: Don't try to hard. We *will* have documents in different versions in the future. This will happen once we have third party apps, or when someone is recovering an old export, or when someone play directly with the API.

It's OK to avoid it as much as we can but we definitly should design our apps and libs to be resistant to having documents of different versions in the same doctype.

packages/cozy-client/src/CozyClient.js
y-lohse: Nit: Should this be called `cozyMetadata`?

edas: Probably not, we are speaking of the cozyClient initialization metadata

packages/cozy-client/src/CozyClient.js
edas: Once we add the date in each object, this convention is to be removed. Let's simplify things

packages/cozy-client/src/CozyClient.js
y-lohse: My thinking was that it the data goes into the `cozyMetadata` field. This more or less makes it sound like we're setting the metadata for cozy-client itself, which is not really a thing. But not a big deal.

edas: I am not convinced. Let's talk about it tomorow

packages/cozy-client/src/CozyClient.js
cedricmessiant: Ok I removed it. Do we want to update it if e.g the contact is attached to a contact account later or is it only if the contact has been created via the konnector? From what I understand here, it's only for creation (and we still have the id of the account in the `io.cozy.contacts.accounts` so it should be OK)

packages/cozy-client/src/CozyClient.js
cedricmessiant: I removed this sentence here: https://github.com/cozy/cozy-doctypes/pull/79/files
But it may be useful (but maybe dangerous) to always have the last update first since we use an array

packages/cozy-client/src/CozyClient.js
cedricmessiant: I hesitated but I found it confusing as it will not be injected directly in `cozyMetadata`. It is more data that will allow us to create the `cozyMetadata`. @edas @ptbrowne opinions on this?

packages/cozy-client/src/CozyClient.js
cedricmessiant: I'm not sure I understand what is the problem.
(However, I was not convinced it was a good idea to create the `doctypeVersion` if it does not exist)

edas: You have a contact document which don't have a `version` attribute (so it's probably an old one). You touch it in a small way, you probably don't want to assume it now follows the version=3 correctly

cedricmessiant: Ok I see. Indeed, it's the caller's responsibility to change the metadata in this case.

packages/cozy-client/src/CozyClient.js
cedricmessiant: (sorry about my last comment, I did not refresh before)

cedricmessiant: @edas @ptbrowne @y-lohse what do you think about `appMetadata`?

cedricmessiant:
Edited the docs, ready for final review

cedricmessiant:
Edited the docs, ready for final review

packages/cozy-client/src/CozyClient.js
edas: 👌 with me

docs/how-tos.md
```diff
@@ -210,13 +210,36 @@ Custom relationships are useful if the relationship data is not stored in a buil
 
 ### Metadata
 
-cozy-client will also automatically insert and update the standard document metadata if you provide a `cozyMetadata` object to the schema.
+cozy-client will also automatically insert and update the standard document metadata. You have to provide some info when you initialize the client:
```
edas: `metadata` being an addition, it's better if we can leave it as optional. "have to" -> "may"

packages/cozy-client/src/CozyClient.js
edas: 👌 with me. It's indeed clearer as appMetadata (we will know that 'version' is the version of the app)

packages/cozy-client/src/CozyClient.js
```diff
@@ -198,42 +198,51 @@ class CozyClient {
...
+
+    const now = new Date().toISOString()
+
+    let cozyMetadata = get(document, 'cozyMetadata', {})
+    if (options.event === TRIGGER_CREATION) {
+      cozyMetadata = {
+        ...cozyMetadata, // custom metadata that are set by the app
+        doctypeVersion,
+        createdByApp: slug,
+        sourceAccount,
```
edas: Not blocking but I am unsure is this one should be after or before `...cozyMetadata` (ie: one app may want to override the default)

packages/cozy-client/src/CozyClient.js
```diff
@@ -198,42 +198,51 @@ class CozyClient {
...
+    const { slug, sourceAccount = null, version } = this.metadata
+
+    const now = new Date().toISOString()
+
+    let cozyMetadata = get(document, 'cozyMetadata', {})
+    if (options.event === TRIGGER_CREATION) {
+      cozyMetadata = {
+        ...cozyMetadata, // custom metadata that are set by the app
+        doctypeVersion,
+        createdByApp: slug,
```
edas: what if one want to override this or define it manually ? should it be only the default before `...cozyMetadata` ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -198,42 +198,51 @@ class CozyClient {
...
-      } else {
-        newValue = get(document, metadataPath, value)
+    if (this.metadata === undefined) return document
+    let doctypeVersion
+    if (document._type) {
+      const schema = this.schema.getDoctypeSchema(document._type)
+      doctypeVersion = get(schema, 'doctypeVersion')
+    }
+
+    const { slug, sourceAccount = null, version } = this.metadata
```
edas: Is there a way to define these values automagicaly in the init if not provided by the calling app ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -198,42 +198,51 @@ class CozyClient {
...
+            slug
+          }
+        ]
+      }
+    } else if (options.event === TRIGGER_UPDATE) {
+      cozyMetadata = {
+        ...cozyMetadata,
+        updatedAt: now,
+        updatedByApps: [
+          { date: now, slug },
```
edas: if "slug" is not mandatory, please avoid using it when not defined ?

packages/cozy-client/src/CozyClient.js
y-lohse: Yep, sounds good!

packages/cozy-client/src/CozyClient.js
cedricmessiant: Upon reflection, I think that we should in fact allow the app to override any metadata. If you provide `cozyMetadata` with some values on save, that's the data you want in the database.

packages/cozy-client/src/CozyClient.js
cedricmessiant: I don't think so. The client has no idea of what app is initializing it, right?

edas: In apps it's available in the DOM (set up by the stack) but this was a true question. I don't know if it's a good thing to let the client fetch that automatically

packages/cozy-client/src/CozyClient.js
edas: 👍

packages/cozy-client/src/CozyClient.js
cedricmessiant: I think that if you provide `appMetadata`, `slug` should be mandatory (as multiple metadata rely on it). If you agree I can change the check at the beginning of `ensureCozyMetadata` by `if (get(this, 'appMetadata.slug') === undefined) return document`

packages/cozy-client/src/CozyClient.js
y-lohse: In the case of mobile apps the info is not in the DOM. Mayyybe it could use the current token to look up this information but I think that's a stretch and way outside of the scope of this PR.

packages/cozy-client/src/CozyClient.js
y-lohse: Ok for me — but only when the document is created, not for updates.

packages/cozy-client/src/CozyClient.js
edas: shouldn't we still set up `updatedAt` ans 'createdAt` anyways ? 

But otherwise, ok for me

packages/cozy-client/src/CozyClient.js
edas: 👌

cedricmessiant:
It should be good now!

edas:
LGTM but leaving @y-lohse do the formal 👍

PR #382, opened 2019-03-08
Add metadata version

cedricmessiant:
Add `metadataVersion` to `cozyMetadata` to match the structure defined in https://github.com/cozy/cozy-doctypes/pull/79

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -199,6 +199,7 @@ class CozyClient {
       event: TRIGGER_CREATION
     }
   ) {
+    const METADATA_VERSION = 1
     if (this.appMetadata === undefined) return document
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -199,6 +199,7 @@ class CozyClient {
       event: TRIGGER_CREATION
     }
   ) {
+    const METADATA_VERSION = 1
```
edas: [very] minor and not blocking: I would have put this in it's own file with an export.

cedricmessiant: At first, I searched for a `constants.js` file without success. But in the end, if we change something in `cozyMetadata`, we are likely to change `METADATA_VERSION` so I think it can stay here if you don't mind.

PR #383, opened 2019-03-08
feat: Make mime-types package optional

ptbrowne:
mime-db (on which mime-types rely) is very big (see https://github.com/cozy/cozy-banks/issues/926) and not all apps rely on automatic
file type from extension.

Related : https://github.com/cozy/cozy-client/pull/373

In this PR, mime-types is made optional, if an app wants to
use automatic file type detection from extension, it can
add manually the mime-types package (Drive :) ).

### DIFF
--- a/packages/cozy-stack-client/package.json
+++ b/packages/cozy-stack-client/package.json
@@ -12,8 +12,10 @@
   },
   "dependencies": {
     "detect-node": "2.0.4",
-    "mime-types": "2.1.20",
     "qs": "6.6.0"
+  },
+  "optionalDependencies": {
+    "mime-types": "2.1.20"
   },
   "scripts": {
     "build": "../../bin/build",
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/package.json
```diff
@@ -12,9 +12,11 @@
   },
   "dependencies": {
     "detect-node": "2.0.4",
-    "mime-types": "2.1.20",
     "qs": "6.6.0"
   },
+  "optionalDependencies": {
+    "mime-types": "2.1.20"
+  },
```
nono: As I understand, if we put mime-types in optionalDependencies, yarn&npm will try by default to install it. I don't know if we shouldn't put in `peerDependencies`.

edas: peerDependency allows us to define the version of the dependency outside of the lib. It's aimed a not duplicating modules across different libraries. peerDependencies are required dependencies, they will always be present in the final lib

ptbrowne:
Closing in favor of https://github.com/cozy/cozy-client/pull/373

An app can use a webpack alias to an empty file if it does not need mime functionality.

PR #392, opened 2019-03-14
StackClient: Add KonnectorCollection extending AppCollection

gregorylegarec:
Based on master before previous revert.

So the first commit reverts the revert, but everything ends well.


gregorylegarec:
Bad code at the bad moment. Closing.

Issue #374, opened 2019-02-27
Discussion on Collection


ptbrowne:
From: https://github.com/cozy/cozy-client/issues/372#issuecomment-467786027

I find the Collection APIs in StackClient a bit awkward to use.

When I read `Collection` I expect it to contain documents and to stay around a long time. Here `Collection` happen to be short lived objects (since you cannot store them since you cannot expect the client not to be changed), more fetchers than collections.

Using doctypes (`io.cozy.konnectors`) to access fetching methods is a bit awkard I think and it also feels like an implementation details.

I would prefer to use APIs like that : 

```
stackClient.apps.fetchMaintenanceList()
stackClient.apps.fetchMaintenanceList()
```

It was the case with the old cozy-client-js, I think it was better than the new "Collection" one.

gregorylegarec:
I think I pretty agree with it. 

But we still have the case of regular documents. I feel taht indicating the desired doctype early is a pretty confortable. Maybe it is just a matter a words, and we could keep the `collection` mechanism but rename it to something like `doctype`.

Our API could become
```
stackClient.apps.fetch()
stackClient.triggers.launch()
stackClient.jobs.subscribe() // Yeah realtime seems now natural an easy
stackClient.files.get()
stackClient.doctype('io.cozy.accounts').update()
// etc.
```

edas:
Ok, as long as we only use that pattern for core specific features.

Please take care to not put document/collections methods there (ie: like no `stackClient.contacts.find('…')`

gregorylegarec:
Why does it seems a bad idea to you ?

`stackClient.apps.all()` [needs a dedicated endpoint](https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/AppCollection.js#L27), for example.

edas:
`.apps.all()` is ok for me because we are not addressing the user's document and own couch databases. Looking for apps in the instance is probably a "specific feature" for me (listing contacts in io.cozy.contacts is not).

Still thinking about it

PR #393, opened 2019-03-15
StackClient: Add Konnector Collection basis

gregorylegarec:
Prepare the future implementation for konnector endpoints.

For now it allow developers to user `stackClient.konnectors.all()` to fetch all konnectors.

It keeps `stackClient.collection('io.cozy.konnetors').all()` plugged to regular `DocumentCollection` to not introduce regression in Cozy-Banks. See https://github.com/cozy/cozy-banks/commit/a7c8ac3823d18300bd3d2ef09c57795fc48d47f3#diff-872b51c7bdb364ada730166a867a9d38R90

Related to https://github.com/cozy/cozy-client/issues/374

### DIFF
--- a/packages/cozy-stack-client/src/__tests__/KonnectorCollection.spec.js
+++ b/packages/cozy-stack-client/src/__tests__/KonnectorCollection.spec.js
@@ -1,0 +1,85 @@
+jest.mock('../CozyStackClient')
+
+import CozyStackClient from '../CozyStackClient'
+import KonnectorCollection from '../KonnectorCollection'
+import ALL_KONNECTORS_RESPONSE from './fixtures/konnectors.json'
+
+const FIXTURES = {
+  ALL_KONNECTORS_RESPONSE
+}
+describe(`KonnectorCollection`, () => {
+  const client = new CozyStackClient()
+
+  describe('all', () => {
+    const collection = new KonnectorCollection(client)
+
+    beforeAll(() => {
+      client.fetchJSON.mockReturnValue(
+        Promise.resolve(FIXTURES.ALL_KONNECTORS_RESPONSE)
+      )
+    })
+
+    it('should call the right route', async () => {
+      await collection.all()
+      expect(client.fetchJSON).toHaveBeenCalledWith('GET', '/konnectors/')
+    })
+
+    it('should return a correct JSON API response', async () => {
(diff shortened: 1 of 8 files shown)


packages/cozy-stack-client/src/__tests__/KonnectorCollection.spec.js
```diff
@@ -0,0 +1,85 @@
...
+      )
+    })
+  })
+
+  describe('update', () => {
+    const collection = new KonnectorCollection(client)
+
+    it('should throw error', async () => {
+      expect(collection.update()).rejects.toThrowError(
+        'update() method is not available for applications'
```
ptbrowne: s/applications/konnectors ?

packages/cozy-stack-client/src/__tests__/KonnectorCollection.spec.js
```diff
@@ -0,0 +1,85 @@
...
+
+    it('should return normalized documents', async () => {
+      const resp = await collection.all()
+      expect(resp.data[0]).toHaveDocumentIdentity()
+    })
+  })
+
+  describe('find', () => {
+    it('throw error', async () => {
+      const collection = new KonnectorCollection(client)
```
edas: In don't see the point of these tests. 

You should test : 
- Public interfaces (what a dev should be able to use)
- Expectable dev errors (the dev should not use the interface that way but it may, so we test the returning error)
- Regression or changes proof (the code had another behaviour, we test that the behaviour has changed)

I don't see why we are confirming that calling non existent methods of a new module are returning an exception

gregorylegarec: I've just duplicated tests for AppCollection.

I don't see any input from dev except calling the method `stackClient.konnectors.all()`, but I could be missing a point.

Actually no behavior is changed, we are just adding a namespace for a `KonnectorCollection`. I wanted to highlight this in the last commit by explicitly returning a `DocumentCollection()` for doctype `io.cozy.konnectors` instead of relying on the default behaviour of the switch/case in `CozyStackClient`. By the way I may explicitly test this to ensure no regression is made. I may allow remove the last commit without changing the behaviour but I feel like make a comment here is more future-proof.

PR #394, opened 2019-03-18
[pouch-link] feat: Log via minilog

ptbrowne:
I propose that our libs use minilog for logging. It has namespaces and filtering abilities. This would allow us to view logs quickly without polluting the logs if we do care for them.

2.5kb minified/gzipped.

A package level logger is created. An application will be able to view logs via

```
require('minilog').enable()
```

Warn/errors logs are visible directly.

If an application wants to see the debug logs of cozy-pouch-link :

```
require('minilog').suggest.allow('cozy-pouch-link', 'debug')
```


ptbrowne:
Done in https://github.com/cozy/cozy-client/pull/394/commits/445ec5022f43f5654cee526d53170b43f3d147f1

PR #395, opened 2019-03-18
WIP: Add selectFromState

gregorylegarec:
We are looking to be able to know the status of a given konnector. For that, we need to:
* Retrieve all triggers related with Worker=konnector (already available from TriggerCollection)
* Select the triggers related to one given konnector from the cozyClient redux store

For now, only `getDocumentFromState` is available, but we need something a little bit more powerful than just selecting a document from its doctype and its id.

I was thinking of a way to run a selector on cozyClient's state. This PR suggest something like:
```js
client.selectFromState(state => Object.keys(documents['io.cozy.triggers']).reduce(
// filter
))
```

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -21,7 +21,8 @@ import {
   receiveMutationResult,
   receiveMutationError,
   getQueryFromState,
-  getDocumentFromState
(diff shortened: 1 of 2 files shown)


gregorylegarec:
Replaced by https://github.com/cozy/cozy-client/pull/396.

PR #396, opened 2019-03-19
Client: Add getCollectionFromState()

gregorylegarec:
Goal: easily get all the fetched triggers.

`AddCollectionFromState()` allows to retrieve all document for a given doctype existing in CozyClient redux state.

Example:
```
await client.query(client.all('io.cozy.accounts').where(auth: '$exists'))

const accounts = client.getCollectionFromState('io.cozy.accounts')
```

It's the "get list" alternative of existing method `getDocumentFromState()`.

### DIFF
--- a/packages/cozy-client/src/__tests__/store.spec.js
+++ b/packages/cozy-client/src/__tests__/store.spec.js
@@ -3,6 +3,7 @@ import sortBy from 'lodash/sortBy'
 
 import reducer, {
   initQuery,
+  getCollectionFromState,
   getDocumentFromState,
   getQueryFromStore,
   receiveQueryResult,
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/__tests__/store.spec.js
```diff
@@ -178,6 +179,57 @@ describe('Store', () => {
...
+  describe('getCollectionFromState', () => {
+    it('should return null when the store is empty', () => {
+      const spy = jest.spyOn(global.console, 'warn').mockReturnValue(jest.fn())
+      expect(getCollectionFromState(store.getState(), 'io.cozy.todos')).toBe(
+        null
+      )
+      spy.mockRestore()
+    })
+
+    it('should return document list for given doctype', () => {
```
ptbrowne: should return all documents for the given doctype
```suggestion
    it('should return document list for given doctype', () => {
```

packages/cozy-client/src/__tests__/store.spec.js
```diff
@@ -178,6 +179,57 @@ describe('Store', () => {
...
+        [TODO_1]
+      )
+    })
+
+    it('should return an updated rev of a doc received in a mutation result', () => {
+      store.dispatch(
+        receiveMutationResult('foo', {
+          data: [TODO_1]
+        })
+      )
```
ptbrowne: I am not sure I would test this logic. It is the role of receiveMutationResult to correctly update the store. Here we are testing receiveMutationResult instead of getCollectionFromState.

The 2 first tests seem sufficient.

gregorylegarec: I agree, I adapted the tests for `getDocumentFromState` but they are not all pertinent for `getCollectionFromSate`.

PR #397, opened 2019-03-19
Client: Update fetchRelationships method

gregorylegarec:
We want to retrieve all the triggers for a given konnector to know which error they are containing.

For that, we will use a custom association in the konnector schema.

This association will have an async query method, which may return existing document.

The goal is to use an Association like this (naive implementation):
```js
class HasManyTriggers {
  static async query(){
    const triggers = await fethTriggers()
    return triggers.filter(isTriggerForKonnector(slug)
  }
}

```

This PR :
* Allows async query methods in Associations
* Allows a query method to return documents
* Fix a few things

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -466,6 +466,19 @@ class CozyClient {
     // storing the document. This makes the data easier to manipulate for the front-end.
     const relationshipsByDocId = {}
 
+    for (const def of documents) {
+      const docIdAndRels = queryDefToDocIdAndRel.get(def)
+      for (const docIdAndRel of docIdAndRels) {
+        if (docIdAndRel) {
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -467,6 +467,19 @@ class CozyClient {
...
+        if (docIdAndRel) {
+          const [docId, relName] = docIdAndRel
+          relationshipsByDocId[docId] = relationshipsByDocId[docId] || {}
+          relationshipsByDocId[docId][relName] = responseToRelationship({
+            data: def
+          })
+        }
+      }
+    }
+
```
ptbrowne: Mmh I think this for and the following could be extracted into a method `reconstructRelationships(queryDefToDocIdAndRel, definitions, documents, responses)`

PR #400, opened 2019-03-19
StackClient/Client: Add HasManyTriggers association and update TriggerCollection

gregorylegarec:
This PR prepares need of retrieving all the triggers for a given konnector, with the goal to detect all errors related to it.

It fixes a few mistakes in TriggerCollection and implements the find() method, which will allow us to do:

```js
const triggers = client.query(client.all('io.cozy.triggers').where({worker: 'konnector'})
```

It also introduces the `HasManyTriggers` association, which fetch all triggers or get them in CozyClient redux store. It then filters them to only get triggers related to the konnector having the association.

It is more or less a hack to manage the way that we only have the endpoint `/jobs/triggers` to retrieve triggers.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -418,12 +418,17 @@ class CozyClient {
     const responseDocs = isSingleDoc ? [response.data] : response.data
 
     const queryDefToDocIdAndRel = new Map()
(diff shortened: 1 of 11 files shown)


gregorylegarec:
Wrong branch, sorry. See #401 instead.

Issue #399, opened 2019-03-19
Client: Mutualize QueryDefinition returned by Association.query


gregorylegarec:
In `client.fetchRelationships`, we execute the QueryDefinitions returned by Associations.query method, even if they are all equal.

It could results in the same request to the stack done a multiple time (for every document in `response`.

We should detect this and run any unique query only once, and then share the results into all related documents.

PR #401, opened 2019-03-19
StackClient/Client: Add HasManyTriggers and update TriggerCollection

gregorylegarec:
This PR prepares need of retrieving all the triggers for a given konnector, with the goal to detect all errors related to it.

It fixes a few mistakes in TriggerCollection and implements the find() method, which will allow us to do:

```js
const triggers = client.query(client.all('io.cozy.triggers').where({worker: 'konnector'})
```

It also introduces the `HasManyTriggers` association, which fetch all triggers or get them in CozyClient redux store. It then filters them to only get triggers related to the konnector having the association.

It is more or less a hack to manage the way that we only have the endpoint `/jobs/triggers` to retrieve triggers.

### DIFF
--- a/docs/api/cozy-client.md
+++ b/docs/api/cozy-client.md
@@ -20,6 +20,9 @@ following the JSON API spec.</p>
 <dd><p>Used when related documents are stored directly under the attribute with
 only the ids.</p>
 </dd>
+<dt><a href="#HasManyTriggers">HasManyTriggers</a> ⇐ <code><a href="#HasMany">HasMany</a></code></dt>
+<dd><p>Association used for konnectors to retrieve all their related triggers.</p>
+</dd>
 <dt><a href="#CozyClient">CozyClient</a></dt>
 <dd><p>Responsible for</p>
 <ul>
@@ -52,6 +55,15 @@ from a Cozy. <code>QueryDefinition</code>s are sent to links.</p>
   }
 }, cozyStackClient)
 </code></pre>
+</dd>
+</dl>
+
+## Constants
+
+<dl>
+<dt><a href="#fetchKonnectorTriggers">fetchKonnectorTriggers</a></dt>
+<dd><p>Memoize the fetch promise. For now we are only fetching it once but it will
+evolve in the future.</p>
 </dd>
 </dl>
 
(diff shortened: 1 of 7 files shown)


packages/cozy-client/src/associations/HasManyTriggers.js
```diff
@@ -0,0 +1,43 @@
...
+  static async query(doc, client) {
+    // Check the memoized promise instead of the redux store to avoid executing
+    // the same query several times (one time per konnector)
+    if (!memoizedFetchPromise) {
+      memoizedFetchPromise = client.query(
+        client.all(TRIGGERS_DOCTYPE).where({ worker: 'konnector' })
+      )
+    }
+
+    await memoizedFetchPromise
```
ptbrowne: The client can change of Cozy in its lifetime, I would use _.memoize and pass a key function based on the client URI to be asfe.

const getKonnectorTriggers = memoize(client => {
  return client.all(TRIGGERS_DOCTYPE).where({ worker: 'konnector' })
}, client => client.stackClient.uri)

gregorylegarec: Thanks for pointing me this point out, I am too little familiar with mobile constraintes 👍

We will be able to get rid of this memoization when https://github.com/cozy/cozy-client/issues/399 will be implemented.

docs/api/cozy-client.md
```diff
@@ -403,6 +415,41 @@ const todo = {
...
+
+* [HasManyTriggers](#HasManyTriggers) ⇐ [<code>HasMany</code>](#HasMany)
+    * _instance_
+        * [.addById()](#HasMany+addById)
+    * _static_
+        * [.query()](#HasManyTriggers.query)
+
+<a name="HasMany+addById"></a>
+
+### hasManyTriggers.addById()
```
CPatchane: s/hasManyTriggers/HasManyTriggers

packages/cozy-client/src/associations/HasManyTriggers.js
gregorylegarec: I fixed up the code.

PR #402, opened 2019-03-19
StackClient: Update KonnectorCollection

gregorylegarec:
Last step for us to be able to get installed konnectors and
their triggers as HasManyTriggers relationship.

Unfortunately we would like to do:

```
const konnectors = client.query(client.all('io.cozy.konnectors').include('triggers)
```

But [remember](https://github.com/cozy/cozy-client/pull/393), we have a blocking change in Cozy-Banks that forbid us to respond to client.all('io.cozy.konnectors') with a KonnectorCollection.

Here the good news, as `client.all('io.cozy.konnectors').where({ something })` rely internally on `KonnectorCollection.find()`, this PR made `AppCollection.find()` encapsulate a call to `DocumentCollection.find()`, which make everything works well (As `KonnectorCollection` extends `AppCollection`.

### DIFF
--- a/packages/cozy-stack-client/src/AppCollection.js
+++ b/packages/cozy-stack-client/src/AppCollection.js
@@ -3,8 +3,8 @@ import { uri } from './utils'
 
 export const APPS_DOCTYPE = 'io.cozy.apps'
 
-export const normalizeApp = app => {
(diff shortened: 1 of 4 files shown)


gregorylegarec:
Related to https://github.com/cozy/cozy-client/pull/401

gregorylegarec:
@kosssi As discussed, I added a `BREAKING CHANGE` and warning message.

PR #403, opened 2019-03-19
Client: Add Schema.add() method

gregorylegarec:
In the cozy-bar we will need to access konnectors, and we cannot be sure that CozyClient has been initialized with the expected schema.

This PR adds a method to dinamically add a schema.

```js
client.addSchema({
  todos: {
    doctype: 'io.cozy.todos'
  }
})
```

### DIFF
--- a/packages/cozy-client/src/Schema.js
+++ b/packages/cozy-client/src/Schema.js
@@ -1,5 +1,6 @@
 import keyBy from 'lodash/keyBy'
 import mapValues from 'lodash/mapValues'
+import merge from 'lodash/merge'
 import size from 'lodash/size'
 import { resolveClass as resolveAssociationClass } from './associations'
 
@@ -52,13 +53,18 @@ const normalizeDoctypeSchema = doctypeSchema => {
  */
 class Schema {
   constructor(schemaDefinition = {}, client = null) {
+    this.add(schemaDefinition)
+    this.client = client
+  }
+
+  add(schemaDefinition = {}) {
     const values = mapValues(schemaDefinition, (obj, name) => ({
       name,
       ...normalizeDoctypeSchema(obj)
     }))
-    this.client = client
-    this.byName = keyBy(values, x => x.name)
-    this.byDoctype = keyBy(values, x => x.doctype)
+
+    this.byName = merge(this.byName || {}, keyBy(values, x => x.name))
+    this.byDoctype = merge(this.byDoctype || {}, keyBy(values, x => x.doctype))
   }
 
   /**
--- a/packages/cozy-client/src/Schema.js
+++ b/packages/cozy-client/src/Schema.js
@@ -2,6 +2,7 @@ import keyBy from 'lodash/keyBy'
 import mapValues from 'lodash/mapValues'
 import merge from 'lodash/merge'
 import size from 'lodash/size'
+import intersectionBy from 'lodash/intersectionBy'
 import { resolveClass as resolveAssociationClass } from './associations'
 
 /**
@@ -30,6 +31,28 @@ const normalizeDoctypeSchema = doctypeSchema => {
   }
 }
 
+const assert = (predicate, errorMessage) => {
+  if (!predicate) throw new Error(errorMessage)
+}
+
+const ensureCanBeAdded = (newSchemas, existingSchemas) => {
+  const sameNames = intersectionBy(newSchemas, existingSchemas, x => x.name)
+  assert(
+    sameNames.length === 0,
+    `Duplicated names in schemas being added: ${sameNames.join(', ')}`
+  )
+
+  const sameDoctypes = intersectionBy(
+    newSchemas,
+    existingSchemas,
+    x => x.doctype
+  )
+  assert(
+    sameDoctypes.length === 0,
+    `Duplicated doctypes in schemas being added: ${sameDoctypes.join(', ')}`
+  )
+}
+
 /**
  * Stores information on a particular doctype.
  *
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/Schema.js
```diff
@@ -52,13 +53,18 @@ const normalizeDoctypeSchema = doctypeSchema => {
...
     const values = mapValues(schemaDefinition, (obj, name) => ({
       name,
       ...normalizeDoctypeSchema(obj)
     }))
-    this.client = client
-    this.byName = keyBy(values, x => x.name)
-    this.byDoctype = keyBy(values, x => x.doctype)
+
+    this.byName = merge(this.byName || {}, keyBy(values, x => x.name))
+    this.byDoctype = merge(this.byDoctype || {}, keyBy(values, x => x.doctype))
```
ptbrowne: Can you add a check to see if some keys are already existing ? I think we should throw in theses cases.

packages/cozy-client/src/Schema.js
```diff
@@ -1,6 +1,9 @@
 import keyBy from 'lodash/keyBy'
+import keys from 'lodash/keys'
```
CPatchane: What's the difference with `Object.keys`? Why do we specially need to use a lodash method instead of the native method which does [the same thing](https://github.com/lodash/lodash/blob/master/keys.js#L34)? I don't think we need to handle array likes here which is the only difference IMHO

gregorylegarec: Readability

ptbrowne: Object.keys seems to be the way to go here :)

I don't know what you mean by "readability" here, I guess it's because "Object." is not there. In this case `const keys = Object.keys` does the job.

packages/cozy-client/src/Schema.js
```diff
@@ -52,13 +63,27 @@ const normalizeDoctypeSchema = doctypeSchema => {
...
+      this.byName = tryMerge(this.byName, keyBy(values, x => x.name))
+    } catch (error) {
+      throw new Error(`Cannot merge Schema by name (${error.message})`)
+    }
+
+    try {
+      this.byDoctype = tryMerge(this.byDoctype, keyBy(values, x => x.doctype))
+    } catch (error) {
+      throw new Error(`Cannot merge Schema by doctype (${error.message})`)
+    }
```
ptbrowne: I think those errors could be improved and it should be done before assigning this.byName or this.byDoctype since if you have an error when assigning this.byDoctype, the assignement to this.byName will have succeeded.

I think the check phase and assignment phase should be separated.

gregorylegarec: You're right, that was how my first version was working but then I refactored to this one without paying attention to this point.

Let's go back !

gregorylegarec: I updated the code with better check.

By the way, `this.byName` was never used so I just replaced it by an array of names.

ptbrowne: > By the way, this.byName was never used

I think we can go further and remove completely schemaNames. Less internal state is good, especially here where schemaNames can be easily computed with `const schemaNames = Object.values(this.byDoctype).map(x => x.name)`

gregorylegarec: You're right, that was my first move to totally remove it but then I get it back to check for the existing names and avoid duplicates.

Your solution is perfect, give me one minute to update the code.

packages/cozy-client/src/Schema.js
```diff
@@ -52,13 +59,43 @@ const normalizeDoctypeSchema = doctypeSchema => {
...
+  add(schemaDefinition = {}) {
+    const namesToAdd = Object.keys(schemaDefinition)
+
+    if (intersectWith(namesToAdd, this.schemaNames))
+      throw new Error(
+        `Cannot add schemas with name(s): ${intersection(
+          namesToAdd,
+          this.schemaNames
+        ).join(', ')} (name(s) already exist(s)`
+      )
```
ptbrowne: I find that we should strive to have the happy case as clear as possible and keep checks very short. I feel that the logic of the method is shadowed by the checks.

```js
const ensureCanBeAdded = (schema, otherSchema) => {
  const sameNames = intersectionBy(newSchemas, existingSchemas, x => x.name)
  assert(sameNames.length === 0, 'Duplicated names in schemas being added: ' + sameNames)

  const sameDoctypes = intersectionBy(newSchemas, existingSchemas, x => x.doctype)
  assert(sameDoctypes.length === 0, 'Duplicated doctypes in schemas being added: ' + sameDoctypes)
}

add(schemaDefinition={}) {
  // Normalize as early as possible to deal with clean objects as soon as possible
  const newSchemas = normalizeSchemaDefinition(schemaDefinition)
  const existingSchemas = Object.values(this.byDoctype)
  ensureCanBeAdded(newSchemas, existingSchemas)
   
  ....
}
```

gregorylegarec: I agree, I am not plenty satisfied by the current implementation. Let's use your suggestion. 👍

gregorylegarec: And good point about the normalization

ptbrowne: intersectionBy is in lodash but not assert, I add it via `const assert = (cond msg) => { if (!cond) { throw new Error(msg) } }

gregorylegarec: Tell me if the last version suits you (I rebased).

PR #406, opened 2019-03-21
StackClient: Make KonnectorCollection.find call instance of DocumentCollection

gregorylegarec:
Using `function.call` was not working as expected and was causing issues in Banks, we actually need a full instance of DocumentCollection here, to have access to other methods and properties like `this.toMangoOptions`, `this.indexes`, etc.

We actually need ~a full instance and not only the find()
method from DocumentCollection~ to extend DocumentCollection


ptbrowne:
What was the problem exactly ?

ptbrowne:
I think it would work if AppCollection extended DocumentCollection. I think it's a better solution.

gregorylegarec:
It would work for sure but I am afraid of all the implications and possible regressions. 

I suggest to merge this fix and add an issue to deal with as soon as I have 30 minutes to test this.

gregorylegarec:
@ptbrowne I updated with an extend. It seems to work.

packages/cozy-stack-client/src/AppCollection.js
```diff
@@ -37,31 +35,6 @@ class AppCollection {
...
-    throw new Error('create() method is not available for applications')
-  }
-
-  async update() {
-    throw new Error('update() method is not available for applications')
-  }
-
-  async destroy() {
-    throw new Error('destroy() method is not available for applications')
-  }
```
drazik: `AppCollection` will use `find` from `DocumentCollection`, but also `get`, `create`, `update` and `destroy`. Maybe you want to keep these methods implementations that throw an error here?

packages/cozy-stack-client/src/KonnectorCollection.js
```diff
@@ -3,8 +3,12 @@ import AppCollection from './AppCollection'
 export const KONNECTORS_DOCTYPE = 'io.cozy.konnectors'
 
 class KonnectorCollection extends AppCollection {
-  doctype = KONNECTORS_DOCTYPE
   endpoint = '/konnectors/'
```
drazik: Nit: you can also put this in the constructor, so there is not a mix between class properties and initialization in the constructor.

packages/cozy-stack-client/src/AppCollection.js
```diff
@@ -37,31 +35,6 @@ class AppCollection {
...
   }
-
-  find(...args) {
-    console.warn(
-      `find() method for doctype '${
-        this.doctype
-      }' relies internally on DocumentCollection.find() and may not fetch all expected data.`
-    )
-    return DocumentCollection.prototype.find.call(this, ...args)
-  }
```
drazik: You will lose the `console.warn`. If you want to keep it, you can use `super.find()`

packages/cozy-stack-client/src/AppCollection.js
gregorylegarec: Mmmmh I am not sure what we really want here. But maybe we should keep throwing while we have not decided.

PR #407, opened 2019-03-21
Client: Pass the expected Map of QueryDefinition

### LINKED ISSUE #404: [Major] Regression on relationships
Following https://github.com/cozy/cozy-client/pull/397, a regression has been introduced.

In [this line](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/associations/helpers.js#L84), it appears that `resp` may be `undefined`, and it's causing an error [when trying to access to `resp.data`](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/associations/helpers.js#L18)

gregorylegarec:
This PR fixes #404.

The wrong Map of QueryDefinition was passed to reconstructRelationShips.

I fixed but in reconstructRelationShips docIdAndRel can now be `undefined`, which is espacially weird. Testing it before reconstructing relationShip seems to do the job,
but I cannot be 100% sure that id does not introduce another issue. By the way, Banks seems to work again.

I am thinking of a whole refactoring of fetchRelationShips, this method is pretty huge and complex.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -465,7 +465,7 @@ class CozyClient {
     const relationshipsByDocId = reconstructRelationships(
       queryDefToDocIdAndRel,
       documents,
-      definitions,
+      optimizedDefinitions,
       responses
     )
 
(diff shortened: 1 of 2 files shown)


gregorylegarec:
Replaced by #411

PR #409, opened 2019-03-25
Client: Better AppColleciton and KonnectorCollection

gregorylegarec:
Get back to throwing method for not implemented endpoints for apps an konnectors.

Follows feedback from #406.

### DIFF
--- a/packages/cozy-stack-client/src/__tests__/AppCollection.spec.js
+++ b/packages/cozy-stack-client/src/__tests__/AppCollection.spec.js
@@ -2,6 +2,7 @@ jest.mock('../CozyStackClient')
 
 import CozyStackClient from '../CozyStackClient'
 import AppCollection from '../AppCollection'
+import DocumentCollection from '../DocumentCollection'
 import ALL_APPS_RESPONSE from './fixtures/apps.json'
 
 const FIXTURES = {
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/__tests__/AppCollection.spec.js
```diff
@@ -34,4 +35,58 @@ describe(`AppCollection`, () => {
...
+      await collection.find({ slug: 'test' })
+
+      expect(DocumentCollection.prototype.find).toHaveBeenLastCalledWith({
+        slug: 'test'
+      })
+    })
+  })
+
+  describe('get', () => {
+    it('throw error', async () => {
```
drazik: Psychorigidité :$

```suggestion
    it('should throw error', async () => {
```

packages/cozy-stack-client/src/__tests__/AppCollection.spec.js
```diff
@@ -34,4 +35,58 @@ describe(`AppCollection`, () => {
...
+        .mockImplementation(() => {})
+
+      const collection = new AppCollection(client)
+      await collection.find({ slug: 'test' })
+
+      expect(DocumentCollection.prototype.find).toHaveBeenLastCalledWith({
+        slug: 'test'
+      })
+    })
+  })
```
drazik: That's implementation detail IMHO. If `DocumentCollection.prototype.find` is well tested, I think we don't need this.

PR #410, opened 2019-03-25
WIP: Add RelationshipsFetcher

gregorylegarec:
Fresh code so please ignore comments for now.

Still WIP, even the test is not passing.

The goal of this PR is to:

* Provides an end to end test for relationships, based on the album example
* Propose an implementation with a dedicated class, as the whole algorithm is pretty complex.

The main difficulty was to keep a track of the optimized queries to be able to dispatch their response to the initial queries. For this I am using several properties of the `RelationshipsFetcher` class.

I need to  add more test and comments to clearly explicit everything.


gregorylegarec:
Replaced by https://github.com/cozy/cozy-client/pull/411/

PR #411, opened 2019-03-26
Client: Fix fetchRelationships

### LINKED ISSUE #404: [Major] Regression on relationships

gregorylegarec:
Fix #404

This PR reverts last changes on fetchRelationships:

* It does not handle asynchronous Association.query method, as `HasManyTriggers`  does not need it anymore (529e7f09dad0c80908c0b5854555e3475dcabd6c)
* It does not handle Association.query returning an array of documents (needs to  be fixed) (bbf809491a4526ad62c0e7e50bf0c7ca033f931e)
* It does handle same QueryDefinition instance returned by several `Association.query` methods (abee793c93edd84c3474fa3df58b077608d0f148)

This revert fixes the errors in Banks.

To match other Association implementations, `HasManyTrigger` evolves and now      returns a QueryDefinition.

During the optimisation step, queries are now deduplicated to avoid making a single trigger query for every konnectors.


ptbrowne:
Cool, it seems it is easier and brings deduplication not only for id queries. 🎉

packages/cozy-client/src/queries/optimize.js
```diff
@@ -25,7 +27,9 @@ const optimizeDoctypeQueries = queries => {
           ids: uniq(flatten(idQueries.map(q => q.id || q.ids)))
         })
       : []
-  return others.concat(groupedIdQueries)
+
+  // Deduplicate before concataining
+  return uniqWith(others, isEqual).concat(groupedIdQueries)
```
ptbrowne: Can you add a test demonstrating this ?

PR #412, opened 2019-04-02
feat: Remove warning

ptbrowne:
Since the route currently works, and the user does not have to do anything, I think we can remove the warning.


edas:
-> Ok to remove the warning as there is no action required and no real problem in using .find()

-> Ok if we want to add the `links` par from cozy-client but let's do a separate issue for that, and I am for leaving that aside (one can ask explicitly for icons if he wants to, so I am unsure an overide is needed)

PR #413, opened 2019-04-03
Add documentation about node environment

CPatchane:
CozyClient need a fetch method on node environment, we can add it using node-fetch

### DIFF
--- a/docs/how-tos.md
+++ b/docs/how-tos.md
@@ -36,6 +36,19 @@ ReactDOM.render(
 )
 ```
 
+### Use the client on Node environment (`ReferenceError: fetch is not defined`)
+
+Cozy-Client will rely on the [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) function included in browsers (or polyfilled), so we have to provide this same function for a Node script. An example is using [`node-fetch`](https://www.npmjs.com/package/node-fetch) like following:
+
+```js
+import fetch from 'node-fetch'
+import CozyClient from 'cozy-client'
+
+global.fetch = fetch
+```
+
+Then you will be able to use all the client methods and fetch data correctly.
+
 ### How to connect to the documents store declaratively ?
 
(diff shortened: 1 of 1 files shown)


docs/how-tos.md
```diff
@@ -36,6 +36,19 @@ ReactDOM.render(
 )
 ```
 
+### Use the client on Node environment (`ReferenceError: fetch is not defined`)
```
ptbrowne:
```suggestion
### How to use the client on Node environment (`ReferenceError: fetch is not defined`)
```

To be in line with other titles.

docs/how-tos.md
```diff
@@ -36,6 +36,19 @@ ReactDOM.render(
...
+
+Cozy-Client will rely on the [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) function included in browsers (or polyfilled), so we have to provide this same function for a Node script. An example is using [`node-fetch`](https://www.npmjs.com/package/node-fetch) like following:
```
ptbrowne:
```suggestion
Cozy-Client relies on the [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) function included in browsers (or polyfilled). This function does not exist on Node environments so we have to provide a polyfill. An example using [`node-fetch`](https://www.npmjs.com/package/node-fetch):
```

docs/how-tos.md
```diff
@@ -36,6 +36,19 @@ ReactDOM.render(
...
+
+```js
+import fetch from 'node-fetch'
+import CozyClient from 'cozy-client'
+
+global.fetch = fetch
+```
+
+Then you will be able to use all the client methods and fetch data correctly.
+
```
ptbrowne: Can you update the TOC at the top of the file please ?

docs/how-tos.md
ptbrowne: I splitted the phrase since clearer sentences are easier to understand.

PR #414, opened 2019-04-03
Fix documentation about node environment

CPatchane:
Not pushed for a previous PR before merging, sorry..

### DIFF
--- a/docs/how-tos.md
+++ b/docs/how-tos.md
@@ -1,6 +1,7 @@
 <!-- MarkdownTOC autolink=true -->
 
 - [How-to integrate with an existing store ?](#how-to-integrate-with-an-existing-store-)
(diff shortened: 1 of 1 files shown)


docs/how-tos.md
```diff
@@ -1,6 +1,7 @@
 <!-- MarkdownTOC autolink=true -->
 
 - [How-to integrate with an existing store ?](#how-to-integrate-with-an-existing-store-)
+- [How to use the client on Node environment ?](#use-the-client-on-node-environment-referenceerror-fetch-is-not-defined)
```
ptbrowne: The anchor is not correct since it has not the how-to

CPatchane: Sorry for those typos :/

PR #418, opened 2019-04-16
feat: Client emits events for log{in,out}

ptbrowne:
Emitting events during login/logout provides a flexible way
for components/objects using CozyClient to setup/destroy
themselves. It lets the setup to be decoupled from where
the login/logout happens.

I think it can be a good way for example to init the bar.

```
cozyClient.on('login', setupBar)
cozyClient.on('logout', destroyBar)
```

I would like to have your views on this.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -690,9 +690,21 @@ class CozyClient {
     if (stackClient) {
       this.stackClient = stackClient
     } else {
+      const options = {
+        ...this.options,
+        onTokenRefresh: token => {
+          this.emit('tokenRefreshed')
+          if (this.options.onTokenRefresh) {
+            deprecatedHandler(
+              `Using onTokenRefresh is deprecated, please use events like this: cozyClient.on('tokenUpdated', token => console.log('Token is updated', token)). https://git.io/fj3M3`
+            )
+            this.options.onTokenRefresh(token)
+          }
+        }
+      }
       this.stackClient = this.options.oauth
-        ? new OAuthClient(this.options)
-        : new CozyStackClient(this.options)
+        ? new OAuthClient(options)
+        : new CozyStackClient(options)
     }
 
     this.client = new Proxy(
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -32,6 +32,7 @@ import fromPairs from 'lodash/fromPairs'
 import flatten from 'lodash/flatten'
 import uniqBy from 'lodash/uniqBy'
 import get from 'lodash/get'
+import MicroEE from 'microee'
 
 const ensureArray = arr => (Array.isArray(arr) ? arr : [arr])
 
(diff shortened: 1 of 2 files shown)


Crash--:
I'm a big fan on an event solution, but it'll bring a new concept in Cozy-Client and I don't know if I like the idea. 

We already have `callback` concept in Cozy-Client, specially in `OAuthClient` with the `onTokenRefresh`. Is it something we can follow for this ?

```jsx 
const client = this.context.client
client.onLogin = () => setupBar()
```

drazik:
I like it, but I agree with @Crash--, we must go full events or full callbacks but mixing the two is weird. IMHO, events are great, I like your example.

ptbrowne:
Yes you are both right, having both is not good.

I would prefer to go full events since with the callback method you need to have all objects used in the callback ready at client instantation.

Crash--:
> I don't mind having both at the same time if it's a temporary thing.
Nothing is temporary :( 

Except the `onTokenResfresh` where do we use callbacks?

ptbrowne:
> Nothing is temporary :(

Yes but we can add deprecation warning since we will not be able to remove onTokenRefresh right away.

I think the best here would be to deprecate onTokenRefresh and add an event tokenRefreshed ?

Then we could merge ? WDYT ?

ptbrowne:
I'd be happy to hear what @gregorylegarec and @CPatchane think about this since the bar would be one of the primary consumer of login/logout events.

kosssi:
👍 I talk with @ptbrowne to add event on `onTokenRefresh` but I never see this PR ^^

I would like use this function to fix realtime on mobile app when token is not good.

packages/cozy-client/src/CozyClient.js
```diff
@@ -690,9 +690,21 @@ class CozyClient {
     if (stackClient) {
       this.stackClient = stackClient
     } else {
+      const options = {
+        ...this.options,
+        onTokenRefresh: token => {
+          this.emit('tokenUpdated', token)
```
ptbrowne: s/tokenUpdated/tokenRefreshed

Let's not introduce new words so that it is easier to transition ;)

ptbrowne: I think we should make it harder to get the token as it should be internal to cozyClient.

Something like `this.emit('tokenRefreshed', { dangerousToken: token })` ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -690,9 +690,21 @@ class CozyClient {
...
+          if (this.options.onTokenRefresh) {
+            deprecatedHandler(
+              `Using onTokenRefresh is deprecated, please use events like this: cozyClient.on('tokenUpdated', token => console.log('Token is updated', token)). https://git.io/fj3M3`
+            )
```
ptbrowne: 👍

packages/cozy-client/src/CozyClient.js
kosssi: I remove token to force the developer to go through the cozyClient to retrieve the token if he needs it.

PR #426, opened 2019-04-25
Retry update on conflict

gregorylegarec:
Allow to retrie update on 409 error :
fetch the actual document, merge it with the data we want to update, and try to perform the update again.

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -180,12 +180,30 @@ class DocumentCollection {
     }
   }
 
-  async update(document) {
-    const resp = await this.stackClient.fetchJSON(
-      'PUT',
-      uri`/data/${this.doctype}/${document._id}`,
-      document
-    )
+  async update(document, options = {}) {
+    const { retriesOnConflict = 0 } = options
+    let resp
+    try {
+      resp = await this.stackClient.fetchJSON(
+        'PUT',
+        uri`/data/${this.doctype}/${document._id}`,
+        document
+      )
+    } catch (error) {
+      const { status } = error
+      if (status === 409 && retriesOnConflict) {
+        const upToDateResp = await this.get(document._id)
+        const updatedDocument = {
+          ...upToDateResp.data,
+          ...omit(document, ['_rev'])
+        }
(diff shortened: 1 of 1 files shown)


packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -180,12 +180,30 @@ class DocumentCollection {
     }
   }
 
-  async update(document) {
-    const resp = await this.stackClient.fetchJSON(
-      'PUT',
-      uri`/data/${this.doctype}/${document._id}`,
-      document
-    )
+  async update(document, options = {}) {
```
Crash--: can you add a js doc about `options`

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -180,12 +180,30 @@ class DocumentCollection {
...
+    let resp
+    try {
+      resp = await this.stackClient.fetchJSON(
+        'PUT',
+        uri`/data/${this.doctype}/${document._id}`,
+        document
+      )
+    } catch (error) {
+      const { status } = error
+      if (status === 409 && retriesOnConflict) {
```
edas:
```suggestion
      if (status === 409 && retriesOnConflict && retriesOnConflict > 0) {
```

ptbrowne:
Apart from the suggestion from edas, the implementation seems OK. Please add test though.

packages/cozy-stack-client/src/DocumentCollection.js
CPatchane: A number less or equal than 0 isn't considered as a falsy value?

edas: `if (-1) { console.log("true") } else { console.log("false") }`
--> "true"

CPatchane: OK I was wrong, thanks 👍

gregorylegarec: Only `0` is falsy.

gregorylegarec:
I think this method has eventually too much impact, and shoud propbably not be done at this level. Let's think on it a little bit more before proposing a better solution.

PR #433, opened 2019-05-10
feat: Cozy-client will autologin if initialized with uri and token

edas:
I was surprised by https://github.com/cozy/cozy-bar/pull/545#discussion-diff-282847190R209

Most apps initialize Cozy-Client by providing an uri and a token to the constructor, without any explicit call to `login()`. This PR allows compatibility with all events and state variables like `isLogged` by calling `login()`automatically in this case.

The promise returned by `login()`is kept in case one explicitly call `login()` anyways so there should be no compatibility break

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -63,6 +63,8 @@ class CozyClient {
    * @param  {Array.Link}   options.links  - List of links
    * @param  {Object}       options.schema - Schema description for each doctypes
    * @param  {Object}       options.appMetadata - Metadata about the application that will be used in ensureCozyMetadata
+   *
+   * Cozy-Client will automatically call `this.login()` if provided with a token and an uri
    */
   constructor({ link, links, schema = {}, appMetadata = {}, ...options }) {
     if (link) {
@@ -86,6 +88,10 @@ class CozyClient {
     this.chain = chain(this.links)
 
     this.schema = new Schema(schema, this.getStackClient())
+
+    if (options.uri && options.token) {
+      this.login()
+    }
   }
 
   addSchema(schemaDefinition) {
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -60,6 +60,20 @@ describe('CozyClient initialization', () => {
...
+    expect(client.isLogged).toBeTruthy()
+  })
+
+  it('should not break explicit login when provided token and uri', () => {
+    const token = 'fake_token'
+    const uri = 'https://example.mycozy.cloud'
+    client = new CozyClient({ token, uri })
+    expect(client.login()).toBeInstanceOf(Promise)
+  })
+
```
ptbrowne: Can you add a test showing that a second login() returns the promise ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -120,12 +126,19 @@ class CozyClient {
    * @return {Promise} - Resolves when all links have been setup and client is fully logged in
    *
    */
-  async login(options) {
+  login(options) {
+    // keep the promise to be able to return it in future calls
+    // this allows us to autoLogin in constructor 
+    // without breaking any compatibility with codes
+    // that uses an explicit login
```
ptbrowne:
```
    // Keep the promise to be able to return it in future calls.
    // This allows us to autoLogin in constructor without breaking any compatibility
    // with codes that uses an explicit login.
```

ptbrowne:
LGTM !

packages/cozy-client/src/CozyClient.spec.js
edas: This *is* the second login (the first one is the automatic one). Do you want to test 2 explicit calls to `login`? (it didn't work before my PR so it was not really expected)

ptbrowne: Ah yes sorry :) All good then ;)

PR #435, opened 2019-05-11
refactor: Sync immediately without start/stop replication loop

ptbrowne:
After @edas review, I tried to improve the loop mechanism to be able to introduce task as soon as possible, something that was not possible in the previous loop.

Thus, the new Loop supports

- running asynchronous tasks periodically such as task_2 happens `<delay>` time after the end of task_1. For example, in a loop where `<delay>` is 30s, if task_1 is started a 0s and finishes 10s later, task_2 will be run at 40s.

- running tasks a soon a possible. Building in the previous example, if a task t_immediate is inserted immediately during the execution of task_1, it will be run *immediately* *after* task_1 finishes. task_2 will be run 30s after the end of t_immediate. If t_immediate is run during the 30 idle seconds, between task_1 and task_2, it will be run immediately, and task_2 will still be run 30s after the run of t_immediate.

I am very interested in reviews and possible way of improvements as I have a feeling that I am overengineering. Perhaps there is a simpler way.


packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -160,67 +159,68 @@ class PouchManager {
...
-      this._stopReplicationLoop()
-      this._stopReplicationLoop = null
+  /**
+   * If a replication is currently ongoing, will start a replication
+   * just after it has finished. Otherwise it will start a replication
+   * immediately
+   */
+  syncImmediately() {
+    if (this.loop) {
+      this.loop.scheduleImmediately()
```
edas: Doesn't appear to exist? Maybe `this.loop.scheduleImmediateTask()`

packages/cozy-pouch-link/src/loop.js
```diff
@@ -0,0 +1,139 @@
...
+      const task = this.immediateTasks.splice(0, 1)[0]
+      await this.runTask(task)
+    }
+  }
+
+  /**
+   * If loop started, schedules tasks that will be run in priority without
+   * delay at next round.
+   */
+  async scheduleImmediateTask(task) {
```
edas: If `task`is optional (it seems to be), maybe state it explicitly : `async scheduleImmediateTask(task=null)`

packages/cozy-pouch-link/src/loop.js
```diff
@@ -0,0 +1,139 @@
...
+    if (!this.started) {
+      console.warn('Cannot schedule immediate task, loop is not started')
```
y-lohse: We're not using logger here?

packages/cozy-pouch-link/src/PouchManager.js
ptbrowne: Fixed

packages/cozy-pouch-link/src/loop.js
ptbrowne: Fixed

packages/cozy-pouch-link/src/loop.js
```diff
@@ -0,0 +1,139 @@
...
+    return
+  }
+
+  /**
+   * Flushes the immediate tasks list and calls each task.
+   * Each task is awaited before the next is started.
+   */
+  async runImmediateTasks() {
+    while (this.immediateTasks.length > 0) {
+      const task = this.immediateTasks.splice(0, 1)[0]
```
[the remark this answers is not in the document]
ptbrowne: Fixed, I forgot about that !

ptbrowne:
OK for you @edas ?

PR #442, opened 2019-05-22
feat: Add Registry API

ptbrowne:
Ability to install/uninstall an app.

```js
import { Registry } from 'cozy-client'
const registryAPI = Registry({ client })
await registryAPI.installApp(app, source) 
```

Functions taken from the store.

### DIFF
--- a/packages/cozy-client/src/registry.js
+++ b/packages/cozy-client/src/registry.js
@@ -1,0 +1,63 @@
+import 'url-search-params-polyfill'
+import termUtils from './terms'
+import { APP_TYPE } from './constants'
+
+const queryPartFromOptions = options => {
+  const query = new URLSearchParams(options).toString()
+  return query ? `?${query}` : ''
+}
+
+const getBaseRoute = app => {
+  const { type } = app
+  // TODO node is an historic type, it should be `konnector`, check with the back
+  const route =
+    type === APP_TYPE.KONNECTOR || type === 'node' ? 'konnectors' : 'apps'
+  return `/${route}`
+}
+
+class Registry {
+  constructor(options) {
+    if (!options.client) {
+      throw new Error('Need to pass a client to instantiate a Registry API.')
+    }
+    this.client = options.client
+  }
+
+  /**
+   * Installs or updates an app from a source.
+   *
+   * Accepts the terms if the app has them.
+
+   * @param  {RegistryApp} app - App to be installed 
+   * @param  {string} source - String (ex: registry://drive/stable)
+   * @return {Promise}
+   */
+  async installApp(app, source) {
+    const { slug, type, terms } = app
+    const searchParams = {}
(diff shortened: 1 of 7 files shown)


packages/cozy-client/src/registry.js
```diff
@@ -0,0 +1,55 @@
...
+   * @return {Promise}
+   */
+  async installApp(app, source) {
+    const { slug, type, terms } = app
+    const searchParams = {}
+    const isUpdate = app.installed
+    if (isUpdate) searchParams.PermissionsAcked = isUpdate
+    if (source) searchParams.Source = source
+    const querypart = queryPartFromOptions(searchParams)
+    if (terms) {
```
edas: Shouldn't we call that explicitly in a distinct call ? I fear that we forget that we make our user accept implicitly and that we accept some terms in his name without him knowing.

packages/cozy-client/src/registry.js
```diff
@@ -0,0 +1,55 @@
...
+      await termUtils.save(this.client, terms)
+    }
+    const verb = app.installed ? 'PUT' : 'POST'
+    const route = type === APP_TYPE.KONNECTOR ? 'konnectors' : 'apps'
```
edas: DRY: This probably should be an helper somewhere. It is called 2 times and has its own internal logic.

packages/cozy-client/src/registry.js
```diff
@@ -0,0 +1,55 @@
...
+    return this.client.stackClient.fetchJSON(
+      verb,
+      `/${route}/${slug}${querypart}`
```
edas: Same here. DRY about the route base.

packages/cozy-client/src/registry.js
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Installs or updates an app from a source.
+   *
+   * Accepts the terms if the app has them.
+
+   * @param  {RegistryApp} app - App to be installed 
+   * @param  {string} source - String (ex: registry://drive/stable)
+   * @return {Promise}
+   */
+  async installApp(app, source) {
```
CPatchane: About this API, I am just thinking that it can't be used by any other apps except the Store since the Stack hardcoded the slug authorized to install applications

ptbrowne: Asked the back for a solution ;)

packages/cozy-client/src/registry.js
ptbrowne: Thanks, good catch. Fixed.

packages/cozy-client/src/registry.js
edas: Seen together. The frontend code should propose the terms and permissions to the user but there is hardly a meaning to install an app without accepting the terms. Let's leave that as is

PR #446, opened 2019-05-24
feat: Add app fetching

ptbrowne:
Add app fetching and manifest sanitization.

### DIFF
--- a/packages/cozy-client/src/manifest.js
+++ b/packages/cozy-client/src/manifest.js
@@ -1,0 +1,85 @@
+const AUTHORIZED_CATEGORIES = [
+  'banking',
+  'cozy',
+  'energy',
+  'health',
+  'host_provider',
+  'insurance',
+  'isp',
+  'mes_infos',
+  'online_services',
+  'partners',
+  'press',
+  'productivity',
+  'ptnb',
+  'public_service',
+  'shopping',
+  'social',
+  'telecom',
+  'transport'
+]
+
+/** Filters unauthorized categories. Defaults to ['others'] if no suitable category. */
+export function sanitizeCategories(categories) {
+  if (!categories) return ['others']
+  const filteredList = categories.filter(c => AUTHORIZED_CATEGORIES.includes(c))
+  if (!filteredList.length) return ['others']
+  return filteredList
+}
+
+export function areTermsValid(terms) {
+  return !!terms && !!terms.id && !!terms.url && !!terms.version
+}
+
+export function isParnershipValid(partnership) {
+  return !!partnership && !!partnership.description
+}
+
+/**
+ * Normalize app manifest, retrocompatibility for old manifests
+ *
+ * @param  {Manifest} manifest
+ * @return {Manifest}
+ */
+export function sanitize(manifest) {
+  const sanitized = { ...manifest }
+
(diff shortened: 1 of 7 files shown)


packages/cozy-client/src/manifest.js
```diff
@@ -0,0 +1,85 @@
...
+  const filteredList = categories.filter(c => AUTHORIZED_CATEGORIES.includes(c))
+  if (!filteredList.length) return ['others']
+  return filteredList
+}
+
+export function areTermsValid(terms) {
+  return !!terms && !!terms.id && !!terms.url && !!terms.version
+}
+
+export function isParnershipValid(partnership) {
```
drazik: `isPartnershipValid` (a `t` is missing)

packages/cozy-client/src/manifest.js
```diff
@@ -0,0 +1,85 @@
...
+  return !!partnership && !!partnership.description
```
drazik: Nit: I think `Boolean(partnership && partnership.description)` is more readable if you absolutely want to return a real boolean.

packages/cozy-client/src/manifest.js
```diff
@@ -0,0 +1,85 @@
+const AUTHORIZED_CATEGORIES = [
```
edas: categories of what ? I do not see that self-explainatory. Maybe APP_CATEGORY (and same thing for the function below).

+ I would externalize the list of categories in a json or in its own JS files

packages/cozy-client/src/manifest.js
```diff
@@ -0,0 +1,85 @@
...
+  'public_service',
+  'shopping',
+  'social',
+  'telecom',
+  'transport'
+]
+
+/** Filters unauthorized categories. Defaults to ['others'] if no suitable category. */
+export function sanitizeCategories(categories) {
+  if (!categories) return ['others']
```
edas: strange as "others" is not a valid category

packages/cozy-client/src/manifest.js
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @param  {Manifest} manifest
+ * @return {Manifest}
+ */
+export function sanitize(manifest) {
+  const sanitized = { ...manifest }
+
+  // Make categories an array and delete category attribute if it exists
+  if (
+    !manifest.categories &&
```
edas: It's probably not for this PR but I really dislike these options that we have to maintain a long time. Should we create an issue about deprecating the  `category` option and only leave `categories`?

ptbrowne: I do not know the status of the category option, I merely moved the code from the store here.

I have no strong opinion on the deprecation here, I do not know for how long the category option has been removed.

packages/cozy-client/src/manifest.js
```diff
@@ -0,0 +1,85 @@
...
+    delete sanitized.terms
+  }
+
+  // Remove invalid partnership
+  if (sanitized.partnership && !isPartnershipValid(sanitized.partnership)) {
+    delete sanitized.partnership
+  }
+
+  return sanitized
+}
```
Crash--, line 85: I didn't even know about this file. I think we should move it to `/models` directory. Can you move it and create a BC for that?

PR #449, opened 2019-05-28
fix: Allow fromEnv instantiation from oauth client

Gara64:
#447 could not work with the [cozy-konnector-dev](https://github.com/konnectors/libs/tree/master/packages/cozy-jobs-cli) script because it uses an OAuth authentication and `fromEnv` was expecting a string from `COZY_CREDENTIALS`
This PR fixes this behaviour.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -103,9 +103,19 @@ class CozyClient {
         'Env used to instantiate CozyClient must have COZY_URL and COZY_CREDENTIALS'
       )
     }
+    let oauth
+    let token
+    try {
+      oauth = JSON.parse(COZY_CREDENTIALS)
+      token = JSON.parse(COZY_CREDENTIALS)
+    } catch (e) {
+      token = COZY_CREDENTIALS.trim()
+    }
+
     return new CozyClient({
+      oauth: oauth,
       uri: COZY_URL.trim(),
-      token: COZY_CREDENTIALS.trim(),
+      token: token,
       ...options
     })
   }
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -95,27 +95,22 @@ class CozyClient {
   }
 
   /** In konnector/service context, CozyClient can be instantiated from environment variables */
-  static fromEnv(env, options) {
+  static fromEnv(env, options = {}) {
     env = env || (typeof process !== 'undefined' ? process.env : {})
-    const { COZY_URL, COZY_CREDENTIALS } = env
+    const { COZY_URL, COZY_CREDENTIALS, NODE_ENV } = env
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -103,9 +103,19 @@ class CozyClient {
         'Env used to instantiate CozyClient must have COZY_URL and COZY_CREDENTIALS'
       )
     }
+    let oauth
+    let token
+    try {
+      oauth = JSON.parse(COZY_CREDENTIALS)
+      token = JSON.parse(COZY_CREDENTIALS)
```
ptbrowne: Why do we have both ?

If parsing is successful, both oauth and token contain the same value, and I do not think the result is valable for the `token` option. Maybe you can remove the `token = JSON.parse` line ?

Can you add a test testing this behavior ?

edas: The logic should probably be from https://github.com/konnectors/libs/blob/db23f283acb10c2ed1651e07855495fe5396f583/packages/cozy-konnector-libs/src/libs/cozyclient.js#L14

packages/cozy-client/src/CozyClient.js
```diff
@@ -103,9 +103,19 @@ class CozyClient {
...
+    } catch (e) {
+      token = COZY_CREDENTIALS.trim()
+    }
+
     return new CozyClient({
+      oauth: oauth,
```
edas: https://github.com/konnectors/libs/blob/db23f283acb10c2ed1651e07855495fe5396f583/packages/cozy-konnector-libs/src/libs/cozyclient.js#L54

packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -84,6 +84,23 @@ describe('CozyClient initialization', () => {
...
+        accessToken: token
+      }
+    }
+    global.process.env.COZY_URL = url
+    global.process.env.COZY_CREDENTIALS = JSON.stringify(creds)
+    global.process.env.NODE_ENV = 'development'
+    client = CozyClient.fromEnv()
+    expect(client.stackClient.uri).toBe(url)
+    expect(client.stackClient.token.accessToken).toBe(token)
+    global.process.env.NODE_ENV = 'test'
```
ptbrowne: It is a bit unhygienic to let COZY_URL etc.. inside process.env after the test. Additionally, if one of the expect fails, the process.env.NODE_ENV is not set back to test.

I think you should use the second parameter env of fromEnv to mock the environment. 

```
client = CozyClient.fromEnv({ COZY_URL: 'blabla', COZY_CREDENTIALS: 'balbla' , NODE_ENV: 'bblabla' })
```

ptbrowne:
Thanks :tada: LGTM, you can merge :)

ptbrowne:
cc @doubleface

PR #451, opened 2019-05-29
feat: Implicitly create unknown schema

ptbrowne:
Schemas are necessary for relationships but are not really necessary for
simple doctypes. If a schema has not been added to the client, for now
we throw. I think it would be more ergonomic either to 

- return an empty object if we do not know the doctype.
- have a list of known doctypes that can be used without declaring them.

Here I added the `io.cozy.terms` doctype (the second solution). WDYT ? 

cc @edas ?

### DIFF
--- a/packages/cozy-client/src/Schema.js
+++ b/packages/cozy-client/src/Schema.js
@@ -100,11 +100,17 @@ class Schema {
 
   /**
    * Returns the schema for a doctype
+   *
+   * Creates an empty schema implicitly if it does not exist
    */
   getDoctypeSchema(doctype) {
-    const schema = this.byDoctype[doctype]
+    let schema = this.byDoctype[doctype]
     if (!schema) {
-      throw new Error(`No schema found for '${doctype}' doctype`)
+      schema = normalizeDoctypeSchema({
+        name: doctype,
+        doctype
+      })
+      this.byDoctype[doctype] = schema
     }
     return schema
   }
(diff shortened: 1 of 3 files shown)


edas:
I would would prefer to create explicitly the (empty) schema in cozy-libs/cozy-doctype and let the app declare it.

Anyways, you're right. There is no need to require something that is not necessary. Having a working cozy-client with no schema declared until we need a relationship is perfectly fine for me.

Having an empty object by default for every doctype without maintaining an explicit list seems a better option to me than this explicit default for io.cozy.terms. It will avoid maintaining this explicit list of doctype and have two places for doctypes, one in cozy-client for empty ones and one in cozy-libs/cozy-doctype for the others.

ptbrowne:
> Having an empty object by default for every doctype without maintaining an explicit list seems a better option to me than this explicit default for io.cozy.terms. It will avoid maintaining this explicit list of doctype and have two places for doctypes, one in cozy-client for empty ones and one in cozy-libs/cozy-doctype for the others.

OK for me also.

packages/cozy-client/src/Schema.js
```diff
@@ -102,7 +106,7 @@ class Schema {
    * Returns the schema for a doctype
    */
   getDoctypeSchema(doctype) {
-    const schema = this.byDoctype[doctype]
+    const schema = this.byDoctype[doctype] || knownSchemas[doctype]
```
edas:
```suggestion
    const schema = this.byDoctype[doctype] || { }
```

ptbrowne: Non pas si simple ;) Il faut le normaliser avant. Je m'en occupe :)

PR #452, opened 2019-05-29
Export cancelable util

cedricmessiant:
Export `cancelable` util directly to be able to do `import { cancelable } from 'cozy-client'`
I had issues trying to import it from `cozy-client/utils` (see #443 ) but it's maybe related to path resolution in drive


ptbrowne:
Yes it was neecessary to do import cancelable from 'cozy-client/dist/utils' before.

PR #453, opened 2019-05-30
feat: Use pouchdb-browser

ptbrowne:
Use a build of PouchDB without support for by default.


packages/cozy-client/src/__tests__/setup.js
```diff
@@ -1,4 +1,7 @@
 import Enzyme from 'enzyme'
 import Adapter from 'enzyme-adapter-react-16'
+import fetch from 'whatwg-fetch'
```
edas: Shouldn't we add it to the package.json ?

packages/cozy-pouch-link/package.json
```diff
@@ -30,5 +30,8 @@
     "watch": "yarn run build --watch",
     "prepublishOnly": "yarn run build"
   },
+  "jest": {
+    "setupFiles": ["./jest.setup.js"]
```
edas: I am not finding this file in the repo, am I missing something ?

ptbrowne: Well spotted. This file is not used as jest is run once for the whole workspace instead of being run in each package individually. 

I put the fetch polyfill in the general setup file that is from cozy-client/packages/cozy-client. I'm moving this file to be at the root of the directory, it will be clearer.

packages/cozy-client/src/__tests__/setup.js
ptbrowne: Done, thanks.

Crash--:
What are the impacts for a code using cozy-client and running on a nodejs? Like a connector or a service?

ptbrowne:
> What are the impacts for a code using cozy-client and running on a nodejs? Like a connector or a service?

Good question.

- If they do not use PouchLink there are no impacts.
- If they do use PouchLink, they have to manually add the adapter for leveldb https://www.npmjs.com/package/pouchdb-adapter-leveldb.

edas:
>     * If they do use PouchLink, they have to manually add the adapter for leveldb https://www.npmjs.com/package/pouchdb-adapter-leveldb.

I don't think we have a service or connector with a PouchLink persistance

Crash--:
Do we need to consider it as a breaking change?

edas:
> Do we need to consider it as a breaking change?

As for the version number, let's do that correctly but I am pretty sure we're not break anything. Persisting a leveldb in a connector or in a service is useless and won't work between requests anyways.

ptbrowne:
Yep, for me it is a breaking change, I update my commit to reflect this.

PR #457, opened 2019-05-31
feat: Add an unified way to get the access token

edas:
`client.getStackClient().getAccessToken()`

I was sick of looking for
```
  cozyClient.stackClient.token.accessToken
  ||
  cozyClient.stackClient.token.token
```


edas:
> Nice! Would you mind adding some JSDoc and maybe a test on `CozyStackClient.getAccessToken`?

I am a bit ashamed not to have done it in the first place. Fixed now

PR #459, opened 2019-06-05
Add cursor-based pagination

Gara64:
This adds the possibility to use cursor-based pagination, just like [this](https://github.com/cozy/cozy-client-js/pull/288).
For a query on 2185 files referenced by a single album, it takes 14.1s with skip-based pagination and 7.6s with cursor-based pagination.
Note the Couch query itself is greatly sped up, especially with large skip, e.g. 343ms for a `skip=2000` vs 20ms with cursor.

### DIFF
--- a/docs/api/cozy-client.md
+++ b/docs/api/cozy-client.md
@@ -710,6 +710,7 @@ from a Cozy. `QueryDefinition`s are sent to links.
     * [.include(includes)](#QueryDefinition+include) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
     * [.limitBy(limit)](#QueryDefinition+limitBy) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
     * [.offset(skip)](#QueryDefinition+offset) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
+    * [.offsetCursor(cursor)](#QueryDefinition+offsetCursor) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
     * [.referencedBy(document)](#QueryDefinition+referencedBy) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
 
 <a name="new_QueryDefinition_new"></a>
@@ -833,12 +834,31 @@ Maximum number of documents returned, useful for pagination. Default is 100.
 ### queryDefinition.offset(skip) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
 Skip the first ‘n’ documents, where ‘n’ is the value specified.
 
+Beware, this [performs badly](http://docs.couchdb.org/en/stable/ddocs/views/pagination.html#paging-alternate-method) on view's index.
+ Prefer cursor-based pagination in such situation.
+
 **Kind**: instance method of [<code>QueryDefinition</code>](#QueryDefinition)  
 **Returns**: [<code>QueryDefinition</code>](#QueryDefinition) - The QueryDefinition object.  
 
 | Param | Type | Description |
 | --- | --- | --- |
 | skip | <code>number</code> | The number of documents to skip. |
+
+<a name="QueryDefinition+offsetCursor"></a>
+
+### queryDefinition.offsetCursor(cursor) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
+Use [cursor-based](https://docs.cozy.io/en/cozy-stack/jsonapi/#pagination) pagination.
+The cursor is a [startkey, startkey_docid] array, where startkey is the view's key,
+e.g. ["io.cozy.photos.albums", "album-id"] and startkey_docid is the id of
+the starting document of the query, e.g. "file-id".
+Use the last docid of each query as startkey_docid to paginate or leave blank for the first query.
+
+**Kind**: instance method of [<code>QueryDefinition</code>](#QueryDefinition)  
+**Returns**: [<code>QueryDefinition</code>](#QueryDefinition) - The QueryDefinition object.  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| cursor | <code>Array</code> | The cursor for pagination. |
 
 <a name="QueryDefinition+referencedBy"></a>
 
--- a/docs/api/cozy-client.md
+++ b/docs/api/cozy-client.md
@@ -848,6 +848,7 @@ Beware, this [performs badly](http://docs.couchdb.org/en/stable/ddocs/views/pagi
 
 ### queryDefinition.offsetCursor(cursor) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
 Use [cursor-based](https://docs.cozy.io/en/cozy-stack/jsonapi/#pagination) pagination.
+*Warning*: this is only useful for views.
 The cursor is a [startkey, startkey_docid] array, where startkey is the view's key,
 e.g. ["io.cozy.photos.albums", "album-id"] and startkey_docid is the id of
 the starting document of the query, e.g. "file-id".
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -480,7 +480,8 @@ class CozyClient {
     }
     const withIncluded = await this.fetchRelationships(
       mainResponse,
-      this.getIncludesRelationships(definition)
+      this.getIncludesRelationships(definition),
+      definition.cursor
     )
     return withIncluded
   }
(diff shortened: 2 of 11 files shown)


docs/api/cozy-client.md
```diff
@@ -833,13 +834,32 @@ Maximum number of documents returned, useful for pagination. Default is 100.
...
 | skip | <code>number</code> | The number of documents to skip. |
 
+<a name="QueryDefinition+offsetCursor"></a>
+
+### queryDefinition.offsetCursor(cursor) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
+Use [cursor-based](https://docs.cozy.io/en/cozy-stack/jsonapi/#pagination) pagination.
+The cursor is a [startkey, startkey_docid] array, where startkey is the view's key,
+e.g. ["io.cozy.photos.albums", "album-id"] and startkey_docid is the id of
+the starting document of the query, e.g. "file-id".
+Use the last docid of each query as startkey_docid to paginate or leave blank for the first query.
```
ptbrowne: I do not understand how to find the view's key ?

packages/cozy-client/src/associations/HasManyFiles.js
```diff
@@ -3,12 +3,27 @@ import HasMany from './HasMany'
 import { QueryDefinition, Mutations } from '../queries/dsl'
 
 export default class HasManyFiles extends HasMany {
-  async fetchMore() {
-    const skip = this.getRelationship().data.length
+  async fetchMore(withCursor) {
```
ptbrowne: Using an options.cursor at true will be more readable IMHO , `fetchMore(true)` will not be very explicit. `fetchMore({ cursor: true })` will be easier to understand.

packages/cozy-client/src/associations/HasManyFiles.js
```diff
@@ -3,12 +3,27 @@ import HasMany from './HasMany'
...
     const queryDef = new QueryDefinition({ doctype: 'io.cozy.files' })
-    const response = await this.query(
-      queryDef.referencedBy(this.target).offset(skip)
-    )
+    const relationships = this.getRelationship().data
+    let response
+    if (withCursor) {
```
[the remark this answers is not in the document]
ptbrowne: If queryDef.sort is not defined, let's use cursors ?

docs/api/cozy-client.md
edas: It's how is created the view at the begining, when you `emit()`

packages/cozy-stack-client/src/querystring.js
```diff
@@ -1,5 +1,21 @@
 import pickBy from 'lodash/pickBy'
 
+const encodeValues = values => {
+  const encoded =
+    '[' +
+    encodeURIComponent(
+      values
+        .map(value => {
+          return Array.isArray(value)
+            ? '[' + value.map(arrayVal => `"${arrayVal}"`).join(',') + ']'
```
ptbrowne: The code in `encode` and `encodeValues` is very close, I wonder if we could use recursion here.

packages/cozy-client/src/associations/HasManyFiles.js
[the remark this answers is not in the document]
Gara64: @edas I'm not a big fan of this kind of magic, but ok if you think it's better this way.
@ptbrowne No, sort is used for mango query, not for views. Also, how do you access the queryDef of the previous query from fetchMore?

packages/cozy-stack-client/src/querystring.js
Gara64: Probably, but I feel like this should be another PR

ptbrowne: Why, it's code that is added in this PR ?

Gara64: I checked the code, I misunderstood what you were implying. So ok!

packages/cozy-client/src/associations/HasManyFiles.js
edas: 👍 with Patrick

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -68,16 +68,22 @@ class FileCollection extends DocumentCollection {
    * async findReferencedBy - Returns the list of files referenced by a document — see https://docs.cozy.io/en/cozy-stack/references-docs-in-vfs/
    *
    * @param  {object} document     A JSON representing a document, with at least a `_type` and `_id` field.
-   * @param  {number} { skip = 0   For pagination, the number of referenced files to skip
-   * @param  {number} limit } For pagination, the number of results to return.
+   * @param  {number} skip    For skip-based pagination, the number of referenced files to skip.
+   * @param  {number} limit For pagination, the number of results to return.
+   * @param  {object} cursor   For cursor-based pagination, the index cursor.
```
ptbrowne: options.skip
options.limit
options.cursor

Reading the doc I was confused, I thought there were 4 arguments to the function.

packages/cozy-client/src/CozyClient.js
```diff
@@ -483,7 +483,8 @@ class CozyClient {
     }
     const withIncluded = await this.fetchRelationships(
       mainResponse,
-      this.getIncludesRelationships(definition)
+      this.getIncludesRelationships(definition),
+      definition.cursor
```
ptbrowne: I am a bit confused, maybe I've missed something but it seems here that we are using the parent definition (the one to fetch albums) to store the `cursor` for a specific relationship. What happens if we have two relationships from the same query ? Shouldn't the `cursor` be specific to a single relationship ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -508,7 +509,12 @@ class CozyClient {
 
     responseDocs.forEach(doc => {
       return forEach(relationshipsByName, (relationship, relName) => {
-        const queryDef = relationship.type.query(doc, this, relationship)
+        const queryDef = relationship.type.query(
+          doc,
+          this,
+          relationship,
+          cursor
+        )
```
ptbrowne: I feel a bit uneasy here, I am not sure that having `cursor` at this point is necessary.

packages/cozy-client/src/associations/HasManyFiles.js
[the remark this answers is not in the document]
ptbrowne: The management of skip was done automatically, looking at the state of the relationship to automatically add the skip/limit. I think it should be the same.

> Also, how do you access the queryDef of the previous query from fetchMore?

I think for the relationship to be able to access the part of the queryDefinition that is relevant to itself is expected. It may not be possible at the moment, but it would not disturb me to have

```
/** Returns the options from the parent query related to the relationship */
relationship.getQueryDefinitionPart()
```

For now, the DSL for relationship is very simple and does not allow to pass options but we could have:

```
// simple case, already supported
client.all('io.cozy.albums').include(['io.cozy.files'])

// more complex case where we need the files to be returned sorted
client.all('io.cozy.albums').include([{ doctype: 'io.cozy.files', sort: 'datime' ])
```

docs/api/cozy-client.md
ptbrowne: > It's how is created the view at the begining, when you emit()

It means those cursors are specific to Cozy ?

edas: > It means those cursors are specific to Cozy ?

No. It means you hav to know how the view was built in the first place (ie: what is indexed and how.

If your view is built on `[fieldA, filedB, filedC]` then your cursor need to be `[fieldA_value, fieldB_value, fieldC, value], docid`

packages/cozy-client/src/queries/dsl.js
```diff
@@ -150,8 +154,21 @@ class QueryDefinition {
   }
 
   /**
-   * Use the [file reference system](https://docs.cozy.io/en/cozy-stack/references-docs-in-vfs/)
+   * Use [cursor-based](https://docs.cozy.io/en/cozy-stack/jsonapi/#pagination) pagination.
+   * The cursor is a [startkey, startkey_docid] array, where startkey is the view's key,
+   * e.g. ["io.cozy.photos.albums", "album-id"] and startkey_docid is the id of
+   * the starting document of the query, e.g. "file-id".
+   * Use the last docid of each query as startkey_docid to paginate or leave blank for the first query.
```
ptbrowne: Add a warning that this can only be used with views.

nono:
Ping @edas @ptbrowne

PR #464, opened 2019-06-12
Support for file metadata creation

Gara64:
Add support for this functionality: https://github.com/cozy/cozy-stack/blob/master/docs/files.md#post-filesuploadmetadata

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -155,7 +155,10 @@ class FileCollection extends DocumentCollection {
     return this.createFile(data, { dirId })
   }
 
-  createFile(data, { name, dirId = '', executable, ...options } = {}) {
+  async createFile(
+    data,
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -167,7 +170,12 @@ class FileCollection extends DocumentCollection {
     if (executable === undefined) {
       executable = false
     }
-    const path = uri`/files/${dirId}?Name=${name}&Type=file&Executable=${executable}`
+    let metadataId = ''
```
y-lohse: Should this be `null` by default, so that a file without `metadata` doesn't end up with a `metadata` field?

Gara64: If `metadataID` is empty, it will be ignored by the stack: https://github.com/cozy/cozy-stack/blob/master/web/files/files.go#L1189 , so there won't be any `metadata` field.
However a null value will be interpreted as 'null' string by the stack, causing an error.

PR #468, opened 2019-06-19
Put proptypes on wrapped component instead of wrapper

CPatchane:
Proptypes was put onto the wrapper instead of the Wrapped.component, causing undefined proptypes issues with React

### DIFF
--- a/packages/cozy-client/src/withMutations.jsx
+++ b/packages/cozy-client/src/withMutations.jsx
@@ -2,6 +2,24 @@ import merge from 'lodash/merge'
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
+const makeMutationsObject = (mutations, client, props) => {
+  return merge(
+    ...mutations.map(
+      mutations =>
+        typeof mutations === 'function' ? mutations(client, props) : mutations
+    )
+  )
+}
+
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/withMutations.jsx
```diff
@@ -2,6 +2,24 @@ import merge from 'lodash/merge'
...
+    ...mutations.map(
+      mutations =>
+        typeof mutations === 'function' ? mutations(client, props) : mutations
+    )
+  )
+}
+
+const mutationsToPropTypes = mutations => {
+  return Object.keys(makeMutationsObject(mutations)).reduce(
+    (acc, mutatioName) => {
```
drazik:
```suggestion
    (acc, mutationName) => {
```

packages/cozy-client/src/withMutations.jsx
```diff
@@ -2,6 +2,24 @@ import merge from 'lodash/merge'
...
+      return { ...acc, [mutatioName]: PropTypes.func.isRequired }
```
drazik:
```suggestion
      return { ...acc, [mutationName]: PropTypes.func.isRequired }
```

PR #470, opened 2019-06-21
Stack: Improve Not Found Error handling in Document Collection

gregorylegarec:
This PR: 
* Make `DocumentCollection.get()` throw error if it's not a "Not Found" error. Could be a **BREAKING CHANGE**.
* Update the function `dontThrowNotFoundError` to make it able to return a specific value, for example `null` for method fetching only one document and not a list. Use this update for `DocumentCollection.get()`

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.spec.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.spec.js
@@ -468,6 +468,22 @@ describe('DocumentCollection', () => {
     })
   })
 
+  describe('get', () => {
+    afterEach(() => {
+      jest.resetAllMocks()
+    })
+
+    it('should throw', async () => {
+      jest.spyOn(client, 'fetchJSON').mockImplementation(() => {
+        throw new Error('Not a 404 error')
+      })
+      const collection = new DocumentCollection('io.cozy.todos', client)
+      await expect(collection.get('whatever_id')).rejects.toEqual(
+        new Error('Not a 404 error')
+      )
+    })
+  })
+
   describe('update', () => {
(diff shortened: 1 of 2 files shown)


nono:
Ping @gregorylegarec

Crash--:
Can we have some context about this change?

gregorylegarec:
@Crash-- If the call to fetchJson in Document.get() was failing, no error was thrown.

packages/cozy-stack-client/src/DocumentCollection.spec.js
```diff
@@ -468,6 +468,22 @@ describe('DocumentCollection', () => {
...
+
+    it('should throw', async () => {
+      jest.spyOn(client, 'fetchJSON').mockImplementation(() => {
+        throw new Error('Not a 404 error')
+      })
+      const collection = new DocumentCollection('io.cozy.todos', client)
+      await expect(collection.get('whatever_id')).rejects.toEqual(
+        new Error('Not a 404 error')
+      )
+    })
```
ptbrowne: Can we have a test for the case where it is a not_found error. Here since there is change with the `null` I think we expect to get `{ data: null }` but I think we will have `{ data: undefined }` since the result of `dontThrowNotFoundError(error, null)` is not used.

ptbrowne:
Ping @gregorylegarec , the build is red and I had one suggestion for a test.

gregorylegarec:
@ptbrowne Thanks for the reminder!

gregorylegarec:
As I am leaving a little sooner as expected, I will not have time to finish this non-mandatory RP. You guys can keep it as reference if you feel the need to ignore Not Found Error in Document Collection.

PR #479, opened 2019-07-05
Add deleteFilePermanently method

cedricmessiant:
Discussed on the mattermost channel a few days ago. Uses the route from the stack : https://docs.cozy.io/en/cozy-stack/files/#patch-filesfile-id-and-patch-filesmetadata

(well, maybe we could name it `permanentDelete` here too)

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -150,6 +150,26 @@ class FileCollection extends DocumentCollection {
     return this.stackClient.fetchJSON('POST', uri`/files/trash/${id}`)
   }
 
+  /**
+   * async deleteFilePermanently - Definitely delete a file
+   *
+   * @param  {string} id - The id of the file to delete
+   * @returns {object} The deleted file object
+   */
+  async deleteFilePermanently(id) {
+    const resp = await this.stackClient.fetchJSON('PATCH', uri`/files/${id}`, {
+      data: {
+        type: 'io.cozy.files',
+        id,
+        attributes: {
+          permanent_delete: true
+        }
+      }
+    })
+
+    return resp.data
+  }
+
   async upload(data, dirPath) {
     const dirId = await this.ensureDirectoryExists(dirPath)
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -150,6 +150,26 @@ class FileCollection extends DocumentCollection {
     return this.stackClient.fetchJSON('POST', uri`/files/trash/${id}`)
   }
 
+  /**
+   * async definitelyDeleteFile - Definitely delete a file
```
ptbrowne: deleteFilePermanently is better IMHO. :)

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -150,6 +150,26 @@ class FileCollection extends DocumentCollection {
...
+   * @param  {string} id - The id of the file to delete
+   * @returns {object} The deleted file object
+   */
+  async definitelyDeleteFile(id) {
+    const resp = await this.stackClient.fetchJSON('PATCH', uri`/files/${id}`, {
+      data: {
+        type: 'io.cozy.files',
+        id,
+        attributes: {
+          permanent_delete: true
```
ptbrowne: And it would be coherent with the stack argument.

packages/cozy-stack-client/src/FileCollection.js
cedricmessiant: it would have been an awful last PR so I fixed it ;-)

cedricmessiant: Thanks !

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -150,6 +150,26 @@ class FileCollection extends DocumentCollection {
     return this.stackClient.fetchJSON('POST', uri`/files/trash/${id}`)
   }
 
+  /**
+   * async deleteFilePermanently - Definitely delete a file
```
Crash--: Without trashing it first ?

PR #481, opened 2019-07-09
 feat: IsChild 

Gara64:
Add a method to check if a file/directory belongs to a parent's hierarchy, based on its _id

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -253,6 +253,37 @@ class FileCollection extends DocumentCollection {
     return resp.links.related
   }
 
+  /**
+   * Checks if the file belongs to the parent's hierarchy.
+   *
+   * @param  {string}  childFileID  The file ID of the child
+   * @param  {string}  childDirID   The dirID of the child
+   * @param  {string}  childPath    The child path
+   * @param  {string}  parentID     The parent's document id
+   * @return {boolean}              Whether the file is a parent's child
+   */
+  async isChild(childFileID, childDirID, childPath, parentID) {
+    if (childFileID === parentID || childDirID === parentID) {
+      return true
+    }
+    const paths = childPath.split('/')
+    paths.shift() // The path starts with '/'
+
+    const targetsPath = ['/' + paths[0]]
+    for (let i = 1; i < paths.length; i++) {
+      const newPath = targetsPath[i - 1] + '/' + paths[i]
+      targetsPath.push(newPath)
+    }
+
+    // Look for all hierarchy in parallel
+    const results = await Promise.all(
+      targetsPath.map(path => this.statByPath(path))
+    )
+    return results.some(dir => {
+      return dir.data._id === parentID
+    })
+  }
+
   async statById(id, options = {}) {
     const { limit, skip = 0 } = options
     const params = {
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -21,6 +21,24 @@ export const isFile = ({ _type, type }) =>
 
 export const isDirectory = ({ type }) => type === 'directory'
 
+const raceWithCondition = (promises, predicate) => {
+  return new Promise(resolve => {
+    promises.forEach(p =>
+      p.then(res => {
+        if (predicate(res)) {
+          resolve(true)
+        }
+      })
+    )
+    Promise.all(promises).then(() => resolve(false))
+  })
+}
+
+const dirName = path => {
+  const lastIndex = path.lastIndexOf('/')
+  return path.substring(0, lastIndex)
+}
+
 /**
  * Implements `DocumentCollection` API along with specific methods for
  * `io.cozy.files`.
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -253,6 +253,37 @@ class FileCollection extends DocumentCollection {
...
+  /**
+   * Checks if the file belongs to the parent's hierarchy.
+   *
+   * @param  {string}  childFileID  The file ID of the child
+   * @param  {string}  childDirID   The dirID of the child
+   * @param  {string}  childPath    The child path
+   * @param  {string}  parentID     The parent's document id
+   * @return {boolean}              Whether the file is a parent's child
+   */
+  async isChild(childFileID, childDirID, childPath, parentID) {
```
edas: 4 parameters, whose order could not be anticipated without the documentation. We may find simething better.

I would say `isChildOf(childId|doc, parentID|doc)` 
- if one gives a doc, then you have the path and the id
- if one gives only the id, then you can send a `stat` request to have the path

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -253,6 +253,37 @@ class FileCollection extends DocumentCollection {
...
+    if (childFileID === parentID || childDirID === parentID) {
```
edas: It's ok to have this shortcut if we have the data. However it seems you are requesting the childDirId only to test this shortcut. Could it be optional ?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -253,6 +253,37 @@ class FileCollection extends DocumentCollection {
...
+      return true
+    }
+    const paths = childPath.split('/')
```
edas: It may be better to use a method like path#dirname in a loop rather than a manual split. I don't want to miss some complex cases

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -253,6 +253,37 @@ class FileCollection extends DocumentCollection {
...
+    paths.shift() // The path starts with '/'
+
+    const targetsPath = ['/' + paths[0]]
+    for (let i = 1; i < paths.length; i++) {
+      const newPath = targetsPath[i - 1] + '/' + paths[i]
+      targetsPath.push(newPath)
+    }
+
+    // Look for all hierarchy in parallel
+    const results = await Promise.all(
```
edas: This will wait for the last request to fullfill even if we have a successfull one already fullfilled. Could we implement a shortcut there ?

Gara64: How do you see the shortcut? We didn't see any with @CPatchane as we cannot cancel promises

packages/cozy-stack-client/src/FileCollection.js
Gara64: Ok with your proposition, but I'm a bit surprised by `not be anticipated without the documentation`: how can one use a function without knowing its signature, known by the code or the doc?

packages/cozy-stack-client/src/FileCollection.js
Gara64: Ok, so based on your previous proposition, `isChildOf(childId|doc, parentID|doc, { dirID } = {})` ?

packages/cozy-stack-client/src/FileCollection.js
Crash--: > The Promise.race() method returns a promise that fulfills or rejects as soon as one of the promises in an iterable fulfills or rejects, with the value or reason from that promise.

something like ? 
```
await Promise.race(targetsPath.map(path => this.statByPath(path).then(result => if(result.data._id === parentID) resolve()))
```

packages/cozy-stack-client/src/FileCollection.js
Gara64: Ok with that, but what do you see as potential complex cases?

packages/cozy-stack-client/src/FileCollection.js
edas: Something like (not tested)

```js
return new Promise( async (resolve) => {
  await Promise.all(
    targetsPath.map(
      async (path) => {
        const stat = await this.statByPath(path)
        if (stat.data._id == parentID) {
          resolve(true)
        }
      }
    )
  )
  resolve(false)
})
```

edas: Promise.race takes the first to resolve or reject. We want to take the first to resolve and discard all rejections, or take the first to resolve true and discard those which resolve to false.

We probably need the Promise.all, but we can return without waiting for it to finish

Gara64: @edas Nice, your solution seems to work!

ptbrowne: I like the `race` idea and I think it would be cleaner if the race part was factored out of the business logic.


```js
const raceWithCondition = (promises, predicate, default) => {
  return new Promise((resolve, reject) => {
    let rejections = 0
    const handleError = err => ++rejections === promises.length && reject(e)
    for (const p of promises) {
      p.then(res => predicate(res) && resolve(res)).catch(handleError)
    }
    Promise.all(promises).then(() => resolve(default))
  })
}

const res = await raceWithCondition(
  targetsPath.map(path => this.statByPath(path)),
  stat => stat.data._id == parentID
)
```

edas: Love it, thanks

packages/cozy-stack-client/src/FileCollection.js
edas: The rules is that once you know the method, everything should be pretty obvious to read or to write. When you have more than one optionnal parameter, or too many mandatory ones, or when you can't really tell which parameter is what without looking at the doc, you probably should go to named ones (which in JS is by object destructuring)

In our case, the name of the method does not make obvious what and where each parameter  should be others than `child` and `parent`

packages/cozy-stack-client/src/FileCollection.js
edas: Ok for me if you think you will really benefit of this because you often know the parentId

packages/cozy-stack-client/src/FileCollection.js
edas: I don't know (and that's why I would prefer delegating that to a lib).

packages/cozy-stack-client/src/FileCollection.js
ptbrowne: I was trying to see if bluebird had this natively. I stumbled upon this little comment in http://bluebirdjs.com/docs/api/promise.race.html which is interesting :

> This method is only implemented because it's in the ES6 standard. If you want to race promises to fulfillment the .any method is more appropriate as it doesn't qualify a rejected promise as the winner. It also has less surprises: .race must become infinitely pending if an empty array is passed but passing an empty array to .any is more usefully a RangeError

PR #485, opened 2019-07-15
feat(cozy-stack-client): Updatefile can create metadata

Crash--:
Same as https://github.com/cozy/cozy-client/commit/c9dbb6f50fe14e9084f8af62d955602b7203e537 but for updateFile. 

We can now attach a metadata to the updatedFile.

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -188,10 +188,14 @@ class FileCollection extends DocumentCollection {
    * @param  {object}  params             Additional parameters
    * @param  {string}  params.fileId      The id of the file to update (required)
    * @param  {boolean} params.executable  Whether the file is executable or not
+   * @param  {object}  params.metadata    Metadata to be attached to the File io.cozy.file
    * @param  {object}  params.options     Options to pass to doUpload method (additional headers)
    * @return {object}                     Updated document
    */
-  updateFile(data, { executable = false, fileId, ...options } = {}) {
+  async updateFile(
+    data,
+    { executable = false, fileId, metadata, ...options } = {}
+  ) {
     if (!fileId || typeof fileId !== 'string') {
       throw new Error('missing fileId argument')
     }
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/FileCollection.spec.js
```diff
@@ -331,6 +331,43 @@ describe('FileCollection', () => {
...
+      client.fetchJSON.mockReturnValueOnce({
+        data: {
+          id: metadataId
+        }
+      })
+      const data = new File([''], 'mydoc.odt')
+      const params = {
+        fileId: '59140416-b95f',
+        checksum: 'a6dabd99832b270468e254814df2ed20',
+        metadata: { type: 'bill' }
```
ptbrowne: [Here](https://github.com/cozy/cozy-doctypes/blob/a83a83df999fd64c7246308120f9928430aed6b4/docs/README.md}, it's written that the metadata key is `cozyMetadata`. Are files a special case here due to history ?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -206,7 +210,14 @@ class FileCollection extends DocumentCollection {
       throw new Error('missing name argument')
     }
 
-    const path = uri`/files/${fileId}?Name=${name}&Type=file&Executable=${executable}`
+    let metadataId = ''
```
```suggestion
    let metadataId
```

It does not seem necessary to initialize to empty string

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -206,7 +210,14 @@ class FileCollection extends DocumentCollection {
...
+    let path = uri`/files/${fileId}?Name=${name}&Type=file&Executable=${executable}`
+    if (metadata) {
+      const meta = await this.createFileMetadata(metadata)
+      metadataId = meta.data.id
+      path = path + `&MetadataID=${metadataId}`
```
ptbrowne: It surprises me that we use a path component to add the metadata id to the io.cozy.files instead of using POST data. What is the reason behind this ?

packages/cozy-stack-client/src/FileCollection.spec.js
Crash--: Files are special anyway, this is why we have a FileCollection. But Files also have a new "metadata" attributes (https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.files.md#files) to store metadata not related to cozy. For instance we can store the type of the file (not PDF or PNG, but its business type like 'bill' or 'identity document' or 'payslip'...). 

Just saw that the PR is not merged yet : https://github.com/cozy/cozy-doctypes/pull/101

packages/cozy-stack-client/src/FileCollection.js
Crash--: We already use the body to send the content of the file. So we have 2 choices : 
- Use an object in a query string to send the metadata 
- create a new header http 

In both case, we have a size limitation depending of the browser. So we had this current workaround where we create the metadata before (no size limit since we can use the body for that) and after we use the ID.

packages/cozy-stack-client/src/FileCollection.spec.js
edas: We had discussions about how to store and index times. This should have no impact on your PR

packages/cozy-stack-client/src/FileCollection.js
ptbrowne: I think a comment with a part of this explanation would be beneficial.

PR #488, opened 2019-08-01
feat(cozy-stack-client): We can find on TriggerCollection

Crash--:
`/triggers/` route is a bit particular. ... 

If we fetch `/triggers`, we can only pass one query string : `worker` to find the wanted triggers. It is not enough when we want to perform a more complex search with more complex selectors.

So we have to used `data/io.cozy.triggers/_find` to make a complex search, so we just need to extends `DocumentsCollection`

Be aware that the both routes do not return the same value. See https://github.com/cozy/cozy-stack/pull/2010

Now we can use :
``` js
triggerCollection.find({worker:'konnector', message:{konnector: 'trainline'}})
client.find('io.cozy.triggers').where({worker:'konnector', message:{konnector: 'trainline'}})
```


Crash--:
@y-lohse Done ;).

packages/cozy-stack-client/src/TriggerCollection.spec.js
```diff
@@ -85,17 +158,42 @@ describe('TriggerCollection', () => {
...
+
+    it('should call /data/io.cozy.triggers/_find route if multiple arguments are passed', async () => {
+      stackClient.fetchJSON.mockReturnValueOnce(
+        Promise.resolve({
+          id: '_design/123456',
+          name: '123456',
+          result: 'exists'
+        })
+      )
+      stackClient.fetchJSON.mockReturnValue(FIND_RESPONSE_FIXTURES)
```
y-lohse: Since we're not using the `FIND_RESPONSE_FIXTURES ` in the test, do we really need them?

Crash--: Yep since we're doing all the work in `.find()` method below, we need to return values with the right attributes. Otherwise, it'll throw an error I think.

Issue #489, opened 2019-08-02
cozy-client: must include react to use it


doubleface:
Even if react is now a peerDependency, cozy-konnector-libs must include react when using cozy-client since index.js imports Provider.js which imports react

ptbrowne:
import CozyClient from "cozyClient/dist/CozyClient" marchera non ?

edas:
I don't understand why it's an issue in cozy-client.

You have a peerDep. If you need it in cozy-konnector-libs, include it in cozy-konnector-libs. If it's depending on what you use in cozy-konnector-libs, then it's up to the konnector to include it or not.

ptbrowne:
I think it is understandably a problem for the user of cozy-konnector-libs to have to do:

```
yarn add cozy-client react
```

when (s)he only wants to use

```
yarn add cozy-client
```

I think using directly the `import CozyClient from "cozy-client/dist/CozyClient"` will solve the problem, maybe an improvement would be to allow `import CozyClient from "cozy-client/CozyClient"`. If we added this in the doc, would it be ok for you @doubleface ?

doubleface:
It's just I did not see the need to put react as a peerDep if you must include react to use cozy-client even if you do not use any react feature

ptbrowne:
> It's just I did not see the need to put react as a peerDep if you must include react to use cozy-client even if you do not use any react feature

Being a peer dep does not mean the dependency is optional. It means that it is the responsibility of the host to install it. It is because cozy-client is used in the context of the React of the application, it does not bring its own version of React, hence we tell yarn/npm not to install a React.

drazik:
We can introduce a `react` entrypoint, so we can `import CozyClient from 'cozy-client'` no matter if we use react or not. And `import { Provider, Query } from 'cozy-client/react'` if we use react.

edas:
> We can introduce a `react` entrypoint, so we can `import CozyClient from 'cozy-client'` no matter if we use react or not. And `import { Provider, Query } from 'cozy-client/react'` if we use react.

I would vote for this but it would also break every react app client of cozy-client.

Would it simply be possible to change Provider.js so it include React through a `require()` with a try/catch and mark it as an optionalDependency?

drazik:
> I would vote for this but it would also break every react app client of cozy-client.

I bet we can write a codemod and ship it with the breaking release. I will try

ptbrowne:
[...]

PR #494, opened 2019-08-19
WIP: refactor(cozy-client): Put react exports in a react entry point

drazik:
The following exports are now located in a react specific entry point:

* `Provider`
* `Query`
* `withMutation`
* `withMutations`
* `queryConnect`
* `withClient`
* `connect`

You should import these modules from `cozy-client/react`. To achieve
this automatically, we ship a codemod that you can run on your codebase
using the following command:

```
npx jscodeshift --parser babel -t ./node_modules/cozy-client/codemods/split-react-exports.js --extensions js,jsx src/
```

This is a WIP because in the current form, `import ... from 'cozy-client/react'` doesn't work, the real import is `import ... from 'cozy-client/dist/react`, which is not ideal. I will search if this is possible to do this but I am not that optimist about that. Anyway, if this is not possible, I think this would still be a good way to avoid problems like https://github.com/cozy/cozy-client/issues/489.

### DIFF
--- a/packages/cozy-client/codemods/split-react-exports.js
+++ b/packages/cozy-client/codemods/split-react-exports.js
@@ -1,0 +1,60 @@
+const groupBy = (arr, fn) => {
+  const res = {}
+  for (let item of arr) {
+    const groupKey = fn(item)
+    res[groupKey] = res[groupKey] || []
+    res[groupKey].push(item)
+  }
+  return res
+}
+
+const reactSpecificIdentifiers = [
+  'CozyProvider',
+  'Query',
+  'withClient',
+  'withMutation',
+  'withMutations',
+  'queryConnect',
+  'withClient',
+  'connect'
+]
+
+const isReactSpecifier = spec => {
+  if (!spec.imported) {
+    return false
+  }
+  return reactSpecificIdentifiers.includes(spec.imported.name)
+}
+
+export default function transformer(file, api) {
+  const j = api.jscodeshift
+  const root = j(file.source)
+
+  const importDeclarations = root.find(j.ImportDeclaration, {
+    source: {
+      value: 'cozy-client'
+    }
+  })
+
+  importDeclarations.forEach(path => {
+    const node = path.node
+
+    const {
+      true: reactSpecifiers = [],
+      false: nonReactSpecifiers = []
+    } = groupBy(node.specifiers, isReactSpecifier)
+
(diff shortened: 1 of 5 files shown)


packages/cozy-client/codemods/split-react-exports.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+const reactSpecificIdentifiers = [
+  'CozyProvider',
+  'Query',
+  'withClient',
+  'withMutation',
+  'withMutations',
+  'queryConnect',
+  'withClient',
```
ptbrowne:
```suggestion
```

ptbrowne:
Indeed it does not seem possible to have `cozy-client/react` instead of `cozy-client/dist/react`.

See 

- https://github.com/nodejs/node/issues/14970 for the closed issue about a `mainDir` property in package.json
- https://github.com/sindresorhus/np for a solution at publish (see --contents), I fear this will not work when linking the library, see other comments on the above issue for why this can be a bad idea.

This is a bit of a bummer but I think we should be ok.

edas:
It should be possible. In the build process, copy the package.json to /dist and modify the copy so its `main` field so it points to './index.js' and not to '/dist/index.js'

In the publish process, only publish the /dist content as root and not your full project directory

drazik:
Actually the current solution is uselessly complicating the usage of cozy-client with react. Another approach discussed with @edas and @ptbrowne would be to leverage the `package.json` [main](https://docs.npmjs.com/files/package.json#main) and [browser](https://docs.npmjs.com/files/package.json#browser) fields to ship the different entry points. main entrypoint would the the one containing cozy-client without react bindings; browser entrypoint would the the one containg cozy-client with react bindings.

The documentation must reflect that react bindings can still be imported by specifying `cozy-client/dist/react` anyway.

We must ensure that when specifying the `node` target, webpack loads the `main` entry point, since services and konnectors are bundled using webpack.

ptbrowne:
> We must ensure that when specifying the node target, webpack loads the main entry point, since services and konnectors are bundled using webpack.

I already tested and this works well, see https://github.com/webpack/webpack/blob/9741565588c64f82d942ec1096f0296ffd9443f4/lib/WebpackOptionsDefaulter.js#L69

drazik:
Closing in favor or https://github.com/cozy/cozy-client/pull/496

PR #506, opened 2019-08-28
chore: Remove webpack deps

ptbrowne:
Thanks to renovate, we saw with drazik that we had webpack and webpack-cli as dependencies and
I do not think they are used.

After removal of the dependencies and the removal of webpack in the doc (2nd commit), we have
no mention of webpack in the repository.


ptbrowne:
Done in #512

PR #515, opened 2019-09-04
fix: Don't memoize if the result is an error

Crash--:
Currently, if we render an `AppIcon` component with a `fetchIcon` props calling  :
```jsx
client.getStackClient().getIconURL({
        type: 'konnector',
        slug: konnector.slug
      })
```

When first call this getIconURL being offline, it'll memoize the return of this function aka an error. If we're now online and then call this same method again, we'll not have the icon since we've memoized the first response. 

This fix was my first idea to quickly see if this was really the issue. And it seems that yes. I'm not sure this is the best way to fix it thought since if we go then offline, we will request the icon and since the `onLine` has changed, we will not hit the cache and try to really fetch the icon. 

So maybe the best solution is to not memoize when we're get an error ?

(I think it can fix the issue, when we launch a mobile app when having no network and then having network but we still see the fallback icons)

### DIFF
--- a/packages/cozy-stack-client/src/getIconURL.js
+++ b/packages/cozy-stack-client/src/getIconURL.js
@@ -1,5 +1,4 @@
-import memoize from './memoize'
-
+import memoize, { ErrorReturned } from './memoize'
 const mimeTypes = {
   gif: 'image/gif',
   ico: 'image/vnd.microsoft.icon',
@@ -101,14 +100,9 @@ const _getIconURL = async (stackClient, opts) => {
   return URL.createObjectURL(icon)
 }
 
-/**
- * We need to catch the error and return a special object in
- * order to be able to delete the memoization if needed
- */
-
 const getIconURL = function() {
   return _getIconURL.apply(this, arguments).catch(e => {
-    return ''
+    return new ErrorReturned()
   })
 }
 
--- a/packages/cozy-stack-client/src/getIconURL.js
+++ b/packages/cozy-stack-client/src/getIconURL.js
@@ -114,7 +114,17 @@ export default memoize(getIconURL, {
   maxDuration: 300 * 1000,
   key: (stackClient, opts) => {
     const { type, slug, priority } = opts
-    return stackClient.uri + +':' + type + ':' + slug + ':' + priority
+    return (
+      stackClient.uri +
+      +':' +
+      type +
+      ':' +
+      slug +
+      ':' +
+      priority +
+      ':' +
+      window.navigator.onLine
+    )
   }
 })
 
--- a/packages/cozy-stack-client/src/getIconURL.js
+++ b/packages/cozy-stack-client/src/getIconURL.js
@@ -73,7 +73,6 @@ const _getIconURL = async (stackClient, opts) => {
       throw new Error(`Error while fetching icon ${resp.statusText}`)
     }
   })
-
   let icon = await resp.blob()
   let app
   if (!icon.type) {
@@ -99,13 +98,16 @@ const _getIconURL = async (stackClient, opts) => {
     }
     icon = new Blob([icon], { type: mimeTypes[ext] })
   }
-
   return URL.createObjectURL(icon)
 }
 
+/**
+ * We need to catch the error and return a special object in
+ * order to be able to delete the memoization if needed
+ */
+
 const getIconURL = function() {
   return _getIconURL.apply(this, arguments).catch(e => {
-    console.warn(e)
     return ''
   })
 }
@@ -114,17 +116,7 @@ export default memoize(getIconURL, {
   maxDuration: 300 * 1000,
   key: (stackClient, opts) => {
     const { type, slug, priority } = opts
-    return (
-      stackClient.uri +
-      +':' +
-      type +
-      ':' +
-      slug +
-      ':' +
-      priority +
-      ':' +
-      window.navigator.onLine
-    )
+    return stackClient.uri + +':' + type + ':' + slug + ':' + priority
   }
 })
 
--- a/packages/cozy-stack-client/src/getIconURL.spec.js
+++ b/packages/cozy-stack-client/src/getIconURL.spec.js
@@ -134,12 +134,39 @@ describe('get icon', () => {
       })
     )
   })
-  it('should not call create Object since it should be memoized', async () => {
-    responses['/konnectors/caissedepargne1/icon'] = {
+
+  it('should call the server the second time if the first time it failed', async () => {
+    responses['/registry/caissedepargne1/icon'] = {
       ok: true,
       blob: () => FakeBlob([svgData], { type: 'image/svg+xml' })
     }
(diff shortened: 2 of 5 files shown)


y-lohse:
> So maybe the best solution is to not memoize when we're get an error ?

That does sound like a better fix :)

edas:
> > So maybe the best solution is to not memoize when we're get an error ?
> 
> That does sound like a better fix :)

Agreed :+1:

packages/cozy-stack-client/src/getIconURL.js
```diff
@@ -114,7 +114,17 @@ export default memoize(getIconURL, {
   maxDuration: 300 * 1000,
   key: (stackClient, opts) => {
     const { type, slug, priority } = opts
-    return stackClient.uri + +':' + type + ':' + slug + ':' + priority
+    return (
```
edas: Je suis gêné par la concurrence d'accès. N'est-il pas possible que le statut online change entre le moment où la requête a été fait et le moment où tu calcules la clef ? D'autant que ce statut peut changer plusieurs fois au cours d'une requête, être "true" lors du lancement de la requête et du traitement de la réponse mais avoir généré une erreur parce que "false" pendant l'exécution réseau… ou inversement.

Crash--: Y a peu de chance, mais c'est bien pour ça que ce fix n'est pas bon ;). Un vrai fix devrait arriver

Crash--:
UPDATE: 
- En gros, maintenant si le résultat dans le memoize est une `promise`, et si celui-ci est en échec (aka === ''), on le retire du cache. Ainsi, plus de pb

packages/cozy-stack-client/src/memoize.js
```diff
@@ -29,6 +29,26 @@ const memoize = (fn, options) => {
...
+       * failed or resolved with a specific key (aka ''), let's remove
+       * the result from the cache since we don't want to
+       * memoize error
+       */
+
+      if (typeof result === 'object') {
+        if (typeof result.then === 'function') {
+          result
+            .then(v => {
+              if (v === '') {
```
ptbrowne: I think we should make this configurable, an empty string could be a valid value to memoize.

I wonder why in your case the called function returns '' as an error ?

ptbrowne: options.shouldCache = value => value !== ''

edas: What ares the errors that may return '' and nod false or undefined?

I agree with @ptbrowne and would prefer having the `if (v === '')` part to be a function given as option. Maybe `option.isError = v => (…)`

packages/cozy-stack-client/src/getIconURL.js
```diff
@@ -99,13 +98,16 @@ const _getIconURL = async (stackClient, opts) => {
...
+/**
+ * We need to catch the error and return a special object in
+ * order to be able to delete the memoization if needed
+ */
+
 const getIconURL = function() {
   return _getIconURL.apply(this, arguments).catch(e => {
-    console.warn(e)
     return ''
   })
```
ptbrowne: I htink we can reemove the catch, this way we can 

- remove the comment above
- remove the special case for '' in memoize

Crash--: In fact we can't without rewriting the method. 

If we don't catch here, we'll not hit the memoize function and will not hit the cache if needed :/. 

if you have any idea to avoid that, I'll be happy to remove that catch and this ugly return

packages/cozy-stack-client/src/memoize.js
Crash--: > I wonder why in your case the called function returns '' as an error ?

I think this is because the return is used directly for . I didn't want to change this current behavior. 

> What ares the errors that may return '' and nod false or undefined?

Typically, NO_INTERNET_CONNEXION, we currently return '' and not the error.

packages/cozy-stack-client/src/getIconURL.js
edas: If we really need to return a value here you could return an object of a specific class, or a Symbol.

Like : 
```
class ErrorReturned extends String { }

const getIconURL = function() {
  return _getIconURL.apply(this, arguments).catch(e => {
    return new ErrorReturned
  })
}
```

Then you could check if the return is an instance of ErrorReturned and not only equal to ''. It should not modify in any way the rest of your code

Crash--: And I need to return '' https://github.com/cozy/cozy-client/pull/515/files#diff-e22916779b00265fe0a667cc73319814R44  to not create a breaking change. 

Will give a shot later

nono:
Ping, what is the next step on this PR?

edas:
ping @Crash-- c'est toujours d'actualité pour toi ?

Crash--:
Oui j'ai pas eu le temps de m'y repencher mais c'est toujours d'actualité ça corrigerai un joli petit bug

Crash--:
I think I'm good here now :)

ptbrowne:
Can you add a test for the error case ? https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/memoize.spec.js

packages/cozy-stack-client/src/getIconURL.spec.js
```diff
@@ -1,20 +1,28 @@
...
 
 const FakeBlob = (data, options) => {
   return { data, ...options }
 }
 
+const resetcreateObjectURL = () => {
+  global.URL.createObjectURL = undefined
+  global.URL.createObjectURL = jest.fn(blob => {
+    return blob
+  })
```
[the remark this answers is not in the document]
Crash--: Good question it's been a long time since I wrote the code. I had a few issues with createObjectURL since this method doesn't exist in jsDom... But since I'm mocking it directly, I think I can remove it :)

packages/cozy-stack-client/src/memoize.spec.js
```diff
@@ -19,4 +20,23 @@ describe('memoize', () => {
...
+      })
+    }
+    const mcounter = memoize(counter, { key: () => 'A', maxDuration: 5 })
+    mcounter()
+    expect(c).toEqual(1)
+    mcounter()
+    //Let's wait the promise resolution
+    setTimeout(() => {
+      expect(c).toEqual(2)
+    }, 10)
```
Crash--: I don't know how to do differently without this timeout since we're waiting the promise resolution before deleting the cache...

ptbrowne: Since the expect is in a timeout, it is not called before the test finishes.

Better to await a sleep for this.

```

  const sleep = duration =>
    new Promise(resolve => setTimeout(resolve, duration))

    await sleep(12)
```

The second expect does not pass currently.

Crash--:
Thank you @ptbrowne !

PR #519, opened 2019-09-05
Perf: Query.jsx not changing objects passed to children

ptbrowne:
By not changing objects passed to children function at each render (the objects changed due to {...spreading}), React.memo can do its job.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -530,7 +530,14 @@ class CozyClient {
     return documents
   }
 
-  watchQuery(queryDefinition, options = {}) {
+  watchQuery(...args) {
+    console.warn(
+      'client.watchQuery is deprecated, please use client.makeObservableQuery.'
+    )
+    return this.makeObservableQuery(...args)
+  }
+
+  makeObservableQuery(queryDefinition, options = {}) {
     this.ensureStore()
     const queryId = options.as || this.generateId()
     this.ensureQueryExists(queryId, queryDefinition)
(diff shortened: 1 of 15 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -506,6 +506,9 @@ class CozyClient {
     this.ensureQueryExists(queryId, queryDefinition)
     try {
       const response = await this.requestQuery(queryDefinition)
+      if (!response) {
```
edas: Ca couvre quel cas ?
Mais sinon, côté style, ça revient à mettre plutôt : 

```
if (response) {
  this.dispatch(
    receiveQueryResult(queryId, response, {
       update
     })
  )
}
```

Et ainsi on évite le early return

ptbrowne: Perso je préfère les early return et j'évite d'indenter. Ca couvre un cas dans les tests, je vais modifier le test plutôt.

packages/cozy-client/src/Query.jsx
```diff
@@ -72,30 +114,18 @@ export default class Query extends Component {
...
   }
 
+  recomputeChildrenArgs() {
+    this.childrenArgs = computeChildrenArgs(this)
+  }
+
   render() {
     const { children } = this.props
-    return children(this.data, this.methods)
+    return children(this.childrenArgs[0], this.childrenArgs[1])
```
drazik: I wonder if using `return children(...this.childrenArgs)` can be good here. So if we add something to childrenArgs, we don't have to think about adding it here.

If `this.childrenArgs` doesn't change over time, this should always return the same object, even if this is a spread.

ptbrowne: Seen together, the overhead of using f(...args) seems too much considering that the number of arguments of the render prop will not change.

PR #522, opened 2019-09-09
feat: Fetch policy

ptbrowne:
`fetchPolicy` gives the ability to decide if the query should be
refetched at mount time. See query prop documentation for more
information ;).

Note: The `cache-only` value did not work and was never used, this is why I went with a breaking change here.


ptbrowne:
> Just a remark: the first commit is not really a fix. It's not a problem, but if it was merged alone, it would trigger a new useless patch release.

Correct, thank I had not thought of that,.

PR #530, opened 2019-10-02
Improvements

ptbrowne:
- Throw when refreshToken is used in a context where the token cannot be refreshed

I saw this error when developing for a service. The original error was "Need absolute URIs" since fetch("/") would fetch a URL that was not absolute (since it did not have the URL of the Cozy). Anyway, this method of refreshing the token by asking the server an HTML page cannot work in services, this is why I prefer a hard throw.

- Privatized methods from DocumentCollection as they were cluttering the docs
- Added updateAll and deleteAll in DocumentCollection. Originally from Document.

### DIFF
--- a/packages/cozy-stack-client/src/CozyStackClient.js
+++ b/packages/cozy-stack-client/src/CozyStackClient.js
@@ -124,6 +124,11 @@ class CozyStackClient {
       method: 'GET',
       credentials: 'include'
     }
+
+    if (!global.document) {
+      throw new Error('Not in a web context, cannot refresh token')
+    }
+
     const response = await fetch('/', options)
 
     if (!response.ok) {
@@ -169,7 +174,12 @@ class CozyStackClient {
       return await this.fetchJSONWithCurrentToken(method, path, body, options)
     } catch (e) {
       if (errors.EXPIRED_TOKEN.test(e.message)) {
-        const token = await this.refreshToken()
+        let token
+        try {
+          token = await this.refreshToken()
+        } catch (refreshError) {
+          throw e
+        }
         this.setToken(token)
         return await this.fetchJSONWithCurrentToken(method, path, body, options)
       } else {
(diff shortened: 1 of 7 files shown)


edas:
This means we have services that are long enough to expire their own token ?

packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -169,7 +174,12 @@ class CozyStackClient {
       return await this.fetchJSONWithCurrentToken(method, path, body, options)
     } catch (e) {
       if (errors.EXPIRED_TOKEN.test(e.message)) {
-        const token = await this.refreshToken()
+        let token
+        try {
+          token = await this.refreshToken()
+        } catch (refreshError) {
+          throw e
```
edas: why do we have a try/catch if we only throw the received error ? Isn't this the same as not having any try/catch ?

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -229,6 +251,49 @@ class DocumentCollection {
...
+        {
+          docs
+        }
+      )
+      return update
+    } catch (e) {
+      if (
+        e.reason &&
+        e.reason.reason &&
+        e.reason.reason == DATABASE_DOES_NOT_EXIST
```
edas:
```suggestion
        e.reason.reason === DATABASE_DOES_NOT_EXIST
```

packages/cozy-stack-client/src/CozyStackClient.js
ptbrowne: We do not throw the error, we throw the original error (token-expired error), not the tried-to-refresh-error. I think it is clearer for the user since she is not supposed to know that we tried to refresh the token.

ptbrowne:
> This means we have services that are long enough to expire their own token ?

No, this means that sometimes I develop slowly enough that the token expires ;)

packages/cozy-stack-client/src/CozyStackClient.js
edas: Oh, you're perfectly right!
I didn't see that.

PR #532, opened 2019-10-06
Improvements

ptbrowne:
Various little improvements. Commits are really short and
have descriptions, so everything should be pretty self-explanatory.

There are two breaking changes, but they should not really concern
anyone.

### DIFF
--- a/packages/cozy-client/src/store/documents.js
+++ b/packages/cozy-client/src/store/documents.js
@@ -90,14 +90,16 @@ export const getDocumentFromSlice = (state = {}, doctype, id) => {
   if (!state[doctype]) {
     if (process.env.NODE_ENV !== 'production') {
       console.warn(
-        `getDocumentFromSlice: ${doctype} is absent from the store documents`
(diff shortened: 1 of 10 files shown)


packages/cozy-client/src/store/documents.js
```diff
@@ -90,15 +90,17 @@ export const getDocumentFromSlice = (state = {}, doctype, id) => {
   if (!state[doctype]) {
     if (process.env.NODE_ENV !== 'production') {
       console.warn(
-        `getDocumentFromSlice: ${doctype} is absent from the store documents`
+        `getDocumentFromSlice: ${doctype} is absent from the store's documents. State is`
       )
+      console.warn(state)
```
y-lohse: Nit: if you pass multiple arguments to `console.warn`, they get printed each after the other. So `console.warn(message, state)` would work too :)

ptbrowne: Done

PR #534, opened 2019-10-10
feat(cozy-stack-client): Filecollection update file metadata

Crash--:
Use the new route to update file's metadata by creating a new
version of the file without having to upload it again.

See: https://github.com/cozy/cozy-stack/pull/2119/files

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -518,15 +518,23 @@ class FileCollection extends DocumentCollection {
   }
 
   /**
-   * async updateFileMetadata - Updates a file / folder's attributes except
-   * the metadata attribute. To update its metadata use `updateFileMetadataAttribute`
-   * see https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.files/#folder
+   * async updateAttributes - Updates a file / folder's attributes except
+   * the metadata attribute. If you want to update its metadata attribute,
+   * then use `updateFileMetadataAttribute` since `metadata` is a specific
+   * doctype.
+   *
+   * For instance, if you want to update the name of a file, you can pass
+   * attributes = { name: 'newName'}
+   *
+   * You can see the attributes for both Folder and File (as they share the
+   * same doctype they have a few in common) here :
+   * https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.files/#iocozyfiles
    *
    * @param  {string} id         File id
-   * @param  {object} attributes New file meta data
+   * @param  {object} attributes New file attributes
    * @returns {object}            Updated document
    */
-  async updateFileMetadata(id, attributes) {
+  async updateAttributes(id, attributes) {
     const resp = await this.stackClient.fetchJSON('PATCH', uri`/files/${id}`, {
       data: {
         type: 'io.cozy.files',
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -518,7 +518,9 @@ class FileCollection extends DocumentCollection {
   }
 
   /**
-   * async updateFileMetadata - Updates a file's metadata
+   * async updateFileMetadata - Updates a file / folder's attributes except
+   * the metadata attribute. To update its metadata use `updateFileMetadataAttribute`
+   * see https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.files/#folder
```
ptbrowne: `updateFileMetada` does not update `metadata` ? IMHO this is surprising and prone to errors. Do you think we could find names that are less prone to errors ? Also, I do not really understand the usecases here, do you think we could add to the docstrings when we want to do one but not the other ?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -561,6 +563,31 @@ class FileCollection extends DocumentCollection {
     }
   }
 
+  /**
+   *
+   * This method allows you to update the metadata attribute of a io.cozy.files
+   * It will result in a creation of a new version of the file
```
```suggestion
   * Creates of a new version of the file
```

packages/cozy-stack-client/src/FileCollection.js
ptbrowne: Also, the link is for "folder" but the method is for a file, is it wanted ?

Crash--: updateFileMetadata should be called `updateAttributes` but even with this name we have : 
```
io.cozy.files: 
data : {
   attributes: {
     id: 'cc',
    name:'ttt',
   metadata: {}
}
}
```

So the explicit name should be : `updateAttributeAndNotMetatada()`... 

`folder` or `file` is the "same thing". It's an io.cozy.files, but folder and files have differents attributes. Like a `file` as a `size` that `folder` doesn't have. 

what do you mean by `doctrings`?

Crash--: TBC, if you want to update the title attribute you use the older method. If you and to update something stored in the `metadata` attribute, then you have to use `updateFileMetadataAttribute`.

ptbrowne: > folder or file is the "same thing"

For the user of these methods, it is not, we should not leak this abstraction into our docs IMHO. As a user of the doc, clicking on a method of a file and landing on the folder is weird :/

> what do you mean by doctrings?

docstrings = doc des fonctions, c est un terme qui vient de python (où les docstrings sont supportés directement par le langage, tu peux acceder a la doc d'une fonction via `maFonction.__doc__`)

ptbrowne: I still think that the names are error prones, it would be easier if you could include usecases directly in the examples of the doc.

ptbrowne:
Would there be any problem to do the check if `attributes.metadata` is changed and call the right route ourselves ? This way we do not leak the abstraction below ?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -561,6 +578,34 @@ class FileCollection extends DocumentCollection {
     }
   }
 
+  /**
+   *
+   * Updates the metadata attribute of a io.cozy.files
+   * Creates a new version of the file without having
+   * to upload again the file's content
```
ptbrowne: Why are we mentioning versioning here ? Is it different from `updateAttributes`. If not, should we mention versioning there too ?

Crash--: I think https://docs.cozy.io/en/cozy-stack/files/#patch-filesfile-id-and-patch-filesmetadata doesn't create a new version of the file. @nono do you confirm?

nono: You are right

PR #540, opened 2019-10-15
feat: Fetch a single app from the registry

y-lohse:
When we need the registry status of just one app instead of all of them :)


packages/cozy-client/src/registry.js
```diff
@@ -95,6 +95,15 @@ class Registry {
   fetchAppsInMaintenance() {
     return this.client.stackClient.fetchJSON('GET', '/registry/maintenance')
   }
+
+  /**
+   * Fetch the status of a single app on the registry
+   *
+   * @return {Array<RegistryApp>}
```
Crash--: I think this is not an Array, right?

Crash--: Also add @param slug...

Issue #493, opened 2019-08-12
cozy-client: query() should hydrateDocuments?


Crash--:
ATM, when we use client.query(), we have to manually hydrate the documents https://github.com/cozy/cozy-drive/pull/1778/files/304a853bbf577d24649cdeda192c3b93ca660abe#diff-8bcb276dfd46a83c6e2e13ff28b8ad23R87 

But with an ObservableQuery, we hydrate the docs automatically. 

Is there any reason to not hydrate the documents when a perform a query() ?

ptbrowne:
I agree, I think hydrating in the `query` call is a good thing.

ptbrowne:
Would you be interested in taking a look @Crash-- ? I think first having an option `{ hydrate: true }` to `query` would be easy and useful.

paultranvan:
Ping @Crash-- ;) we recently got a situation of missing relationships because of `client.query` made in node. Adding `client.hydrateDocuments` works, but it's easy to miss / forget

PR #542, opened 2019-10-16
feat: GetQueryFromState can hydrate documents directly

ptbrowne:
/!\ : Should change base branch to master when docstrings is merged

Since selectors can be a way for an app to get documents from client, it
is useful to be able to select already hydrated documents. It also enables to remove hydratation logic from the Query.

This is useful to fix a bug in Banks that was introduced when we changed to use selectors on the store : since we previously received the documents from `queryConnect`, they were already hydrated. After changing to use selectors, they
were not hydrated.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -863,12 +863,18 @@ class CozyClient {
    * Get a query from the internal store.
    *
    * @param {String} id - Id of the query (set via Query.props.as)
+   * @param {Boolean} options.hydrated - Whether documents should be returned already hydrated (default: false)
    *
    * @return {QueryState} - Query state or null if it does not exist.
    */
-  getQueryFromState(id) {
+  getQueryFromState(id, options = {}) {
+    const hydrated = options.hydrated || false
     try {
-      return getQueryFromState(this.store.getState(), id)
+      const queryResults = getQueryFromState(this.store.getState(), id)
+      const data = hydrated
+        ? this.hydrateDocuments(queryResults.doctype, queryResults.data)
+        : queryResults.data
+      return { ...queryResults, data }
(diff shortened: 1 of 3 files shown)


ptbrowne:
This is related to https://github.com/cozy/cozy-client/issues/493

packages/cozy-client/src/CozyClient.js
```diff
@@ -863,12 +863,18 @@ class CozyClient {
    * Get a query from the internal store.
    *
    * @param {String} id - Id of the query (set via Query.props.as)
+   * @param {Boolean} options.hydrated - Whether documents should be returned already hydrated (default: false)
```
edas:
```suggestion
   * @param {Boolean} options.hydrated - Whether documents should be returned already hydrated (default: false)
      We can have already hydrated documents when we get them from the local store instead of Couchdb
```

ptbrowne: Mmh it seems a bit confusing to put this here since getQueryFromState is always taking data from the local store.

edas: Do you have a better way to explain when the swith is used and why ? I would welcome some context in comments

ptbrowne: "If you want to have relationships resolved when getting data from the store, you should toggle this flag" ?

Why wouldn't you want the data to be always hydrated ? Maybe you want to directly touch raw relationships ? The relationship resolving mechanic has a performance hit since it need to find all related documents in the store for each documents of the query.

Since this is a change a behavior from getQueryFromState I preferred to have an opt-in flag here but this could be rediscussed.

edas: Oh! it seems I understood wrongly what you did, sorry.

👌 for me

PR #543, opened 2019-10-16
chore: Tweak eslint

ptbrowne:
:warning: Change base branch to master when hydrate from state is merged.

- Unused variables are errors instead of warnings
- Added eslint plugin and ran autofix

The eslint plugin should help us write good docstrings
and can catch a lot of small mistakes that are tedious
to raise during a review.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,7 +3,7 @@
   "rules": {
     "no-console": 0,
     "no-unused-vars": [
-      "warn",
+      "error",
       {
         "args": "none",
         "ignoreRestSiblings": true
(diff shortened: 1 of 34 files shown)


ptbrowne:
> Cool. Wouldn't it be cool to have a lib entry in our eslint config that enables this plugin, so we have it on all our libs?

Yes I think it would make sense to have it in all our libs. But I am more circonspect to introduce it directly in our eslint-config-cozy-lib config since it generates a lot of warnings that can be considered noise. I do not want all our or libs to suddenly have a lot of warnings in eslint, this is why I prefer that first maintainers of each lib to activate the plugin themselves. It is pretty easy to install and activate as is IMHO :)

drazik:
Seems legit. Maybe we can add it when the majority of our libs manually added it and have a good jsdoc coverage.

PR #554, opened 2019-10-29
feat: Log out a web client

y-lohse:
The PR changes the behavior of `client.logout` to perform either an OAuth unregistration or a plain web client log out, depending on the situation.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -282,16 +282,24 @@ class CozyClient {
     this.emit('beforeLogout')
     this.isLogged = false
 
-    try {
-      // unregister client on stack
-      if (
-        this.stackClient.unregister &&
-        (!this.stackClient.isRegistered || this.stackClient.isRegistered())
-      ) {
-        await this.stackClient.unregister()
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -282,16 +282,24 @@ class CozyClient {
...
+        }
+      } catch (err) {
+        console.warn(`Impossible to unregister client on stack: ${err}`)
+      }
+    } else {
+      try {
+        await this.stackClient.fetch('DELETE', '/auth/login')
+      } catch (err) {
+        console.warn(`Impossible to log out: ${err}`)
       }
```
ptbrowne: I wonder if an alternative solution with unregister method on the "web" stack client would have not been better:

- Same method as oauth client, you do not need to know if it is an oauth or web stack client to call the right method

- Not need for introspection of `this.stackClient` by cozy-client

y-lohse: Ah, didn't think of that, but yeah that could have worked! It would also require a sort of mock "isRegistered" but still doable.

PR #555, opened 2019-10-31
feat: Added a simple settings collection

y-lohse:
This PR adds a simple Settings collection that can be used to query routes on [`/settings`](https://docs.cozy.io/en/cozy-stack/settings/#get-settingsinstance)., as discussed [here](https://github.com/cozy/cozy-home/pull/1369#discussion_r340784655).

### DIFF
--- a/packages/cozy-stack-client/src/SettingsCollection.js
+++ b/packages/cozy-stack-client/src/SettingsCollection.js
@@ -1,0 +1,31 @@
+import DocumentCollection from './DocumentCollection'
+
+export const SETTINGS_DOCTYPE = 'io.cozy.settings'
+
+/**
+ * Implements `DocumentCollection` API to interact with the /settings endpoint of the stack
+ */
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/SettingsCollection.js
```diff
@@ -0,0 +1,36 @@
...
+    try {
+      const resp = await this.stackClient.fetchJSON('GET', `/settings/${path}`)
+
+      return DocumentCollection.normalizeDoctypeJsonApi(SETTINGS_DOCTYPE)(
+        resp.data,
+        resp
+      )
+    } catch (error) {
+      return dontThrowNotFoundError(error)
+    }
```
Crash--: can you test the `dontThrow`  ? :)

y-lohse: Actually I took that from an existing file but it makes no sense in this context, we should pass down errors. I added a test for that.

PR #564, opened 2019-11-13
feat(cozy-stack-client): Download file version

Crash--:
Cozy-stack now supports file versioning. This PR adds the ability to download a specific revision of the file. We use the stack's route: https://github.com/cozy/cozy-stack/blob/94e62d16054c0bec33d0f283a6fc49ed1f6c8300/docs/files.md#post-filesdownloadsversionidfile_idversion_id 

Added a few test for the `download` method. 

Also add a small utility : getBeautifulSize() that expose a new utility to beautify the size of a file

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -1,7 +1,7 @@
 import mime from 'mime/lite'
 import has from 'lodash/has'
 import DocumentCollection, { normalizeDoc } from './DocumentCollection'
-import { uri, slugify, forceFileDownload } from './utils'
+import { uri, slugify, forceFileDownload, formatBytes } from './utils'
 import * as querystring from './querystring'
 const ROOT_DIR_ID = 'io.cozy.files.root-dir'
 const CONTENT_TYPE_OCTET_STREAM = 'application/octet-stream'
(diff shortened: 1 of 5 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -363,10 +371,40 @@ class FileCollection extends DocumentCollection {
...
-    const href = await this.getDownloadLinkById(file._id)
-    forceFileDownload(`${href}?Dl=1`, file.name)
+  /**
+   * Download a file or a specific version of the file
+   *
+   * Passing a filename to forceFileDownload is not enought
+   * for a few browsers since the stack's response header will
+   * not contain that name. Passing the filename to getDownloadLinkBy*
+   * will ask the stack to return this filename in its content-disposition
+   * header response
```
ptbrowne: It seems this comment is explaning the "how" of this function, I am not sure it belongs to the docstrings as it does not seem to be helpful to the user of this function. I would be in favor of moving the comment above the `if`.

Nit: I thought the start in `getDownloadLinkBy*` was a typo, I think it is clearer if written with `getDownloadLinkBy{Id,Revision}`

Nit: `enought` -> `enought`

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -363,10 +371,40 @@ class FileCollection extends DocumentCollection {
...
+   *
+   * @param {object} file io.cozy.files object
+   * @param {string} versionId Id of the io.cozy.files.version
+   * @param {string} filename The name you want for the downloaded file
+   *                            (by default the same as the file)
+   */
+  async download(file, versionId = null, filename = '') {
```
ptbrowne: Is there a reason to use an empty string instead of undefined ?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -363,10 +371,40 @@ class FileCollection extends DocumentCollection {
...
+    let href
+    const filenameToUse = filename !== '' ? filename : file.name
+    if (!versionId) {
+      href = await this.getDownloadLinkById(file._id, filenameToUse)
+    } else {
+      href = await this.getDownloadLinkByRevision(versionId, filenameToUse)
+    }
+    forceFileDownload(`${href}?Dl=1`, filenameToUse)
+  }
+  /**
```
ptbrowne: Nit: Can you add a line between the end of the method and the next docstring ? It makes the separation between functions clearer. I wonder why this is not done by eslint ?

packages/cozy-stack-client/src/FileCollection.spec.js
```diff
@@ -480,7 +480,47 @@ describe('FileCollection', () => {
...
+      client.fetchJSON.mockResolvedValue({
+        data: {}
+      })
+    })
+
+    afterEach(() => {
+      client.fetchJSON.mockClear()
+    })
+
+    it('should use default file object', async () => {
```
ptbrowne: I don't understand the description, maybe "it should work when not specifying a revision ?"

packages/cozy-stack-client/src/FileCollection.js
Crash--: Not needed anymore. Thanks

PR #565, opened 2019-11-18
feat(cozy-stack-client): Files get binary of a file

Crash--:
In a mobile env, we need to get the content of a file (or a version) in order to write it on the local FS. 

This is an equivalent cozy-client-js' `downloadById`

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -400,6 +400,18 @@ class FileCollection extends DocumentCollection {
   }
 
   /**
+   * Fetch the binary of a file or a specific version of a file
+   * Useful for instance when you can't download the file directly
+   * (via a content-disposition attachement header) and need to store
+   * it before doing an operation.
+   *
+   * @param {string} id Id of the io.cozy.files or io.cozy.files.version
+   *
+   */
+  async fetchFileContent(id) {
+    return this.stackClient.fetch('GET', `/files/download/${id}`)
+  }
+  /**
    * Get a beautified size for a given file
    * 1024B => 1KB
    * 102404500404B => 95.37 GB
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -399,6 +399,16 @@ class FileCollection extends DocumentCollection {
...
   }
 
+  /**
+   * Get the binary of a file or a specific version of a file
+   * used for Mobile native download
+   *
+   * @param {string} id Id of the io.cozy.files or io.cozy.files.version
+   *
+   */
+  async getFileContent(id) {
```
ptbrowne: I think it would be clearer to use "get" only when the data is local, here since we need to fetch it, I'd prefer fetchFileContent, it makes it clear to the caller that an asynchronous call is going to be done.

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -399,6 +399,16 @@ class FileCollection extends DocumentCollection {
     forceFileDownload(`${href}?Dl=1`, filenameToUse)
   }
 
+  /**
+   * Get the binary of a file or a specific version of a file
+   * used for Mobile native download
```
ptbrowne: I am not sure we should write in the doc for which usage a function is here. It is probable that at some point it is used in another place (or that the mobile native download does not use it anymore) and the comment will no tell the truth anymore.

ptbrowne:
Could you add a test ? LGTM otherwise.

packages/cozy-stack-client/src/FileCollection.js
Crash--: I like having an exemple of why I should use that method instead of, let's see for instance... `download()`. 

We have 3 methods to download a file from the stack, and I'm just figuring why (after few months). 

` Get the binary of a file or a specific version of a file.
Useful for instance when you can't download the file directly (via a content-disposition attachement header) and need to store it before doing an operation.`

Is it better?

PR #566, opened 2019-11-19
feat: Expose instance options

y-lohse:
Adds a method for CozyClient to expose the options injected by the stack into our web page templates.

### DIFF
--- a/docs/api/cozy-client.md
+++ b/docs/api/cozy-client.md
@@ -520,6 +520,8 @@ Responsible for
         * [.handleRevocationChange()](#CozyClient+handleRevocationChange)
         * [.handleTokenRefresh()](#CozyClient+handleTokenRefresh)
         * [.createClient()](#CozyClient+createClient)
+        * [.getInstanceOptions()](#CozyClient+getInstanceOptions) ⇒ <code>object</code>
+        * [.loadInstanceOptionsFromDOM([selector])](#CozyClient+loadInstanceOptionsFromDOM) ⇒ <code>void</code>
         * [.setData(data)](#CozyClient+setData)
     * _static_
         * [.fromOldClient()](#CozyClient.fromOldClient)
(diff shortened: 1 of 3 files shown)


docs/api/cozy-client.md
```diff
@@ -883,6 +885,23 @@ revocation and token refresh.
 If `oauth` options are passed, stackClient is an OAuthStackClient.
 
 **Kind**: instance method of [<code>CozyClient</code>](#CozyClient)  
+<a name="CozyClient+getInstanceOptions"></a>
+
+### cozyClient.getInstanceOptions() ⇒ <code>object</code>
+getInstanceOptions - Returns a set of instance options (such as domain or current app slug) that has been previously loaded
```
ptbrowne: When reading "a set", I am thinking of the Set object. What do you think of "Return instance options" ?

"that has been previously loaded": it is unclear from where and how, maybe add something like "Should be called after one of CozyClient::loadInstanceOptions method has been called" ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -1102,6 +1103,27 @@ class CozyClient {
...
+   * loadInstanceOptionsFromDOM - Loads the dataset injected by the Stack in web pages and exposes it through getInstanceOptions
+   *
+   * @param {string} [selector=[role=application]] A selector for the node that holds the dataset to load
+   *
+   * @returns {void}
+   */
+  loadInstanceOptionsFromDOM(selector = '[role=application]') {
+    const root = document.querySelector(selector)
+    this.instanceOptions = { ...root.dataset } // convert from DOMStringMap to plain object
+  }
```
ptbrowne: Do you think it could be done when instantiating the client ? This way we do not have to remember to do it ?

y-lohse: Yeah, I thought about that after I submitted the PR actually. I think that would be a good idea, i'll add it.

docs/api/cozy-client.md
ptbrowne: A throw if instance options is not defined could be helpful for the programmer ?

PR #567, opened 2019-11-20
feat: Add Instance Model to manage instance information

Crash--:
Add an Instance model to manage instance's informations

### DIFF
--- a/packages/cozy-client/src/models/instance.js
+++ b/packages/cozy-client/src/models/instance.js
@@ -1,0 +1,20 @@
+import get from 'lodash/get'
+const GB = 1000 * 1000 * 1000
+const PREMIUM_QUOTA = 50 * GB
+
+// If manager URL is present, then the instance is not self-hosted
+export const isSelfHosted = instanceInfo => {
+  return get(instanceInfo, 'context.data.attributes.manager_url') ? false : true
+}
+export const arePremiumLinksEnabled = instanceInfo => {
+  return get(instanceInfo, 'context.data.attributes.enable_premium_links')
+    ? true
+    : false
+}
+export const isFreemiumUser = diskUsage => {
+  const quota = get(diskUsage, 'data.attributes.quota', false)
+  return parseInt(quota) <= PREMIUM_QUOTA
+}
+export const getUuid = instanceInfo => {
+  return get(instanceInfo, 'instance.data.attributes.uuid')
+}
--- a/packages/cozy-client/src/models/index.js
+++ b/packages/cozy-client/src/models/index.js
@@ -1,1 +1,4 @@
 export { triggerResults } from './triggers'
+import * as instance from './instance'
+
+export { instance }
--- a/packages/cozy-client/src/models/instance.spec.js
+++ b/packages/cozy-client/src/models/instance.spec.js
@@ -1,0 +1,43 @@
+import { instance } from './'
+
+const noSelfHostedInstance = {
+  context: {
+    data: {
+      attributes: {
+        manager_url: 'https://manager.cozy.cc',
+        enable_premium_links: true
+      }
+    }
+  },
+  instance: {
+    data: {
+      attributes: {
+        uuid: '1234'
+      }
+    }
+  }
+}
+
+const selftHostedInstance = {
+  context: {
+    data: {
+      attributes: {}
+    }
+  }
+}
+describe('instance', () => {
+  it('should test if self hosted or not', () => {
+    expect(instance.isSelfHosted(noSelfHostedInstance)).toBe(false)
+    expect(instance.isSelfHosted(selftHostedInstance)).toBe(true)
+  })
+
+  it('should test if premium links are on', () => {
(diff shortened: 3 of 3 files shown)


packages/cozy-client/src/models/instance.js
```diff
@@ -0,0 +1,26 @@
...
+    return get(instanceInfo, 'context.data.attributes.manager_url')
+      ? false
+      : true
+  },
+  arePremiumLinksEnabled: instanceInfo => {
+    return get(instanceInfo, 'context.data.attributes.enable_premium_links')
+      ? true
+      : false
+  },
+  isFreemiumUser: diskUsage => {
```
ptbrowne: This is the only function here that does not take instanceInfo, is it expected ? Disk usage comes from another request ?

packages/cozy-client/src/models/instance.js
```diff
@@ -0,0 +1,26 @@
...
+    const quota = get(diskUsage, 'data.attributes.quota', false)
+    return parseInt(quota) < PREMIUM_QUOTA
+  },
+  getUuid: instanceInfo => {
+    return get(instanceInfo, 'instance.data.attributes.uuid')
+  }
+}
+
+export { instance }
```
ptbrowne: An alternative to exporting an object is to export the functions.

```
export const getUuid = instanceInfo => get(instanceInfo, 'instance.data.attributes.uuid')

export * as instance from './instance'
```

This would enable tree shaking.

or to export as default 

```
export instance from './instance'
```

I personally both of these alternative to be more predictible as user that the exporting of { instance }.

packages/cozy-client/src/models/instance.js
```diff
@@ -0,0 +1,26 @@
+import get from 'lodash/get'
+const GB = 1000 * 1000 * 1000
+const PREMIUM_QUOTA = 50 * GB
+
+const instance = {
+  //if managet url is here, then it's not selfhosted
```
```suggestion
  // If manager URL is here, then the instance is not self-hosted
```

packages/cozy-client/src/models/instance.js
Crash--: Thanks for asking because I didn't remember that I merged context and instance (currently in a hook something like : 
```jsx
const instancePromise = client
          .getStackClient()
          .fetchJSON('GET', '/settings/instance')

        const contextPromise = client
          .getStackClient()
          .fetchJSON('GET', '/settings/context')
        const promises = [instancePromise, contextPromise]

        const [instanceFetched, contextFetched] = await Promise.all(
          promises.map(p => p.catch(e => e))
        )
        if (!(instanceFetched instanceof Error)) {
          setInstance(instanceFetched)
        }
        if (!(contextFetched instanceof Error)) {
          setContext(contextFetched)
        }
```

In fact we have 3 requests : 
`/settings/context`
`/settings/instance`
`/settings/disk-usage`


`seflHosted` and `premiumLinks` are coming from `/context`
`uuid` from `/instance`
`diskusage` from `/disk-usage`


So I'm wondering if : 
- we should create an unique instanceInfo object with context / instance and disk-usage (we can imagine creating a method in SettingsCollection to fetch the three of them) 
- if we need to isolate each of this setting in its own model 

Personally, I prefer to have only one method to call to get the three and pass the fetched information to an unique InstanceModel. WDY?

ptbrowne: > Personally, I prefer to have only one method to call to get the three and pass the fetched information to an unique InstanceModel. WDY?

Yes I agree, I am a bit worried though because one call seems very benign but 3 network requests is pretty expensive. It is not expected. Could the stack regroup all this info in a single request ? @nono ?

Crash--: Yep having a `GET /settings` returning (disk-usage / contexte / instance / capabilities) can be a great enhancement!

packages/cozy-client/src/models/instance.js
Crash--: I like the simplicity of
```jsx

export const getUuid = instanceInfo => get(instanceInfo, 'instance.data.attributes.uuid')
export * as instance from './instance'
```

ptbrowne: The change in triggers.js would be breaking I guess, I think it is better if done in another PR.

packages/cozy-client/src/models/instance.js
nono: Well, we could, but with HTTP/2, I don't think there is a big difference between one call to fetch with the regrouped information, and 3 calls make in parallel.

@edas what is your opinion?

ptbrowne: Does HTTP2 reduces the concurrent rqust limit the browsers have ?

ptbrowne: It seems that the concurrent request limit is raised to 100 with HTTP2, see https://stackoverflow.com/questions/36835972/does-the-per-host-connection-limit-is-raised-with-http-2

edas: For requests sent in parallel (you do not wait the response from the first to start the second), having one or three requests is about the same. This was also true for HTTP 1.1 with pipelining.

PR #568, opened 2019-11-21
docs: Not call setStore too late

Crash--:
I was trying to use cozy-client in the bar by making a query. But since I was instancing cozy-bar before making `client.setStore` in Drive, I had this error and I didn't know what to do.

thanks to @ptbrowne, I now know that, when you call a query before setting a store, cozy-client creates by itself its store. So calling setStore later throws an exception. 

Add a line about the error

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -1023,7 +1023,10 @@ client.query(Q('io.cozy.bills'))`)
       throw new Error('Store is undefined')
     } else if (this.store && !force) {
       throw new Error(
-        'Client already has a store, it is forbidden to change store.'
+        `Client already has a store, it is forbidden to change store.
+setStore must be called before any query is executed. Try to 
+call setStore earlier in your code, preferably just after the 
+instantiation of the client.`
       )
     }
 


packages/cozy-client/src/CozyClient.js
```diff
@@ -985,7 +985,9 @@ class CozyClient {
       throw new Error('Store is undefined')
     } else if (this.store && !force) {
       throw new Error(
-        'Client already has a store, it is forbidden to change store.'
+        `Client already has a store, it is forbidden to change store.
+         You might have called setStore too late ie you might have a dispatching 
+         a query before calling setStore`
```
ptbrowne: The error will contain the spaces of indentation, it'd be better to remove the indentation spaces for the 2nd and 3rd lines. 

I think here the problem is described but the solution could be put in a more explicit manner.
What do you think of 
"setStore must be called before any query is executed. Try to call setStore earlier in your code, preferably just after the instantiation of the client".

nono:
Ping @Crash--

Crash--:
Thanks for the reminder @nono! Text edited @ptbrowne

PR #569, opened 2019-11-21
fix: SettingsCollection fix data response

Crash--:
ATM, if we use `SettingsCollection` as we don't return `data` even if we received a QUERY_RESULT our store is not updated with the latest status (and result). 

It can be considered as a BC but I think it is not currently usable since <Query or queryConnect will be "loading" for ever. I don't know... @y-lohse do you use it already?


ptbrowne:
breaking change right ?

ptbrowne:
We do not use it in banks, but I think tagging the change as breaking is good form.

Crash--:
done :)

PR #570, opened 2019-11-22
feat: Add shoulDisplayOffers method

Crash--:
We now have an standardized way of saying if an instance is concerned by our offers or not. 

Changed the `isFreemiumUser` to take instanceInfo like other methods (as suggested by @ptbrowne in a previous PR) + added a test in order to prevent BC. Since this method is not used yet, I didn't tag the commit as BC

Next PR soon, an helper to wrap those 3 queries (context / instance / diskusage)

### DIFF
--- a/packages/cozy-client/src/models/instance.js
+++ b/packages/cozy-client/src/models/instance.js
@@ -11,10 +11,27 @@ export const arePremiumLinksEnabled = instanceInfo => {
     ? true
     : false
 }
-export const isFreemiumUser = diskUsage => {
-  const quota = get(diskUsage, 'data.attributes.quota', false)
+export const isFreemiumUser = instanceInfo => {
+  const quota = get(instanceInfo, 'diskUsage.data.attributes.quota', false)
   return parseInt(quota) <= PREMIUM_QUOTA
 }
 export const getUuid = instanceInfo => {
   return get(instanceInfo, 'instance.data.attributes.uuid')
+}
+
+/**
+ * Returns whether an instance is concerned by our offers
+ *
+ * @param {object} data Object containing all the results from /settings/*
+ * @param {object} data.context Object returned by /settings/context
+ * @param {object} data.instance Object returned by /settings/instance
+ * @param {object} data.diskUsage Object returned by /settings/disk-usage
+ */
+export const shouldDisplayOffers = data => {
+  return (
+    !isSelfHosted(data) &&
+    arePremiumLinksEnabled(data) &&
+    getUuid(data) &&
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/models/instance.js
```diff
@@ -11,10 +11,27 @@ export const arePremiumLinksEnabled = instanceInfo => {
...
+
+/**
+ * Determines if an instance is concerned by our offers or not
+ *
+ * @param {object} data Object containing all the results from /settings/*
+ * @param {object} data.context Object returned by /settings/context
+ * @param {object} data.instance Object returned by /settings/instance
+ * @param {object} data.diskUsage Object returned by /settings/disk-usage
+ */
+export const shoulDisplayOffers = data => {
```
```suggestion
export const shouldDisplayOffers = data => {
```

packages/cozy-client/src/models/instance.spec.js
```diff
@@ -23,6 +30,13 @@ const selftHostedInstance = {
     data: {
       attributes: {}
     }
+  },
+  diskUsage: {
+    data: {
+      attributes: {
+        quota: '6000000000000'
```
ptbrowne: Is it normal that the quota is a string instead of a number here ?

packages/cozy-client/src/models/instance.js
```diff
@@ -11,10 +11,27 @@ export const arePremiumLinksEnabled = instanceInfo => {
...
+export const isFreemiumUser = instanceInfo => {
+  const quota = get(instanceInfo, 'diskUsage.data.attributes.quota', false)
   return parseInt(quota) <= PREMIUM_QUOTA
 }
 export const getUuid = instanceInfo => {
   return get(instanceInfo, 'instance.data.attributes.uuid')
 }
+
+/**
+ * Determines if an instance is concerned by our offers or not
```
```suggestion
 * Returns whether an instance is concerned by our offers
```

packages/cozy-client/src/models/instance.spec.js
Crash--: Normal idk, but if the question is: is it coherent with stack's response? yes : https://docs.cozy.io/en/cozy-stack/settings/#response

ptbrowne: OK for me, cc @nono, is it because encoding of large numbers in JSON should be done via strings ?

nono: I don't think we have numbers large enough to have precision issues (by a large margin), but I don't see any other explanations, and it was made a long time ago by people that are no longer in the team, so I assume it was the reason to use a string.

PR #572, opened 2019-11-25
feat: Instance model add buildPremiumLink

Crash--:
Create a link to the cozy's manager.

### DIFF
--- a/packages/cozy-client/src/models/instance.js
+++ b/packages/cozy-client/src/models/instance.js
@@ -35,3 +35,20 @@ export const shouldDisplayOffers = data => {
     isFreemiumUser(data)
   )
 }
+
+/**
+ * Returns the link to the Premium page on the Cozy's Manager
+ *
+ * @param {object} instanceInfo
+ */
+export const buildPremiumLink = instanceInfo => {
+  const managerUrl = get(
+    instanceInfo,
+    'context.data.attributes.manager_url',
+    false
+  )
+  const uuid = getUuid(instanceInfo)
+  if (managerUrl && uuid) {
+    return `${managerUrl}/cozy/instances/${uuid}/premium`
+  }
+}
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/models/instance.js
```diff
@@ -35,3 +35,20 @@ export const shouldDisplayOffers = data => {
     isFreemiumUser(data)
   )
 }
+
+/**
+ * Returns the link to the Premium page on the cozy's manager
```
```suggestion
 * Returns the link to the Premium page on the Cozy's Manager
```

I feel like when you're not familiar with facturation, the term "manager" is too broad and generic to be understood here. Is it the same thing as the nuagerie ? Could there be maybe a reference link to a doc explaning what the "manager" is in this context ?

Crash--: Cozy-Manager is the public face of the cloudery. 

For reference we also use it in `cozy-authenticate` (https://manager.cozycloud.cc/cozy/reminder / https://manager.cozycloud.cc/cozy/create for instance https://github.com/cozy/cozy-libs/blob/0cff9a0b41784b87ecede8052a4c6058d69001d2/packages/cozy-authentication/src/locales/en.json#L13 https://github.com/cozy/cozy-libs/blob/18c301428c3f3be31b7657a974d15cf29804e31c/packages/cozy-authentication/src/steps/ButtonLinkRegistration.jsx#L19) 


My understanding is that Cloudery / Nuagerie is more the APIs, Manager is the front (you can find it by looking for `backend-cozy` in gitlab). The manager can be different depending on where it runs (you'll see another backend project ;)). 

I don't know what to do to resolve the point you rose

ptbrowne: I think the main point is out of the scope of this PR, main point being that "manager" is not a good term since it is too gneric. I think in the scope of the PR that a link to a doc presenting the manager would be good, I cannot find such doc by searching "manager" in https://docs.cozy.io though :/ Ping @aeris @nono do you have a link to any such doc ?

nono: As I understand, manager can be the cloudery for the Cozy instances on our infra, but it can also be a web interface elsewhere where the user can pay for more space/features. For example, if we had an offer with Gandi, it could be the manager interface of Gandi. AFAIK, we don't have public documentation on that (and I'm not sure for internal documentation).

PR #573, opened 2019-11-28
feat: Expose cozy client's version

Crash--:
In order to be able to make a call to a new API or dealing with
new / old version, exposing the version makes it easier.

At the beginning I created a `getVersion` function in order to get the version, but I don't think it makes sense and brings complexecity since we'll have, in the app, to check if `getVersion` exists before calling it (or try/catch). Here, we juste have to do check if the result is not undefined 

To compare the versions, we just use something like this package : https://github.com/substack/semver-compare

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -1144,8 +1144,9 @@ class CozyClient {
     })
   }
 }
-
 CozyClient.fetchPolicies = fetchPolicies
+//COZY_CLIENT_VERSION_PACKAGE in replaced by babel. See babel config
+CozyClient.version = 'COZY_CLIENT_VERSION_PACKAGE'
 
 MicroEE.mixin(CozyClient)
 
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -36,6 +36,7 @@ import zip from 'lodash/zip'
 import forEach from 'lodash/forEach'
 import get from 'lodash/get'
 import MicroEE from 'microee'
+import jsonPackage from '../package.json'
```
ptbrowne: Nit: Downside here is that the whole package.json is included in the bundle where we only need the version.

I thought of using babel to do this, the search and replace plugin would work I think.

https://www.npmjs.com/package/babel-plugin-search-and-replace

Something along the line of

```
CozyClient.version = typeof PACKAGE_JSON_VERSION === 'undefined' ? '0.0.0' : PACKAGE_JSON_VERSION
```

with

babel.config.js

```
import package from './package.json

{
  ...,
  "plugins": [
    [
      "search-and-replace",
      {
        "rules": [
          {
            "search": "PACKAGE_JSON_VERSION",
            "replace": package.version
          }
        ]
      }
     ]
   ]
}
```

Crash--: Good Idea 💡

PR #574, opened 2019-11-28
docs: Added FilsCollection.get

y-lohse:
Added the docstring for a missing method.

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -53,6 +53,12 @@ class FileCollection extends DocumentCollection {
     this.specialDirectories = {}
   }
 
+  /**
+   * Fetches the file's data
+   *
+   * @param {string} id File id
+   * @returns {{data, included}} Information about the file or folder and it's descendents
+   */
   get(id) {
     return this.statById(id)
   }
--- a/docs/api/cozy-stack-client.md
+++ b/docs/api/cozy-stack-client.md
@@ -317,6 +317,7 @@ files associated to a specific document
 **Kind**: global class  
 
 * [FileCollection](#FileCollection)
+    * [.get(id)](#FileCollection+get) ⇒ <code>Object</code>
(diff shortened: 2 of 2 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -53,6 +53,9 @@ class FileCollection extends DocumentCollection {
     this.specialDirectories = {}
   }
 
+  /**
+   * Get a file by id
```
Crash--: What is the return? ;)

ptbrowne: I think we should try to use the verb fetch when doing something remote and get only when it's local.

y-lohse: I agree, that said the method is also called get.

ptbrowne: > I agree, that said the method is also called get.

So ? There are several options here : 

- add a new method with a new name, deprecate the other
- Create an issue not to forget
- Having the docstring clearer than the name can help

IMHO we should not stay with bad names.

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -53,6 +53,11 @@ class FileCollection extends DocumentCollection {
     this.specialDirectories = {}
   }
 
+  /**
+   * get - Fetches the file's data
```
```suggestion
   * Fetches the file's data
```

? 

I think the `get` is superfluous as it is picked up by jsdoc.

y-lohse: Ah yes, it's generated by my IDE and assumed it would be deduplicated, but it should indeed not be here. Pushing again.

packages/cozy-stack-client/src/FileCollection.js
y-lohse: In general I think fetch instead of get is a clearer, especially in cozy-stack-client, though I'm less sure about doing the same thing in cozy-client.
The goal of this PR was only to make the existing `get` method appear in the docs, not to refactor anything, so I used `fetch` in the docstring without changing the code.

PR #575, opened 2019-11-28
App model

y-lohse:
Migrates the existing `io.cozy.apps` doctype and adds a new method on it.

### DIFF
--- a/packages/cozy-client/src/models/applications.js
+++ b/packages/cozy-client/src/models/applications.js
@@ -38,7 +38,7 @@ export const isInstalled = (apps = [], wantedApp = {}) => {
 
 /**
  *
- * @param {object} app io.cozy.app object
+ * @param {object} app io.cozy.apps document
  * @returns {string} url to the app
  */
 export const getUrl = app => {
--- a/packages/cozy-client/src/models/applications.js
+++ b/packages/cozy-client/src/models/applications.js
@@ -1,0 +1,45 @@
+const STORE_SLUG = 'store'
+
+/**
+ * Return Store URL where an app/konnector can be installed / updated
+ *
+ * @param  {Array}  [appData=[]]   Apps data, as returned by endpoint /apps/ or
+ * /konnectors/
+ * @param  {object} [app={}] AppObject
+ * @returns {string}                URL as string
+ */
+export const getStoreInstallationURL = (appData = [], app = {}) => {
+  if (!app.slug) {
+    throw new Error('Expected app / konnector with the defined slug')
+  }
+
+  const storeApp = isInstalled(appData, { slug: STORE_SLUG })
+  if (!storeApp) return null
+
+  const storeUrl = storeApp.links && storeApp.links.related
+
+  if (!storeUrl) return null
+
+  return `${storeUrl}#/discover/${app.slug}/install`
+}
+
+/**
+ *
+ * @param {Array} apps Array of apps returned by /apps /konnectors
+ * @param {object} wantedApp io.cozy.app with at least a slug
+ * @returns {object} The io.cozy.app is installed or undefined if not
+ */
+export const isInstalled = (apps = [], wantedApp = {}) => {
+  return apps.find(
+    app => app.attributes && app.attributes.slug === wantedApp.slug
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/models/applications.js
```diff
@@ -0,0 +1,69 @@
+import get from 'lodash/get'
+const STORE_SLUG = 'store'
+
+/**
+ * Return Store URL where an app/konnector can be installed / updated
```
ptbrowne:
```suggestion
 * Returns the store URL to install/update an app/konnector
```

packages/cozy-client/src/models/applications.js
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @param {object} app io.cozy.app object
+ * @returns {string} url to the app
+ */
+export const getUrl = app => {
+  return app.links && app.links.related
+}
+
+/**
+ * getAppDisplayName
```
ptbrowne: Can you add a description ?

packages/cozy-client/src/models/applications.js
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @param {object} app Object io.cozy.apps or io.cozy.konnectors
+ * @param {string} lang Locale to use
+ *
+ * @returns {string} The name of the app suited for UI display
```
ptbrowne:
```suggestion
 * @returns {string} Name of the app suitable for display
```

packages/cozy-client/src/models/applications.js
```diff
@@ -0,0 +1,69 @@
...
+ * @returns {string} url to the app
+ */
+export const getUrl = app => {
+  return app.links && app.links.related
+}
+
+/**
+ * getAppDisplayName
+ *
+ * @param {object} app Object io.cozy.apps or io.cozy.konnectors
```
ptbrowne:
```suggestion
 * @param {object} app io.cozy.apps or io.cozy.konnectors document
```

packages/cozy-client/src/models/applications.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+export const isInstalled = (apps = [], wantedApp = {}) => {
+  return apps.find(
+    app => app.attributes && app.attributes.slug === wantedApp.slug
+  )
+}
+
+/**
+ *
+ * @param {object} app io.cozy.app object
```
ptbrowne:
```suggestion
 * @param {object} app io.cozy.apps document
```

packages/cozy-client/src/models/applications.js
```diff
@@ -0,0 +1,69 @@
...
+  return `${storeUrl}#/discover/${app.slug}/install`
+}
+
+/**
+ *
+ * @param {Array} apps Array of apps returned by /apps /konnectors
+ * @param {object} wantedApp io.cozy.app with at least a slug
+ * @returns {object} The io.cozy.app is installed or undefined if not
+ */
+export const isInstalled = (apps = [], wantedApp = {}) => {
```
ptbrowne: The signature looks weird and the implementation looks like it is very easy to make a mistake a have a O(n2) behavior (since to check installation status we have to iterate on all the list. Keying by slug would solve this problem IMHO).

packages/cozy-client/src/models/applications.js
Crash--: I'm wondering who write this comment first... 🤔 ^^"

packages/cozy-client/src/models/applications.js
Crash--: `Keying by slug?` what do you mean by that? How can I access to apps....slug===slug? 

Do you suggest that instead of having a simple Array apps[0], apps[1], we had something `keyed` like apps[slug]?

ptbrowne: I think that if apps were instead an appsBySlug object, it would be harder to shoot yourself in the foot.

Crash--: https://github.com/cozy/cozy-libs/issues/875

PR #576, opened 2019-11-29
feat: Add helper functions

ptbrowne:
Extracted helpers from Banks

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -991,6 +991,13 @@ class CozyClient {
     }
   }
 
+  /**
+   * Returns whether the client has been revoked on the server
+   */
+  async checkForRevocation() {
+    return this.stackClient.checkForRevocation()
+  }
+
   /** Sets public attribute and emits event related to revocation */
   handleRevocationChange(state) {
     if (state) {
--- a/packages/cozy-client/src/utils.js
+++ b/packages/cozy-client/src/utils.js
@@ -22,3 +22,22 @@ const cancelable = promise => {
 }
 
 export { cancelable }
+
+/**
+ * Returns whether the result of a query (given via queryConnect or Query)
+ * is loading.
+ */
+export const isQueryLoading = col => {
+  if (!col) {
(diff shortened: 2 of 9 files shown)


packages/cozy-client/src/utils.js
```diff
@@ -22,3 +22,38 @@ const cancelable = promise => {
...
 
 export { cancelable }
+
+/**
+ * Returns whether the result of a query (given via queryConnect or Query)
+ * is loading.
+ */
+export const isQueryLoading = col => {
+  if (!col) {
+    console.warn('isCollectionLoading called on falsy value.') // eslint-disable-line no-console
```
Crash--:
```suggestion
    console.warn('isQueryLoading called on falsy value.') // eslint-disable-line no-console
```

ptbrowne: Done

packages/cozy-client/src/CozyClient.js
```diff
@@ -991,6 +991,22 @@ class CozyClient {
...
+      await this.stackClient.fetchInformation()
+      return false
+    } catch (err) {
+      if (err.message && err.message.indexOf('Client not found') > -1) {
+        return true
+      } else {
+        return false
+      }
+    }
+  }
```
Crash--: What is the difference between this one and the one already in stackClient ? https://github.com/cozy/cozy-client/blob/fa0c1d6bea0c9e6d0f5a8c7cf0d74102982a88ab/packages/cozy-stack-client/src/CozyStackClient.js#L105 

Can we at least use this const https://github.com/cozy/cozy-client/blob/f67e6831e871dc0cc2d596c5a8387db1323cbf11/packages/cozy-stack-client/src/errors.js#L2 ?

ptbrowne: The one in CozyStackClient was a private helper used when running into a fetch error. It was not meant to used from the exterior.

Since the const needed to be use, I figured that it was better to have the method directly on CozyStackClient. So I removed the helper checkForRevocation and changed it into the checkForRevocation that I had put in CozyClient at first.

Crash--: 👍

PR #577, opened 2019-11-29
feat: Export applications model

y-lohse:
Forgot to export the new model!


Crash--:
Instead of importing the model from ./ApplicationModel we should import it from './' in our tests, no? Like that, we are sure to use it the same way that in an app 

https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/applications.spec.js#L1

PR #583, opened 2019-12-05
fix: Don't change my headers

Crash--:
We fount out a strange behavior in drive when our token was expired and we called a `find` with indexes and sort in the same time. 

The call to refresh token was done and we got a new token, but the recall of this find failed with a nice error  

Capture d’écran 2019-12-04 à 12 21 27

After looking in my request panel, I find out that my request had a strange payload... 

Capture d’écran 2019-12-05 à 08 44 27

With the help of @y-lohse we fixed the issue. 

It's pretty simple but deceitful in the same time. The fact is that we are adding a Content-Type header (https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/CozyStackClient.js#L214) to `options`. But since options is an object, the scope of this change is not the one we want. By adding this header, in the second call, we do not stringify our body (resulting in the `Object object`).

To fix it, we simply cloned the option before adding parameters. 

I added a test to be sure that the second call of `fetch` is done with the same arguments than the first one

### DIFF
--- a/packages/cozy-stack-client/src/CozyStackClient.js
+++ b/packages/cozy-stack-client/src/CozyStackClient.js
@@ -1,3 +1,4 @@
+import cloneDeep from 'lodash/cloneDeep'
 import AppCollection, { APPS_DOCTYPE } from './AppCollection'
 import AppToken from './AppToken'
 import DocumentCollection from './DocumentCollection'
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -205,8 +205,10 @@ class CozyStackClient {
   }
 
   async fetchJSONWithCurrentToken(method, path, body, options = {}) {
-    const headers = (options.headers = options.headers || {})
-
+    //Since we modify the object later by adding in some case a
+    //content-type, let's clone this object to scope the modification
+    const clonedOptions = { ...options }
+    const headers = (clonedOptions.headers = clonedOptions.headers || {})
```
ptbrowne: Nit: If there are nested options, we can still have a bug since the rest of the method could change stuff (thinking "oh the var is called 'cloned', there is no problem to change any value inside"). It would be even safer to deepClone the object.

Crash--: Nice catch! Changed to use lodash.cloneDeep()

ptbrowne:
🎉 Thanks ! LGTM. Takeaway : never never change the options without cloning them.

PR #585, opened 2019-12-09
feat: Instance add method to check if already subscribed

Crash--:
We need to check if an instance already subscribed to an offer

### DIFF
--- a/packages/cozy-client/src/models/instance.js
+++ b/packages/cozy-client/src/models/instance.js
@@ -35,6 +35,23 @@ export const shouldDisplayOffers = data => {
     isFreemiumUser(data)
   )
 }
+/**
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/models/instance.js
```diff
@@ -35,6 +35,23 @@ export const shouldDisplayOffers = data => {
     isFreemiumUser(data)
   )
 }
+/**
+ * Returns if an instance has subsribre to one of our offers
```
y-lohse:
```suggestion
 * Returns if an instance has subscribed to one of our offers
```

Crash--: thanks!

PR #588, opened 2019-12-09
feat: Fix regeneratorRuntime

edas:
This commit is only a jsDoc change. Its purpose is to trigger a
new release of cozy-stack-client. The real feature is on #584


edas:
> Can you rename it to doc?

Am I wrong to say that a `doc` commit won't trigger a new release ?

Issue #72, opened 2018-08-03
How to test components with `Query`


ptbrowne:
When testing components which have a `Query` (at the root of the component or below in the tree), the query should not be executed by the client, and fixtures should be returned from the store. This removes the side-effects so that we do not need a running cozy for the tests.

Apollo solves this problem with a `MockedProvider`. This MockProvider is passed data that should be returned from the GraphQL endpoints and the `fetch`es ared mocked to return this data. I think it would be a good idea to add this functionality to cozy-client's testUtils.

http://blog.dideric.is/2018/03/18/Testing-apollo-containers/

```
const ComponentToTest = () => {
  return <Query>( data => <div>{ data.title }</div> )</Query>
}
```

Crash--:
I had the same issue, ie I wanted to test something like : 
```jsx 
//ContactsList.jsx

<Query query={query} name="getContacts">
        {({ data: contacts, fetchStatus }) => {
         contacts.map(c => <Contact contact={c} />)
        }}
</Query>
```

As we can't easily mock our `Query` component ATM, I had to create a new component taking a `contacts` props.

I like the idea of having a MockedProvider : 
```jsx
const queryMocksResult = [{
   name: 'getContacts',
    result: {
      data: [
        { id: '1', name: 'contact 1'},
      ],
     fetchStatus: 'loaded'
    }
  }
]

it('renders without error', () => {
  mount(
    <MockedProvider mocks={queryMocksResult} >
       <ContactLists /> 
    </MockedProvider>,
  );
});
```

(for reference here is the implementation of MockedProvider : https://github.com/apollographql/react-apollo/blob/master/src/test-utils.tsx)

cedricmessiant:
For inspiration: https://www.apollographql.com/docs/react/recipes/testing.html#MockedProvider

I already used it (directly with the client via a `mockedNetworkInterface` and it was ok.

Debug was a bit painful because it doesn't provide full information about what is wrong in your query / params.

Also, it would be nice to be able to use mocked data when we don't use the provider (something like `MockedCozyClient` for example). It may be useful for non react code that may use `cozy-client` (connectors, etc).

ptbrowne:
We should use the mock client when testing

```
import { createMockClient } from 'cozy-client/dist/mock'
```

Maybe mock could be renamed `testing`, it would make it more obvious that this file is used for testing.

PR #589, opened 2019-12-11
feat: Add mockClient creation for use in tests

ptbrowne:
Move mock client from banks to here. Client internal store can be
prefilled with data from fixtures and query is mocked so that calls
to the server use fixtures.

Related: https://github.com/cozy/cozy-client/issues/72


babel.config.js
```diff
@@ -1,5 +1,7 @@
 const packageJson = require('./packages/cozy-client/package.json')
 
+console.log('oto')
```
Crash--: I knew it! https://en.wikipedia.org/wiki/Ordo_Templi_Orientis

ptbrowne: You got me !

ptbrowne:
Good idea: https://github.com/cozy/cozy-client/issues/590, I assigned myself to it.

PR #594, opened 2019-12-16
feat: Add Note model

Crash--:
Since Note is coming, we need an helper to centralize few utils.

Here, I just added a method to generate the link to a note from an other app.

### DIFF
--- a/packages/cozy-client/src/models/note.js
+++ b/packages/cozy-client/src/models/note.js
@@ -1,0 +1,8 @@
+/**
+ *
+ * @param {string} notesAppUrl URL to the Notes App (https://notes.foo.mycozy.cloud)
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/models/note.js
```diff
@@ -0,0 +1,8 @@
+/**
+ *
+ * @param {string} notesAppUrl URL to the Notes App (https://notes.foo.mycozy.cloud)
+ * @param {File} file io.cozy.files object
+ */
+export const generateUrlForOneNote = (notesAppUrl, file) => {
```
y-lohse: I guess it could be `generateUrlForNote ` no?

Crash--: You absolutely right!

PR #597, opened 2019-12-18
feat: Destroy all removes the _type from the document

ptbrowne:
Otherwise we have an error from the stack, since _type is a reserved
attribute


ptbrowne:
I guess that the stack does that automatically on regular updates.

https://github.com/cozy/cozy-stack/blob/3b1e500493177694ba13d4e5920c844ff09b23aa/pkg/couchdb/couchdb.go#L183

cc nono ?

PR #598, opened 2019-12-18
feat: Share a link with special permissions

edas:
This will be useful for cozy-notes. We want links to accept PATCH and PUT so the recipient can edit the note.

### DIFF
--- a/packages/cozy-stack-client/src/PermissionCollection.js
+++ b/packages/cozy-stack-client/src/PermissionCollection.js
@@ -104,7 +104,15 @@ class PermissionCollection extends DocumentCollection {
     return { ...resp, data: resp.data.map(a => ({ _id: a.id, ...a })) }
   }
 
-  async createSharingLink(document) {
+  /**
+   * Create a share link
+   *
+   * @param {{_id, _type}} document - cozy document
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -104,15 +104,23 @@ class PermissionCollection extends DocumentCollection {
...
+    const { verbs } = options
     const resp = await this.stackClient.fetchJSON(
       'POST',
       `/permissions?codes=email`,
       {
         data: {
           type: 'io.cozy.permissions',
           attributes: {
-            permissions: getPermissionsFor(document, true)
+            permissions: getPermissionsFor(document, verbs ? { verbs } : true)
```
y-lohse: I don't think this works as intended, if we pass `verbs` we will call `getPermissionsFor (document, options)` bu the signature is `getPermissionsFor (document, publicLink, options)`

edas: Stupid me, should be better now, and I added tests

edas:
> Not sure this works as intended, a test to be safe would be good :)

done

PR #600, opened 2019-12-18
feat: Create links for multiple documents

edas:
This will allow links to embed couchdb along
with files, and is needed for cozy-notes

### DIFF
--- a/packages/cozy-stack-client/src/PermissionCollection.js
+++ b/packages/cozy-stack-client/src/PermissionCollection.js
@@ -105,7 +105,7 @@ class PermissionCollection extends DocumentCollection {
   }
 
   /**
-   * Create a new set of permission on the stack
+   * Create a new set of permissions on the stack
    * @param {object} permissions - list of permissions
    * @see https://docs.cozy.io/en/cozy-stack/permissions/#post-permissions
    * @private
--- a/packages/cozy-stack-client/src/PermissionCollection.js
+++ b/packages/cozy-stack-client/src/PermissionCollection.js
@@ -105,14 +105,12 @@ class PermissionCollection extends DocumentCollection {
   }
 
   /**
-   * Create a share link
-   *
-   * @param {{_id, _type}} document - cozy document
-   * @param {object} options - options
-   * @param {string[]} options.verbs - explicit permissions to use
+   * Create a new set of permission on the stack
+   * @param {object} permissions - list of permissions
+   * @see https://docs.cozy.io/en/cozy-stack/permissions/#post-permissions
+   * @private
    */
-  async createSharingLink(document, options = {}) {
-    const { verbs } = options
+  async createPermissions(permissions) {
     const resp = await this.stackClient.fetchJSON(
       'POST',
       `/permissions?codes=email`,
@@ -120,16 +118,41 @@ class PermissionCollection extends DocumentCollection {
         data: {
           type: 'io.cozy.permissions',
           attributes: {
-            permissions: getPermissionsFor(
-              document,
-              true,
-              verbs ? { verbs } : {}
-            )
+            permissions: permissions
           }
         }
       }
     )
     return { data: normalizePermission(resp.data) }
+  }
+
+  /**
+   * Create a share link
+   *
+   * @param {{_id, _type}} document - cozy document
+   * @param {object} options - options
+   * @param {string[]} options.verbs - explicit permissions to use
+   */
+  async createSharingLink(document, options = {}) {
+    const permissions = getPermissionsFor(document, true, options)
+    return this.createPermissions(permissions)
+  }
+
+  /**
+   * Create a share link from multiple documents
+   *
+   * @param {object} documents - key/value
+   * where each value is either a document {_id, _type}
+   * or an array with a document and an option object
+   */
+  async createCompositeSharingLink(documents) {
+    const permissions = Object.keys(documents).map(key => {
+      const item = documents[key]
+      const document = Array.isArray(item) ? item[0] : item
+      const options = Array.isArray(item) ? item[1] : {}
+      return getPermissionsFor(document, true, { name: key, ...options })
+    })
+    return this.createPermissions(Object.assign({}, ...permissions))
   }
 
   async revokeSharingLink(document) {
@@ -156,6 +179,19 @@ class PermissionCollection extends DocumentCollection {
 }
 
 /**
+ * Should we also give access to referenced_by files ?
+ *
+ * @param {{_type}} document - document to share
+ * @returns {boolean}
+ */
+export const shouldReferencedFiledBeIncluded = document => {
+  if (isFile(document)) return false
+  if (document._type == 'io.cozy.notes') return false
+  if (document._type == 'io.cozy.notes.events') return false
+  return true
+}
+
+/**
  * Build a permission set
  *
  * @param {{_id, _type}} document - cozy document
@@ -171,27 +207,29 @@ export const getPermissionsFor = (
 ) => {
   const { _id, _type } = document
   const verbs = options.verbs ? options.verbs : publicLink ? ['GET'] : ['ALL']
+  const name = options.name || 'files'
+  const collectionName = (options.name || '') + 'collection'
   // TODO: this works for albums, but it needs to be generalized and integrated
   // with cozy-client ; some sort of doctype "schema" will be needed here
-  return isFile(document)
+  return shouldReferencedFiledBeIncluded(document)
     ? {
-        files: {
-          type: 'io.cozy.files',
-          verbs,
-          values: [_id]
-        }
-      }
-    : {
-        collection: {
+        [collectionName]: {
           type: _type,
           verbs,
           values: [_id]
         },
-        files: {
+        [name]: {
           type: 'io.cozy.files',
           verbs,
           values: [`${_type}/${_id}`],
           selector: 'referenced_by'
+        }
+      }
+    : {
+        [name]: {
+          type: _type,
+          verbs,
+          values: [_id]
         }
       }
 }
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -105,33 +105,49 @@ class PermissionCollection extends DocumentCollection {
...
-   * Create a share link
-   *
-   * @param {{_id, _type}} document - cozy document
-   * @param {object} options - options
-   * @param {string[]} options.verbs - explicit permissions to use
+   * @private
    */
-  async createSharingLink(document, options = {}) {
-    const { verbs } = options
+  async createPermissions(permissions) {
```
Crash--: can we have a jsdoc for this `permissions` ?

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -155,6 +171,18 @@ class PermissionCollection extends DocumentCollection {
...
 
+/**
+ * Is the document a collection document?
+ * @param {{_type}} document
+ * @returns {boolean}
+ */
+export const isCollection = document => {
+  if (isFile(document)) return false
+  if (document._type == 'io.cozy.notes') return false
+  if (document._type == 'io.cozy.notes.events') return false
```
Crash--: should we create an `isNote(document)` testing if _type starts with `io.cozy.notes` ? 

Also it's a bit disturbing that sometime a note is a `io.cozy.file` and sometimes (in the realtime use case) we work with an `io.cozy.notes`

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -105,33 +105,49 @@ class PermissionCollection extends DocumentCollection {
...
+   */
+  async createSharingLink(document, options = {}) {
+    const permissions = getPermissionsFor(document, true, options)
+    return this.createPermissions(permissions)
+  }
+
+  /**
+   * Create a share link
+   */
+  async createCompositeSharingLink(documents) {
```
Crash--: jsdoc ;)

packages/cozy-stack-client/src/PermissionCollection.js
edas: We always work with a io.cozy.files.
The couchdb io.cozy.notes exists but we shouldn't have to use it.

The addition here is more because I was on the function and it was free to fill some examples.

packages/cozy-stack-client/src/PermissionCollection.js
edas: done

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -105,33 +105,52 @@ class PermissionCollection extends DocumentCollection {
   }
 
   /**
-   * Create a share link
-   *
-   * @param {{_id, _type}} document - cozy document
-   * @param {object} options - options
-   * @param {string[]} options.verbs - explicit permissions to use
+   * Create a new set of permission on the stack
```
ptbrowne:
```suggestion
   * Create a new set of permissions on the stack
```

packages/cozy-stack-client/src/PermissionCollection.js
edas: done

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -105,33 +105,56 @@ class PermissionCollection extends DocumentCollection {
...
+   */
+  async createCompositeSharingLink(documents) {
+    const permissions = Object.keys(documents).map(key => {
+      const item = documents[key]
+      const document = Array.isArray(item) ? item[0] : item
+      const options = Array.isArray(item) ? item[1] : {}
+      return getPermissionsFor(document, true, { name: key, ...options })
+    })
+    return this.createPermissions(Object.assign({}, ...permissions))
+  }
```
ptbrowne: If I understand well, this is the responsibility of the callers to know that they should use a different method depending on whether they have several documents or only one document ?

If so, I think it would be ergonomic if `createSharingLink` could also accept an array of documents and call createCompositeSharingLink by itself.

- Only one method to know for the caller
- External API surface reduced

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -155,6 +178,18 @@ class PermissionCollection extends DocumentCollection {
...
+ * Is the document a collection document?
+ * @param {{_type}} document
+ * @returns {boolean}
+ */
+export const isCollection = document => {
+  if (isFile(document)) return false
+  if (document._type == 'io.cozy.notes') return false
+  if (document._type == 'io.cozy.notes.events') return false
+  return true
+}
```
ptbrowne: I don't understand what is a collection in this context ? How should one use this method ?

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -105,33 +105,56 @@ class PermissionCollection extends DocumentCollection {
...
+   */
+  async createSharingLink(document, options = {}) {
+    const permissions = getPermissionsFor(document, true, options)
+    return this.createPermissions(permissions)
+  }
+
+  /**
+   * Create a share link from multiple documents
+   *
+   * @param {} documents - key/value
```
ptbrowne:
```suggestion
   * @param {Array} documents - key/value
```

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -171,23 +206,25 @@ export const getPermissionsFor = (
 ) => {
   const { _id, _type } = document
   const verbs = options.verbs ? options.verbs : publicLink ? ['GET'] : ['ALL']
+  const name = options.name || 'files'
+  const collectionName = (options.name || '') + 'collection'
   // TODO: this works for albums, but it needs to be generalized and integrated
   // with cozy-client ; some sort of doctype "schema" will be needed here
-  return isFile(document)
+  return !isCollection(document)
```
ptbrowne: Nit: I think it is clearer when the condition is positive

```patch
- isCollection(document) ? otherStuff : stuff
+ isCollection(document) ? stuff : otherStuff
```

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -155,6 +178,18 @@ class PermissionCollection extends DocumentCollection {
   }
 }
 
+/**
+ * Is the document a collection document?
```
ptbrowne: Nit: Can the description be clearer ? Explaining the context and not the what ?

"Is the document a collection document? " 

is not much clearer than

` export const isCollection = document => { `

I think it is because the stack needs to have "collection" in the name of the
attributes in the permission set ? Maybe we could rename this to reflect this
fact ? `isCollection` -> `needsToHaveCollectionNameInPermissionSet` ?

packages/cozy-stack-client/src/PermissionCollection.js
edas: thx

packages/cozy-stack-client/src/PermissionCollection.spec.js
```diff
@@ -180,4 +259,27 @@ describe('getPermissionsFor', () => {
...
+
+  it('io.cozy.notes.events is not a collection', () => {
+    const document = { _type: 'io.cozy.notes.events', _id: '1234' }
+    expect(isCollection(document)).toBe(false)
+  })
+
+  it('io.cozy.photos.albums is a collection', () => {
+    const document = { _type: 'io.cozy.photos.albums', _id: '1234' }
+    expect(isCollection(document)).toBe(true)
+  })
```
ptbrowne: I am wondering if isCollection should be tested explicitly ? AMHA it should be tested as part of the `createSharingLink` tests. Testing internal methods makes test brittle as we cannot easily rename methods (breaking the tests).

packages/cozy-stack-client/src/PermissionCollection.js
edas: 👍

packages/cozy-stack-client/src/PermissionCollection.spec.js
edas: I usually like to test methods that have a real functional meaning. If we delete it later, we can easily delete the associated tests, but this garantee the functional.

This is less true for purely technical helpers.

isCollection (renamed now) is in the former case

packages/cozy-stack-client/src/PermissionCollection.js
edas: :+1:

packages/cozy-stack-client/src/PermissionCollection.js
Crash--: I agree with you. I think having only one exposed method to create link is better.

edas: I wrote another function because the prototype was really different.

It's not just an array of document, otherwise I would agree with you.

It's a key/value of documents with their own options, which differ significantly from the first function. I don't feel confortable with having the same name with very different parameters.

`createSharingLink(document)`
`createSharingLink(document, { verbs: ['ALL'] })`
`createSharingLink({
   name1: [ document, { verbs: ['ALL'] } ],
   name2: [ document, { verbs: ['ALL'] } ],
})`

ptbrowne: s/prototype/signature I guess.

AMHA in this case we should deprecate the first API that proves itself not very flexible.

I would be comfortable having different parameters. But I understand your reason. Your call ;)

nono:
What is the status of this PR?

Crash--:
@edas I close this issue since we don't need anymore to create multiple permissions at once. For a note, we just the permission on the file.

PR #602, opened 2019-12-23
feat: File normalization function

edas:
As requested in https://github.com/cozy/cozy-notes/pull/41#discussion_r360844792

### DIFF
--- a/packages/cozy-client/src/models/file.js
+++ b/packages/cozy-client/src/models/file.js
@@ -34,23 +34,30 @@ export const isNote = file => {
 /**
  * Normalizes an object representing a io.cozy.files object
  *
- * Ensures existence of `_id` and `_type` and defines a `path`
- * on file if the parent directory is provided.
+ * Ensures existence of `_id` and `_type`
  *
+ * @public
  * @param {object} file - object representing the file
- * @param {object} parent - object representing the parent directory
  * @returns {object} full normalized object
  */
-export function normalize(file, parent) {
+export function normalize(file) {
   const id = file._id || file.id
   const type = file.type || file._type || 'io.cozy.files'
-  const path = parent && parent.path ? parent.path + '/' + file.name : undefined
-  return {
-    _id: id,
-    id: id,
-    _type: type,
-    type: type,
-    path: path,
-    ...file
-  }
+  return { _id: id, id, _type: type, type, ...file }
+}
+
+/**
+ * Ensure the file has a `path` attribute, or build it
+ *
+ * @public
+ * @param {object} file - object representing the file
+ * @param {object} parent - parent directory for the file
+ * @returns {object} file object with path attribute
+ */
+export function ensureFilePath(file, parent) {
+  if (file.path) return file
+  if (!parent || !parent.path)
+    throw new Error(`Could not define a file path for ${file._id || file.id}`)
+  const path = parent.path + '/' + file.name
+  return { path: path, ...file }
 }
--- a/packages/cozy-client/src/models/file.js
+++ b/packages/cozy-client/src/models/file.js
@@ -30,3 +30,27 @@ export const isNote = file => {
     return true
   return false
 }
+
+/**
+ * Normalize an object representing a io.cozy.file object
+ *
+ * Ensure existence of `_id` and `_type` and define a `path`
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/models/file.js
```diff
@@ -30,3 +30,27 @@ export const isNote = file => {
...
+  const id = file._id || file.id
+  const type = file.type || file._type || 'io.cozy.files'
+  const path = parent ? parent.path + '/' + file.name : undefined
+  return {
+    _id: id,
+    id: id,
+    _type: type,
+    type: type,
+    path: path,
+    ...(file || {})
```
drazik: Nit: `...file` should be enough.

```js
const obj = { ...undefined }
console.log(obj)
// {}
```

packages/cozy-client/src/models/file.spec.js
```diff
@@ -29,4 +29,36 @@ describe('File Model', () => {
...
+      const doc = { id }
+      expect(file.normalize(doc)._type).toBe('io.cozy.files')
+    })
+
+    it('should define the path if parent is provided', () => {
+      const doc = { id, name: 'myfile.ext' }
+      const parent = { path: '/my/path' }
+      expect(file.normalize(doc, parent).path).toBe('/my/path/myfile.ext')
+    })
+  })
```
Crash--: Can you add a test when there is no path? Since you have a condition on it

edas: A parent is a directory, so it should always have a path. I may throw an error if a path is missing but I usualy test only how a function should be used, and not what erroneous parameters could do.

edas: I have added two tests, one on missing `parent` and one for a `parent` without `path`. Is it what you missed ?

packages/cozy-client/src/models/file.js
```diff
@@ -30,3 +30,27 @@ export const isNote = file => {
     return true
   return false
 }
+
+/**
+ * Normalize an object representing a io.cozy.file object
+ *
+ * Ensure existence of `_id` and `_type` and define a `path`
```
ptbrowne:
```suggestion
 * Ensures existence of `_id` and `_type` and defines a `path`
```

packages/cozy-client/src/models/file.js
```diff
@@ -30,3 +30,27 @@ export const isNote = file => {
...
+ * on files if provided the parent directory.
```
ptbrowne:
```suggestion
 * on file if the parent directory is provided.
```

packages/cozy-client/src/models/file.js
```diff
@@ -30,3 +30,27 @@ export const isNote = file => {
...
+ *
+ * @param {object} file - object representing the file
+ * @param {object} parent - object representing the parent directory
+ * @returns {object} full normalized object
```
ptbrowne: Is it a private or public method ?

packages/cozy-client/src/models/file.js
```diff
@@ -30,3 +30,27 @@ export const isNote = file => {
...
+  const type = file.type || file._type || 'io.cozy.files'
+  const path = parent && parent.path ? parent.path + '/' + file.name : undefined
+  return {
+    _id: id,
+    id: id,
+    _type: type,
+    type: type,
+    path: path,
+    ...file
+  }
```
ptbrowne: In other parts of cozy-client's code, normalize only deals with _id and _type. Here we add logic about the path, are we sure that it should be done in another function ?

Also, what happens if file has a path and parent has a path ? Here the file's path would override the parent's path, is is something that we want ? It seems like an unwanted side-effect, this is why I am a bit hesitant with having the path logic inside the normalize method.

edas: Logic was "Ensure it has a path, build it from parent if needed", like we ensure an _id and a _type (but won't override the document's ones if they are present)

packages/cozy-client/src/models/file.js
edas: Today it's used by cozy-sharing, so public.

I've added the corresponding jsdoc

edas:
I've separated the two methods, let me know if it's better for you

PR #607, opened 2020-01-03
docs: Fix broken link

Gara64:


Gara64:
> Good catch ! We should find a way to automatically test all of our links in the doc...

Agreed, I created an [issue](https://github.com/cozy/cozy.github.io/issues/160) on the doc repo, as this is not limited to cozy-client

PR #612, opened 2020-01-08
 feat: Use bookmark pagination for queryAll

Gara64:
This enables [bookmark pagination](https://docs.couchdb.org/en/2.2.0/api/database/find.html#pagination) for "all" documents queries (the "find" queries will come afterwards).

It also adds a performance example, with some measures:
- 1 000 docs:
    - no pagination: ~108 ms
    - skip pagination: ~1 429 ms
    - bookmark pagination: ~563 ms
 
- 10 000 docs:
  - no pagination: ~932 ms
  - skip pagination: ~165 741ms
  - bookmark pagination: ~7 850 ms

### DIFF
--- a/packages/cozy-client/src/store/queries.js
+++ b/packages/cozy-client/src/store/queries.js
@@ -1,7 +1,7 @@
 import mapValues from 'lodash/mapValues'
-import union from 'lodash/union'
 import difference from 'lodash/difference'
 import intersection from 'lodash/intersection'
+import concat from 'lodash/concat'
 import isPlainObject from 'lodash/isPlainObject'
 
 import { getDocumentFromSlice } from './documents'
@@ -141,8 +141,10 @@ const updateData = (query, newData) => {
   const toUpdate = intersection(originalIds, matchedIds)
 
   const changed = toRemove.length || toAdd.length || toUpdate.length
-
-  const updatedData = difference(union(originalIds, toAdd), toRemove)
+  // concat doesn't check duplicates (contrarily to union), which is ok as
+  // toAdd does not contain any id present in originalIds, by construction.
+  // It is also faster than union.
+  const updatedData = difference(concat(originalIds, toAdd), toRemove)
 
   return {
     ...query,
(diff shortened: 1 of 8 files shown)


packages/cozy-client/src/store/queries.js
```diff
@@ -141,8 +141,9 @@ const updateData = (query, newData) => {
   const toUpdate = intersection(originalIds, matchedIds)
 
   const changed = toRemove.length || toAdd.length || toUpdate.length
-
-  const updatedData = difference(union(originalIds, toAdd), toRemove)
+  // concat doesn't check duplicates (contrarily to union), which is ok as
+  // toAdd has no originalIds, and is faster.
```
ptbrowne: // toAdd does not contain any id present in originalIds, by construction

? I had trouble to understand at first.

packages/cozy-stack-client/src/DocumentCollection.spec.js
```diff
@@ -144,6 +152,25 @@ describe('DocumentCollection', () => {
...
+        rows: [{}],
+        total_rows: 3
+      })
+      const respWithoutNext = await collection.all({
+        limit: 2,
+        bookmark: 'mark'
+      })
+      expect(respWithoutNext.next).toBe(false)
+    })
+
```
ptbrowne: Nit: I think it's better to add the tests inside the commit adding the functionality, I was actually looking for them when I saw the implementation commit.

packages/cozy-stack-client/src/DocumentCollection.spec.js
```diff
@@ -176,7 +176,7 @@ describe('DocumentCollection', () => {
       await collection.all({ keys: ['abc', 'def'] })
       expect(client.fetchJSON).toHaveBeenCalledWith(
         'GET',
-        '/data/io.cozy.todos/_all_docs?include_docs=true&keys=[%22abc%22,%22def%22]'
+        '/data/io.cozy.todos/_all_docs?include_docs=true&limit=100&keys=[%22abc%22,%22def%22]'
```
ptbrowne: This commit should be squashed/rebased, we should not have any commit that is red otherwise we can have bad surprises when bisecting.

packages/cozy-client/src/store/queries.js
Gara64: Right, I'll clarify

packages/cozy-stack-client/src/DocumentCollection.spec.js
Gara64: I actually prefer to have a separated commit for the test, to keep the commits focus on one thing (feature, test, doc, etc), but no problem to squash that

PR #618, opened 2020-01-16
fix: Refresh token on expiration

edas:
When Cozy-Client detect an expired token, it calls
`setToken` with the result of `refreshToken`.

Problem : refreshToken returns an AppToken but
setToken expect a jwt string.

OAuthClient has a code dealing with this. StackClient
doesn't. This change fix StackClient.

### DIFF
--- a/packages/cozy-stack-client/src/CozyStackClient.js
+++ b/packages/cozy-stack-client/src/CozyStackClient.js
@@ -1,6 +1,7 @@
 import cloneDeep from 'lodash/cloneDeep'
 import AppCollection, { APPS_DOCTYPE } from './AppCollection'
 import AppToken from './AppToken'
+import AccessToken from './AccessToken'
 import DocumentCollection from './DocumentCollection'
 import FileCollection from './FileCollection'
 import JobCollection, { JOBS_DOCTYPE } from './JobCollection'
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -254,9 +255,22 @@ class CozyStackClient {
...
+    if (!token) {
+      this.token = null
+    } else {
+      if (token.toAuthHeader) {
+        // AppToken or AccessToken
+        this.token = token
+      } else {
+        // jwt string
+        this.token = new AppToken(token)
+      }
```
Crash--: Why not the same as `OauthClient` ? :
```jsx 
      setToken(token) {
    if (token) {
      this.token = token instanceof AccessToken ? token : new AccessToken(token)
    } else {
      this.token = null
    }
  }
```

edas: Duck typing. I do not see any good reason to limit the token class to AppToken. If someone ever want to send an AccessToken or something else with the same interface, that's fine with me.

packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -254,9 +255,22 @@ class CozyStackClient {
...
-    this.token = token ? new AppToken(token) : null
-    if (token) {
+    if (!token) {
+      this.token = null
+    } else {
+      if (token.toAuthHeader) {
+        // AppToken or AccessToken
+        this.token = token
+      } else {
+        // jwt string
```
ptbrowne: Should we be more defensive here and checek typeof token ?

```
else if (typeof token === 'string') {
  this.token = new AppToken(token)
} else {
  console.warn('Bad token', token)
  throw new Error('Bad token')
}
```

PR #619, opened 2020-01-20
feat: Basic Notes collection

y-lohse:
This PR adds a basic collection for `io.cozy.notes`. Here is the corresponding doc: https://docs.cozy.io/en/cozy-stack/notes/

I say basic because:

1. It doesn't support all the special routes
2. It has no pagination support yet, I'm creating an issue detailing the problem right away

### DIFF
--- a/packages/cozy-stack-client/src/NotesCollection.js
+++ b/packages/cozy-stack-client/src/NotesCollection.js
@@ -1,0 +1,54 @@
+import DocumentCollection from './DocumentCollection'
+import { uri } from './utils'
+
+export const NOTES_DOCTYPE = 'io.cozy.notes'
+
+const normalizeDoc = DocumentCollection.normalizeDoctypeJsonApi(NOTES_DOCTYPE)
+const normalizeNote = note => ({
+  ...normalizeDoc(note),
+  ...note.attributes
+})
+
+/**
+ * Implements `DocumentCollection` API to interact with the /notes endpoint of the stack
+ */
+class NotesCollection extends DocumentCollection {
+  constructor(stackClient) {
+    super(NOTES_DOCTYPE, stackClient)
+  }
+
+  /**
+   * all - Fetches all notes
+   *
+   * @returns {{data, links, meta}} The JSON API conformant response.
+   */
+  async all() {
+    //TODO: add support for pagination
+    const resp = await this.stackClient.fetchJSON('GET', '/notes')
+
+    return {
+      ...resp,
+      data: resp.data.map(normalizeNote)
+    }
+  }
+
+  /**
+   * destroy - Destroys the note on the server
+   *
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/NotesCollection.js
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * destroy - Destroys the note on the server
+   *
+   * @param {io.cozy.notes} note     The note document to destroy
+   * @param {string}   note._id The note's id
+   *
+   * @returns {{ data }} The deleted note
+   */
+  async destroy({ _id }) {
```
ptbrowne: Is the note really deleted or is it just marked as deleted ? Why do we have to return the doc with _deleted ?

packages/cozy-stack-client/src/NotesCollection.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    it('should call the appropriate route', async () => {
+      jest.spyOn(stackClient, 'fetchJSON').mockResolvedValue({
+        data: [],
+        links: {},
+        meta: { count: 0 }
+      })
+
+      await collection.all()
+      expect(stackClient.fetchJSON).toHaveBeenCalledWith('GET', '/notes')
+    })
```
ptbrowne: Normalization is not tested, is it expected  ?

packages/cozy-stack-client/src/NotesCollection.spec.js
```diff
@@ -0,0 +1,50 @@
+import NotesCollection from './NotesCollection'
+import CozyStackClient from './CozyStackClient'
+
+describe('NotesCollection', () => {
+  const stackClient = new CozyStackClient()
+  const collection = new NotesCollection(stackClient)
```
ptbrowne: The setup pattern seems better (where we do not instantiate suite global variables) since we are sure that tests do not interfere with each other.

packages/cozy-stack-client/src/NotesCollection.js
y-lohse: It is really deleted, but without the `_deleted` flag the document won't be removed from queries; I Was under the impression this was the way to do it, [for example here](https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/DocumentCollection.js#L258).

packages/cozy-stack-client/src/NotesCollection.spec.js
y-lohse: No, will add it.

packages/cozy-stack-client/src/NotesCollection.js
ptbrowne: I had forgotten about that, thanks !

PR #622, opened 2020-01-21
feat: HasMany prunes absent documents before passing them down

ptbrowne:
BREAKING:

- HasMany::data does not return null values
- existsById changes of semantics
- containsById retains the previous behavior of existsById

---

- containsById does not check for document existence in the store
- existsById checks for existence of the document in the store

### DIFF
--- a/packages/cozy-client/src/associations/HasMany.js
+++ b/packages/cozy-client/src/associations/HasMany.js
@@ -52,15 +52,22 @@ class HasMany extends Association {
   }
 
   get data() {
-    return this.getRelationship().data.map(({ _id, _type }) =>
-      this.get(_type, _id)
-    )
+    return this.getRelationship()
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/associations/HasMany.js
```diff
@@ -52,15 +52,22 @@ class HasMany extends Association {
...
   get hasMore() {
     return this.getRelationship().next
   }
 
+  /**
+   * Returns the total number of documents in the relationship.
+   * Does not handle documents absent from the store. If you want
+   * to do that, you can use .data.length.
+   *
+   * @returns {Numb} - Total number of documents in the relationships
```
drazik:
```suggestion
   * @returns {Number} - Total number of documents in the relationships
```

PR #623, opened 2020-01-22
Add get by id for apps and konnectors

Gara64:
This implements `get(id)` for apps and konnectors and adds the associated doc and tests.
This allow to call the routes `apps/<id>` for an app, and `konnectors/id` for a konnector.


packages/cozy-stack-client/src/AppCollection.js
```diff
@@ -35,8 +36,16 @@ class AppCollection extends DocumentCollection {
...
+  /**
+   * Get an app by id
+   *
+   * @param  {string} id The document id.
+   * @returns {object}  JsonAPI response containing normalized document as data attribute
+   */
+  async get(id) {
+    return Collection.get(this.stackClient, `${this.endpoint}${id}`, {
+      normalize: this.constructor.normalizeDoctype(this.doctype)
+    })
```
ptbrowne: Instead of copy/pasting the method get here (from DocumentCollection), I think it would be better (in the sense that it would be more DRY) to have :

```
    const endpoint = this.endpoint || `/data/${this.doctype}/`
    return Collection.get(this.stackClient, uri`${this.endpoint}${id}`, {
      normalize: this.constructor.normalizeDoctype(this.doctype)
    })
```

in DocumentCollection.get.

What do you think ?

I see that the `uri` function is not used on the template string also, is there a particular reason for that ?

ptbrowne: Better, maybe DocumntCollection could set `this.endpoint` in its constructor. This would remove the repetition of `/data/${this.doctype}` in the DocumentCollection code and would make it symmetrical with AppCollection.

Gara64: Agreed with that. The only thing is, I didn't use `uri` because the `/apps/id` was encoded in `%2Fapps%2Fid`, which was failing with a `'400 Bad Request'`.
Any idea why?

ptbrowne: Yes. uri works on every part of the template literal expressions. So in

```
uri`/data/${urlPart}`
```

Only ${urlPart} will go through the encodeURIComponent. When ${endpoint} goes inside, it gets uri-encoded.

See https://github.com/cozy/cozy-client/blob/7e38f5b06493184c102c7ace2ffab5965e5249b8/packages/cozy-stack-client/src/utils.js#L1 for the implementation.

So `uri` should not be applied to all the URL if ${endpoint} is used. I am not sure `uri` is very ergonomic/explicit here. It is easy to misuse. Feel free to use another technique to be sure that parts that need to be encoded are encoded ;)

Gara64: Ok, thanks for the literal insights! See the [new commit](058e239a811f60839ce95cc764a8b8fa843b7640)

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -35,6 +35,7 @@ class DocumentCollection {
     this.doctype = doctype
     this.stackClient = stackClient
     this.indexes = {}
+    this.endpoint = '/data/'
```
ptbrowne: `this.endpoint = '/data/${this.doctype}/'` no ?

Gara64: Wops, right... Tests was missing here, I added some

PR #625, opened 2020-01-23
fix: Revoke sharing links fetch all the related permissions

Crash--:
If we have more than 30 sharing by links (/sharing-by-links is paginated at 30 per default), we had a bug when deleting a file with a related sharing. 

We now fetch all the sharing by links (by following resp.links.next) and then we map on them to check if they are concerned by the removal. If yes, we destroy the permission.

### DIFF
--- a/packages/cozy-stack-client/src/PermissionCollection.spec.js
+++ b/packages/cozy-stack-client/src/PermissionCollection.spec.js
@@ -158,6 +158,92 @@ describe('PermissionCollection', () => {
   })
 })
 
+describe('revokeSharingLink', () => {
+  const client = new CozyStackClient()
+  const collection = new PermissionCollection('io.cozy.permissions', client)
+
+  beforeEach(() => {
+    client.fetchJSON.mockReset()
+    client.fetchJSON.mockReturnValue(Promise.resolve({ data: [] }))
+  })
+
+  it('should revoke a sharing link', async () => {
+    client.fetchJSON.mockReturnValue(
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/PermissionCollection.spec.js
```diff
@@ -158,6 +158,92 @@ describe('PermissionCollection', () => {
...
+      'GET',
+      '/permissions/doctype/io.cozy.files/shared-by-link'
+    )
+
+    expect(client.fetchJSON).toHaveBeenNthCalledWith(
+      2,
+      'GET',
+      '/permissions/next_page'
+    )
+  })
```
y-lohse: Nit, but an even better test might have been to check that we delete the permission on a result of the second page, as this would test the initial problem rather than the internal implementation :)

Crash--: You right, i'll update it ASAP

nono:
What is the status of this PR?

Crash--:
I'll not the time to update the test. Are you agreed to merge without @y-lohse ?

nono:
@y-lohse can we merge this PR?

y-lohse:
Oops, missed the previous message.  But yes it was marked as nit so I'm happy to merge this as it is.

nono:
If this PR is still relevant, we should merge it!

ptbrowne:
Ping @Crash--

PR #628, opened 2020-01-31
feat: Add generateWebLink helper

taratatach:
This helper meant to build URLs to Cozy applications lived in
`cozy-ui` although it would be useful to apps that don't need
`cozy-ui`.

This version adds the possibility to pass search parameters as well.
This should be useful for apps that need a share code for example.

### DIFF
--- a/packages/cozy-client/src/helpers.js
+++ b/packages/cozy-client/src/helpers.js
@@ -18,3 +18,56 @@ export const dehydrate = document => {
   )
   return dehydrated
 }
+
+const ensureFirstSlash = path => {
+  if (!path) {
+    return '/'
+  } else {
+    return path.startsWith('/') ? path : '/' + path
+  }
+}
+
+/**
+ * generateWebLink - Construct a link to a web app
+ *
+ * This function does not get its cozy url from a CozyClient instance so it can
+ * be used to build urls that point to other Cozies than the user's own Cozy.
+ * This is useful when pointing to the Cozy of the owner of a shared note for
+ * example.
+ *
+ * @param {object} options               Object of options
+ * @param {string}   options.cozyUrl       Base URL of the cozy, eg. cozy.tools or test.mycozy.cloud
+ * @param {Array}    options.searchParams  Array of search parameters as [key, value] arrays, eg. ['username', 'bob']
+ * @param {string}   options.pathname      Path to a specific part of the app, eg. /public
+ * @param {string}   options.hash          Path inside the app, eg. /files/test.jpg
+ * @param {string}   options.slug          Slug of the app
+ * @param {string}   options.subDomainType Whether the cozy is using flat or nested subdomains. Defaults to flat.
+ *
+ * @returns {string} Generated URL
+ */
+export const generateWebLink = ({
+  cozyUrl,
+  searchParams = [],
+  pathname,
+  hash,
+  slug,
+  subDomainType
+}) => {
+  const url = new URL(cozyUrl)
+
+  url.host =
+    subDomainType === 'nested'
+      ? `${slug}.${url.host}`
+      : url.host
+          .split('.')
+          .map((x, i) => (i === 0 ? x + '-' + slug : x))
+          .join('.')
+  url.pathname = pathname
+  url.hash = ensureFirstSlash(hash)
+
+  for (const [param, value] of searchParams) {
+    url.searchParams.set(param, value)
+  }
+
+  return url.toString()
+}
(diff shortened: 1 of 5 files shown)


packages/cozy-client/src/helpers.js
```diff
@@ -1,4 +1,5 @@
 import { Association } from './associations'
+import { URL } from 'url'
```
drazik: This dep doesn't seem to be in package.json. Is it normal?

packages/cozy-client/src/helpers.js
```diff
@@ -18,3 +19,46 @@ export const dehydrate = document => {
...
+      ? `${slug}.${url.host}`
+      : url.host
+          .split('.')
+          .map((x, i) => (i === 0 ? x + '-' + slug : x))
+          .join('.')
+  for (const [param, value] of searchParams) {
+    url.searchParams.set(param, value)
+  }
+  url.hash = nativePath
+  return url.toString()
```
ptbrowne: Nit: It feels like the ergonomic of this function are a bit off. The subDomainType could be taken from the client, the cozyURL too, this would simplify the options.

Crash--: Yep I plan to rewrite this method specially since this method is now living in CozyClient. 

We still need to mimic https://github.com/cozy/cozy-client/blob/61f5de62a98ac58d04a09d2d955842c4fdfec638/packages/cozy-client/src/CozyClient.js#L1176 for an OAuth client ;)

taratatach: This method will be used to build urls that point to another Cozy than the one used by the client.
This is why I put this function in the `helpers` module.

packages/cozy-client/src/helpers.js
taratatach: No it's not. Actually this is needed when running the code in Node but I believe it's not when running it in the browser.
But I don't know how to have different dependencies depending on the target :/.

packages/cozy-client/src/helpers.spec.js
```diff
@@ -109,3 +109,33 @@ describe('dehydrate', () => {
...
+      })
+    ).toEqual(
+      `http://alice-notes.cozy.tools/?sharecode=${sharecode}&username=${username}#/n/4`
+    )
+  })
+
+  it('should generate the right link to a nested cozy', () => {
+    expect(
+      generateWebLink({
+        cozyUrl: 'https://alice.cozy.tools/public/',
```
Crash--: I don't like to put the "public" path to the cozyURL. CozyURL should only be the base domain.

taratatach:
I would vote for exporting the helpers

packages/cozy-client/src/helpers.spec.js
taratatach: I thought of adding a `publicView` boolean option but then I thought that using the `cozyUrl` option would be more generic.

Crash--: I don't think so. `https://alice.cozy.tools/public/` is not a valid URL and should not be put to a `cozyURL` var. 

Today I would prefer to have a `path` attribute since some apps can have : `foo.cozy.tools/public?shareCode` and others `foo.cozy.tools/public/?shareCode`

packages/cozy-client/src/helpers.js
taratatach: Actually tests are fine without it now. I don't know why it was an issue before.

drazik:
> I would vote for exporting the helpers

Me too. At least this one :)

packages/cozy-client/src/helpers.spec.js
```diff
@@ -109,3 +109,33 @@ describe('dehydrate', () => {
...
+describe('generateWebLink', () => {
+  it('should generate the right link to a flat cozy', () => {
+    const sharecode = 'sharingIsCaring'
+    const username = 'alice'
+
+    expect(
+      generateWebLink({
+        cozyUrl: 'http://alice.cozy.tools',
+        searchParams: [['sharecode', sharecode], ['username', username]],
+        nativePath: '/n/4',
```
Crash--: (this was for the console.warn)
It's a bit brutal since we already have method to generate an url for an App. 


I think the Note Model should export the `#/n/` path.

packages/cozy-client/src/helpers.spec.js
taratatach: Why is it not valid? I'm ok with adding a `path` parameter though. Would you rather have a `/` at the end or not?

Crash--: By not valid I mean that the URL doesn't exist. 
```
>curl -o /dev/null --silent --head --write-out '%{http_code}\n' https://crash-notes.mycozy.cloud/public/
200

> curl -o /dev/null --silent --head --write-out '%{http_code}\n' https://crash.mycozy.cloud/public/
404
```

We should follow the URL spec and use pathname in fact : https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname

PR #629, opened 2020-01-31
feat: Add Notes collection `fetchURL` method

taratatach:
This method calls the `/notes/:id/open` stack endpoint to fetch the
required details to build a url that points to the note in the Notes
application on the note's owner's Cozy.

This means that a shared note will return a NoteUrl that points to the
owner Cozy, with the original note's id and a share code to get
permissions to see the note.
If the note is not shared then the details will point to the client's
Cozy and no share code.

### DIFF
--- a/packages/cozy-stack-client/src/NotesCollection.js
+++ b/packages/cozy-stack-client/src/NotesCollection.js
@@ -2,11 +2,17 @@ import DocumentCollection from './DocumentCollection'
 import { uri } from './utils'
 
 export const NOTES_DOCTYPE = 'io.cozy.notes'
+export const NOTES_URL_DOCTYPE = 'io.cozy.notes.url'
 
 const normalizeDoc = DocumentCollection.normalizeDoctypeJsonApi(NOTES_DOCTYPE)
 const normalizeNote = note => ({
   ...normalizeDoc(note),
   ...note.attributes
+})
+
+const normalizeNoteUrl = noteUrl => ({
+  ...DocumentCollection.normalizeDoctypeJsonApi(NOTES_URL_DOCTYPE)(noteUrl),
+  ...noteUrl.attributes
 })
 
 /**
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/NotesCollection.js
```diff
@@ -47,6 +53,26 @@ class NotesCollection extends DocumentCollection {
       data: { ...normalizeNote(resp.data), _deleted: true }
     }
   }
+
+  /**
+   * open - Returns the details to build the note's url
```
ptbrowne:
```suggestion
   * Returns the details to build the note's url
```

packages/cozy-stack-client/src/NotesCollection.js
```diff
@@ -47,6 +53,26 @@ class NotesCollection extends DocumentCollection {
...
+   *
+   * @see https://github.com/cozy/cozy-stack/blob/master/docs/notes.md#get-notesidopen
+   *
+   * @param {io.cozy.notes} note The note document to open
+   * @param {string}   note._id The note's id
+   *
+   * @returns {{ data }} The note's url details
+   */
+  async open({ _id }) {
```
ptbrowne: When I read the description, I find that `fetchDetails` or `fetchURL` would be more appropriate ?

packages/cozy-stack-client/src/NotesCollection.js
taratatach: I'm just following the naming scheme of the comments in this file. Should I remove the name of the methods in all of them?

packages/cozy-stack-client/src/NotesCollection.js
taratatach: Sounds good to me.

packages/cozy-stack-client/src/NotesCollection.spec.js
```diff
@@ -73,4 +73,58 @@ describe('NotesCollection', () => {
       expect(result.data._deleted).toBe(true)
     })
   })
+
+  describe('open', () => {
```
ptbrowne:
```suggestion
  describe('fetchURL', () => {
```

PR #635, opened 2020-02-05
Add bookmark support for documents find queries

Gara64:
Add support for bookmark pagination for documents `find` queries. For the record, bookmark are now the [preferred way to paginate](https://github.com/cozy/cozy-stack/blob/master/docs/mango.md#pagination-cookbook), as it performs way faster than skip.

Note we remove the `meta: {count}` response, as we can no longer know the total count of returned docs, at the query level. This shouldn't be a problem as we can easily retrieve this information by doing a `data.length`, but it might break existing code if some apps are relying on it.

### DIFF
--- a/packages/cozy-client/examples/perfs.js
+++ b/packages/cozy-client/examples/perfs.js
@@ -60,7 +60,29 @@ const getAllContactsBookmark = async () => {
     doctype: 'io.cozy.contacts',
     limit: 100
   })
-  const contacts = await client.queryAll(query, { as: 'bookmark-query' })
+  const contacts = await client.queryAll(query, { as: 'bookmark-all-query' })
+  return contacts
+}
+
+const getFilteredContactsBookmark = async () => {
+  const selector = {
+    _id: {
+      $gt: null
+    }
+  }
+  const query = new QueryDefinition({
+    doctype: 'io.cozy.contacts',
+    limit: 100,
+    selector
+  })
+  const contacts = []
+  let resp = { next: true, bookmark: '' }
+  while (resp && resp.next) {
+    resp = await client.query(query.offsetBookmark(resp.bookmark), {
+      as: 'bookmark-find-query'
+    })
+    contacts.push(...resp.data)
+  }
   return contacts
 }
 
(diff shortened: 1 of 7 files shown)


edas:
> it might break existing code if some apps are relying on it

Why not pass a `meta: { get count() { return data.length } }` then? I will avoid any breakage of compatibility. We could even add a deprecation warning if needed

ptbrowne:
> Note we remove the meta: {count} response, as we can no longer know the total count of returned docs, at the query level. This shouldn't be a problem as we can easily retrieve this information by doing a data.length, 

If I am not mistaken, `data.lengh` is not the same as `meta.count` since `data.length` will give us the number of fetched documents from the query. If the implicit limit is 30 for example, for a query that returns in total 100 documents, if we have fetched only the first page of results data.length = 30 whereas meta.count = 100 documents returned

edas:
> If I am not mistaken, `data.lengh` is not the same as `meta.count` since `data.length` 

Right, sorry. In that case this is a breaking change and a quick search in our  code base is necessary to see if someone did require the `count` to be there

Gara64:
> If I am not mistaken, `data.lengh` is not the same as `meta.count` since `data.length`

Right, I meant on the app-side, after having retrieved all the data. 

> Right, sorry. In that case this is a breaking change and a quick search in our code base is necessary to see if someone did require the `count` to be there

From what I see, this `count` is necessary in cozy-drive [here](https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/navigation/duck/async.js#L40-L42) and cozy-home [here](https://github.com/cozy/cozy-home/blob/master/src/lib/redux-cozy-client/adapters/CozyStackAdapter.js#L19), but both are using cozy-client-js and it seems we could avoid the `count` with cozy-client or a `.length`. The exception using cozy-client is [here](https://github.com/cozy/cozy-drive/blob/master/src/photos/ducks/clustering/files.js#L30-L31) but I think this part could be fixed (and is actually inherited from cozy-client-js). I didn't find other occurences in banks, contacts, store, bar, notes, settings.

PR #636, opened 2020-02-07
feat: Checks if a doc has read-only permissions

edas:
It is needed in cozy-notes, to see if we  should allows
write when displaying a shared document.

### DIFF
--- a/packages/cozy-client/src/models/file.js
+++ b/packages/cozy-client/src/models/file.js
@@ -1,3 +1,5 @@
+import get from 'lodash/get'
+
 const FILE_TYPE = 'file'
 const DIR_TYPE = 'directory'
 
@@ -60,4 +62,15 @@ export function ensureFilePath(file, parent) {
     throw new Error(`Could not define a file path for ${file._id || file.id}`)
   const path = parent.path + '/' + file.name
   return { path: path, ...file }
+}
+
+/**
+ * Get the id of the parent folder (`null` for the root folder)
+ *
+ * @param {object} file  - io.cozy.files document
+ * @returns {string|null} id of the parent folder, if any
+ */
+export function getParentFolderId(file) {
+  const parentId = get(file, 'attributes.dir_id')
+  return parentId === '' ? null : parentId
 }
--- a/packages/cozy-client/src/models/permission.js
+++ b/packages/cozy-client/src/models/permission.js
@@ -1,0 +1,142 @@
+import intersection from 'lodash/intersection'
+import get from 'lodash/get'
+import CozyClient from 'cozy-client'
+import { getParentFolderId } from './file'
+
+/**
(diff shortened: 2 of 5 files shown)


packages/cozy-client/src/models/file.js
```diff
@@ -61,3 +63,13 @@ export function ensureFilePath(file, parent) {
   const path = parent.path + '/' + file.name
   return { path: path, ...file }
 }
+
+/**
+ * Get the id of the parent folder
+ *
+ * @param {object} file  - io.cozy.files document
+ * @returns {string|undefined} id of the parent folder, if any
```
ptbrowne: What does it mean if there is no parent folder ?

packages/cozy-client/src/models/permission.js
```diff
@@ -0,0 +1,139 @@
...
+ * @property {string[]} values
+ * @property {string} type - a couch db database like 'io.cozy.files'
+ */
+
+/**
+ * Is this permission read only ?
+ *
+ * @private
+ * @param {PermissionItem} perm - permission node in a io.cozy.permissions document
+ * @param {object} options -
```
ptbrowne: Nit: Throughout this file, there are end of line dashes, I think it's best to remove them.

packages/cozy-client/src/models/file.js
edas: It is usualy that youare at the root folder (or that your files database is somehow broken)

packages/cozy-client/src/models/permission.js
```diff
@@ -0,0 +1,139 @@
...
+ * Get the list of permissions blocks
+ * @private
+ */
+async function getOwn(client) {
+  const collection = client.collection('io.cozy.permissions')
+  const data = await collection.getOwnPermissions()
+  const permissions = get(data, 'data.attributes.permissions')
+  if (!permissions) throw `Can't get self permissions`
+  return Object.values(permissions)
+}
```
ptbrowne: I think it is clearer when method doing asynchronous things use the verb `fetch`.

packages/cozy-client/src/models/file.js
ptbrowne: Mmh I thought we had an id for the root folder, was I mistaken @nono ?

nono: We have a root (with the id `io.cozy.files.root-dir`), and it has no parent.

edas: Oh, I was wrong about the root having an undefined dir_id. It is defined to an empty string.

Let me fix this method so it returns `null` when at root.

packages/cozy-client/src/models/permission.js
edas: I've  put them toremove prettier warnings about not having descriptions for these param (and well, don't add useless descriptions)

PR #638, opened 2020-02-11
feat: Add deprecation message for withMutations

ptbrowne:
Instead of using withMutations, we have found that we prefer to use
withClient and methods using this client.

```
const doSomethingWithClient = client => {
  ...
}

const Component = withClient(({ client  }) => {
  const handleClick = () => {
    return doSomethingWithClient(client)
  }
  return <button onClick={handleClick}>click</button>
})
```

### DIFF
--- a/packages/cozy-client/src/withMutations.jsx
+++ b/packages/cozy-client/src/withMutations.jsx
@@ -12,8 +12,8 @@ const makeMutationsObject = (mutations, client, props) => {
 
 /**
  * @function
- * @description HOC to provide mutations to components. Needs client in context
- * or as prop.
+ * @description HOC to provide mutations to components. Needs client in context or as prop.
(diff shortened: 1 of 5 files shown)


packages/cozy-client/src/withMutations.jsx
```diff
@@ -12,8 +12,8 @@ const makeMutationsObject = (mutations, client, props) => {
 
 /**
  * @function
- * @description HOC to provide mutations to components. Needs client in context
- * or as prop.
+ * @description HOC to provide mutations to components. Needs client in context or as prop.
+ * This is deprecated, prefer to use withClient and access directly the client.
```
drazik: Can you add a @deprecated tag? https://jsdoc.app/tags-deprecated.html

ptbrowne: TIL, thanks

PR #639, opened 2020-02-12
fix: Consider data object from stat

Gara64:
`isChildOf` was failing if the given child wasn't an object and the hierarchy deeper than 1

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -466,8 +466,8 @@ class FileCollection extends DocumentCollection {
     }
     if (!childPath) {
       const childDoc = await this.statById(childID)
-      childPath = childDoc.path
-      childDirID = childDoc.dirID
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -466,8 +466,8 @@ class FileCollection extends DocumentCollection {
     }
     if (!childPath) {
       const childDoc = await this.statById(childID)
-      childPath = childDoc.path
-      childDirID = childDoc.dirID
+      childPath = childDoc.data.path
+      childDirID = childDoc.data.dirID
```
Crash--: can you add a test for that?

Gara64: Certainly : https://github.com/cozy/cozy-client/pull/639/commits/70b4d3f9c1d3be3c784bf0cab0eb4e6402190dac

Issue #646, opened 2020-02-18
Feature parity between <Query /> and client.query


y-lohse:
Based on:
- https://github.com/cozy/cozy-libs/pull/960#discussion_r380551744
- https://github.com/cozy/cozy-libs/pull/960#discussion_r380556922

The `<Query />` component (or queryConnect) has some extra features compared to calls to `client.query`, and these features could be useful. 

For example it would be good to use fetch policies. One use case for that are calls to `client.query` inside a hook, which will result in many requests.

Another aspect is the `fetchStatus` of the query, as we had to implement a similar mechanism in multiple hooks. Ideally we could use the status kept by cozy-client.

the biggest use case for these are probably within hooks, as most other use cases are covered by `queryConnect`. So this is something to keep in mind when we work on the `useQuery` hook (#547).

ptbrowne:
I think that now that we have useQuery, we can close ?

y-lohse:
Yep, I think so!

PR #651, opened 2020-02-24
feat: Add useAppLinkWithStoreFallback hook

Crash--:
Add the hook already used by Harvest and now Drive. It's time to add it to cozy-client. 

The goal of this hook is to give the link to the app or to the store if it's not installed yet 

I know we have the request status issue see https://github.com/cozy/cozy-client/issues/646


packages/cozy-client/src/node.js
```diff
@@ -31,3 +31,5 @@ export * from './cli'
 
 import * as models from './models'
 export { models }
+
+export * from './hooks'
```
y-lohse: Why do we want to export hooks for the node version?

Crash--: Currently, jest is based on our `main` entry point aka our node entry and not our browser one.

y-lohse: Mmh ok, but that means `react` is imported on node again, and I think that was a problem?

Crash--: fixed https://github.com/cozy/cozy-client/pull/651/commits/1b049d71996fe7ef7f065bf433c7ac649a824ef7

package.json
```diff
@@ -34,19 +34,43 @@
...
-    ],
-    "setupFilesAfterEnv": [
-      "<rootDir>/packages/cozy-stack-client/src/__tests__/setup.js"
-    ],
-    "testMatch": [
-      "**/(*.)(spec|test).js?(x)"
-    ],
-    "modulePathIgnorePatterns": [
-      "<rootDir>/packages/.*/dist/"
+    "projects": [
```
y-lohse: TIL

package.json
```diff
@@ -34,19 +34,43 @@
...
+          "jest-localstorage-mock"
+        ],
+        "setupFilesAfterEnv": [
+          "<rootDir>/packages/cozy-stack-client/src/__tests__/setup.js"
+        ],
+        "browser": true,
+        "modulePathIgnorePatterns": [
+          "<rootDir>/packages/.*/dist/"
+        ]
+      }
```
ptbrowne: Most of the configs are shared, could they be regrouped in a commonConfig and then a `merge` could be used to only change what differs between the two ?

Crash--: https://github.com/cozy/cozy-client/pull/651/commits/6cab6ed73a8fec3b53a17e898ddc8d00c6c9ebdc done thanks :)

PR #652, opened 2020-02-24
feat: Add fake import to trigger lint error

ptbrowne:
I want to test that the script on Travis correctly exits as soon as
one command fails, and that it makes the build red.


ptbrowne:
Close as it was only a test to check Travis' error handling after refactor of the `script` part of `.travis.yml`.

PR #655, opened 2020-02-25
fix: Allow custom token file path in createClientInteractive

doubleface:
lodash/merge option were in the wrong order for server options

### DIFF
--- a/packages/cozy-client/src/cli/index.spec.js
+++ b/packages/cozy-client/src/cli/index.spec.js
@@ -29,6 +29,7 @@ describe('createClientInteractive', () => {
       servers.push(server)
       return server
     })
+    fs.writeFileSync.mockClear()
   })
 
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/cli/index.spec.js
```diff
@@ -55,7 +56,7 @@ describe('createClientInteractive', () => {
     OAuthClient.prototype.fetchAccessToken.mockRestore()
   })
 
-  const setup = () => {
+  const setup = (withCustomFilePath = false) => {
```
ptbrowne: I think it would more extensible if instead of `withCustomFilePath`, we'd use an option object with `getSavedCredentials`.

```
   const setup = ({ getSavedCredentials } = {}) => { 
```

This way, in the test for custom file path, we can have

```
setup({ getSavedCredentials: () => 'custom file path' })
```

which would be more understandable IMHO. It lets define the mocked file path in the same
function where we `expect` it which is a plus.

Finally, we do not have to have a new "concept" `withCustomFilePath`.

PR #660, opened 2020-02-27
CreateClientInteractive improvements

ptbrowne:
### Remove open handles

After resolve/reject of the oauth process, it is important to clean all
timeouts, otherwise it leaves open handles that prevent node process
to completely shutdown. It was particularly visible since there was a
30 sec rejection timeout

### Option for fully automatic oauth client creation

See commit description for more details

### DIFF
--- a/packages/cozy-client/src/cli/index.js
+++ b/packages/cozy-client/src/cli/index.js
@@ -192,4 +192,22 @@ const createClientInteractive = (clientOptions, serverOpts) => {
   })
 }
 
+const main = async () => {
+  const client = await createClientInteractive({
+    scope: ['io.cozy.files'],
+    uri: 'http://cozy.tools:8080',
+    oauth: {
+      softwareID: 'io.cozy.client.cli'
+    }
+  })
+  console.log(client.toJSON())
+}
+
+if (require.main === module) {
+  main().catch(e => {
+    console.error(e)
+    process.exit(1)
+  })
+}
+
 export { createClientInteractive }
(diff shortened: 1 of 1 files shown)


packages/cozy-client/src/cli/index.js
```diff
@@ -63,7 +63,12 @@ const mkServerFlowCallback = serverOptions => authenticationURL =>
           'debug',
           'OAuth callback server started, waiting for authentication'
         )
-        opn(authenticationURL, { wait: false })
+        if (serverOptions.shouldOpenAuthenticationPage !== false) {
+          opn(authenticationURL, { wait: false })
+        }
+        if (serverOptions.onListen) {
+          serverOptions.onListen({ authenticationURL })
+        }
```
Crash--: There is no documentation about `serverOptions` ?

packages/cozy-client/src/cli/index.js
```diff
@@ -192,4 +192,22 @@ const createClientInteractive = (clientOptions, serverOpts) => {
...
+  console.log(client.toJSON())
+}
+
+if (require.main === module) {
+  main().catch(e => {
+    console.error(e)
+    process.exit(1)
+  })
+}
+
```
Crash--: Je comprends pas trop le message du commit par rapport au code ?

packages/cozy-client/src/cli/index.js
ptbrowne: Fixed in db5eab4

packages/cozy-client/src/cli/index.js
ptbrowne: This commits adds a way to execute the cli/index.js, to test the createClientInteractive method.

Crash--: Aaaah, makes sens now! Thanks

PR #547, opened 2019-10-18
Use query

ptbrowne:
- [x] useQuery hook
- [x] Ability to use fetchMore
- [x] Adding a playground

PR #665, opened 2020-03-01
useClient hook

ptbrowne:
We can now use a hook to get access to the CozyClient in
the context. Backward compatibility with withClient and
old context API is assured.

This is the first part of the ["hooks" PR](https://github.com/cozy/cozy-client/pull/547) that was
a bit too big, and takes a long time to be merged since
there is the useQuery work inside that still needs work. I think it is preferable
to split and get the useClient merged.

### DIFF
--- a/packages/cozy-client/src/hoc.jsx
+++ b/packages/cozy-client/src/hoc.jsx
@@ -2,6 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import compose from 'lodash/flowRight'
 import Query from './Query'
+import useClient from './hooks/useClient'
 
 /**
  * @function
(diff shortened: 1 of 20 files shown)


packages/cozy-client/src/hoc.jsx
```diff
@@ -11,13 +12,12 @@ import Query from './Query'
  * @returns {Function} - Component that will receive client as prop
  */
 export const withClient = Component => {
-  const Wrapped = (props, context) => (
-    <Component {...props} client={context.client} />
+  const Wrapped = props => (
+    <clientContext.Consumer>
+      {({ client }) => <Component {...props} client={client} />}
+    </clientContext.Consumer>
```
edas:
```suggestion
  const Wrapped = props => {
    const client = useClient()
    return <Component { ...props, client } />
  }
```

ptbrowne: Thanks, fixed :)

PR #666, opened 2020-03-05
Feat/add use fetch shortcut

Crash--:
We now have a new shortcut file's type and a dedicated API endpoint to retrieve the data. 

This hook is here to easily request them and also add a timeout to display the image (if we want). I added that, in order to be able to make the http request for the image before displaying it to avoid a blank between the placeholder and the final image

### DIFF
--- a/packages/cozy-client/src/hooks/useFetchShortcut.jsx
+++ b/packages/cozy-client/src/hooks/useFetchShortcut.jsx
@@ -1,0 +1,47 @@
+import { useState, useEffect } from 'react'
+
+const useFetchShortcut = (client, id) => {
+  const [shortcutInfos, setShortcutInfos] = useState()
+  const [shortcurtImg, setShotcutImg] = useState()
+  const [fetchStatus, setFetchStatus] = useState('idle')
+  const [shouldDisplayImg, setShouldDisplayImg] = useState(false)
+  useEffect(() => {
+    let timeout
+    const fetchData = async () => {
+      setFetchStatus('loading')
+      try {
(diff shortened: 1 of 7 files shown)


packages/cozy-client/src/hooks/useFetchShortcut.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+        const imgUrl = `${client.getStackClient().uri}/bitwarden/icons/${
+          shortcutRemoteUrl.host
+        }/icon.png`
+
+        setShotcutImg(imgUrl)
+        // this is used in conjonction with CSS display none/block to
+        // make the http request to load the image and not having a blank
+        // element for a few ms
+        timeout = setTimeout(() => setShouldDisplayImg(true), 400)
```
y-lohse: I don't understand this, why wait 400ms after getting the image to display it?

Crash--: You have the choice to display it before or not. I use it like a kind of preloader. Let's take a look at the current flow : 

- App => makes a request to /shortcut => wait a few ms to get the result from the stack => setShortCutInfo and build url => component refresh and replace the placeholder by . Then the app makes a request to shortcutinfo.url to get the image and then display it. 

Between those two last action, we display a blank space... The idea of this timeout is to reduce this blank and preload the image.

Crash--: As discussed together, we will remove it from the hook itself but create a new component to preload an image.

PR #668, opened 2020-03-10
fix: Make query data duplicate free

y-lohse:
This is not an actual PR at this point, I just. want to discuss the issue and evaluate if it needs fixing, and if so how.

The problem occurs after the following steps:

1. Init query `toto` 👉query data is empty
2. Receive results for that query 👉query data contains some ids
3. Init query `toto` again 👉[query data still contains the ids](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/queries.js#L41)
4. Receive the same results for the same query 👉[query data gets duplicated](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/queries.js#L74)

To me this is a legitimate use case, it can happen simply by navigating several times to a view that makes the same query.

I'm less sure about the correct fix. initially I felt that when we init a query, we should not keep the previous data (eg. fix step n°3). Not only because it's an `init`, but also because the definition could have changed.  
But fixing the problem that way caused [this test to fail](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/__tests__/ObservableQuery.spec.js#L127-L138), so it seems like we *do* want to keep the previous data around, presumably to display something while we load the new results, which I can understand.

So the other option is to fix at step n°4, it breaks no tests and is probably correct.

Another option would be to run the query against the stored documents at init time (using [this](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/queries.js#L120)). This would be a breaking change but possibly a good one? It means query's would immediately return cached results, and update again once the actual results are received.

Let me know what you think!

### DIFF
--- a/packages/cozy-client/src/store/queries.js
+++ b/packages/cozy-client/src/store/queries.js
@@ -3,6 +3,7 @@ import difference from 'lodash/difference'
 import intersection from 'lodash/intersection'
 import concat from 'lodash/concat'
 import isPlainObject from 'lodash/isPlainObject'
+import uniq from 'lodash/uniq'
 
 import { getDocumentFromSlice } from './documents'
 import { isReceivingMutationResult } from './mutations'
@@ -71,7 +72,7 @@ const query = (state = queryInitialState, action) => {
           response.meta && response.meta.count
             ? response.meta.count
             : response.data.length,
-        data: [...state.data, ...response.data.map(properId)]
+        data: uniq([...state.data, ...response.data.map(properId)])
       }
     }
     case RECEIVE_QUERY_ERROR:
--- a/packages/cozy-client/src/store/queries.spec.js
+++ b/packages/cozy-client/src/store/queries.spec.js
@@ -50,6 +50,13 @@ describe('queries reducer', () => {
         })
       )
       expect(state).toMatchSnapshot()
+      expect(state['a'].data).toEqual([TODO_1._id, TODO_2._id])
+      applyAction(
+        receiveQueryResult('a', {
+          data: [TODO_1, TODO_2]
+        })
+      )
+      expect(state['a'].data).toEqual([TODO_1._id, TODO_2._id])
     })
     it('should correctly update two queries', () => {
       applyAction(


Crash--:
We want to have the previous loaded data to make our "cache first strategy in action" ;).

I don't like the `uniq` approach it's not explicit enough for the result we keep. Is it the previous result ? The new one ? 

Also, in order to not have a new render cycle if the result doesn't change, we should make a few check before updating the store, no ?  (edit In fact since we store the lastFetch with Date() in this slice, it doesn't make sense to check if the previous result is the same ^^")

I don't understand your other proposal. Can you add a psuedo code or soemthing?

y-lohse:
> I don't understand your other proposal. Can you add a  but ac code or soemthing?

Yeah sure. It would involve changing [this](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/queries.js#L40-L45) to something like

```
return {
        ...state,
        id: action.queryId,
        definition: action.queryDefinition,
        fetchStatus: 'loading',
        data: findMatchingDocuments(documentsInStore, action.queryDefinition)
      }
```
…but actually it wouldn't solve the problem here. `data` would still contain the previous results, and when the new ones arrive they will still be duplicated. So really aside from making sure that `data` does not contain duplicate values, I don't see a fix here.

ptbrowne:
I think it's alright to use uniq here.

> I don't like the uniq approach it's not explicit enough for the result we keep. Is it the previous result ? The new one ?

Here the `data` is only containing ids. Full documents are contained in the `documents` slice.

I wonder though why we did not have a problem earlier ? Are the documents `uniqued` when passed to React components ? I am a bit stunned to have this problem actually.

y-lohse:
> I wonder though why we did not have a problem earlier ? Are the documents uniqued when passed to React components ?

I don't know banks very well, but I suspect the queries are run once high up in the react tree and the same query is never re-run. In drive/contact/photos we mostly use <Query /> components without `as`, so that avoids the problem. I don't think documents are uniqued later on either, I think it's just that we never ran into this use case before.

I added a unit test demonstrating the problem… so the PR is now ready for review!

PR #669, opened 2020-03-11
feat: Support cursor on FilesCollection.find

y-lohse:
This PR adds support for using bookmark based pagination on FileCollection.find(). 
The stack API is a bit strange on this route, it returns a link containing a `cursor` param which is technically a bookmark, and we need to call that route with the bookmark named `bookmark` 😁.  The inconsistencies start to be addressed [here](https://github.com/cozy/cozy-stack/pull/2427) but in the meantime we can abstract the difference away.

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -1,5 +1,6 @@
 import mime from 'mime/lite'
 import has from 'lodash/has'
+import get from 'lodash/get'
 import DocumentCollection, { normalizeDoc } from './DocumentCollection'
 import { uri, slugify, forceFileDownload, formatBytes } from './utils'
 import * as querystring from './querystring'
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -69,22 +70,29 @@ class FileCollection extends DocumentCollection {
...
     const resp = await this.stackClient.fetchJSON(
       'POST',
       '/files/_find',
       await this.toMangoOptions(selector, options)
     )
+
+    const nextLink = get(resp, 'links.next')
+    const nextBookmark = nextLink
+      ? nextLink.match(/\[cursor\]=(.+)/)[1]
+      : undefined
```
edas:
```suggestion
    const nextBookmark = nextLink && nextLink.match(/\[cursor\]=(.+)/)[1]
```

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -69,22 +70,29 @@ class FileCollection extends DocumentCollection {
...
     return {
       data: resp.data.map(f => normalizeFile(f)),
       meta: resp.meta,
       next: resp.meta.count > skip + resp.data.length,
-      skip
+      skip,
+      bookmark: nextBookmark
```
edas:
```suggestion
      bookmark: nextBookmark || undefined
```

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -69,22 +70,29 @@ class FileCollection extends DocumentCollection {
...
+
     const resp = await this.stackClient.fetchJSON(
       'POST',
       '/files/_find',
       await this.toMangoOptions(selector, options)
     )
+
+    const nextLink = get(resp, 'links.next')
+    const nextBookmark = nextLink
+      ? nextLink.match(/\[cursor\]=(.+)/)[1]
```
edas: Should we really use a regex and not a SearchParam ?

PR #670, opened 2020-03-13
fix: Use bookmark with fetchMore

Gara64:
This allows to use bookmark through `fetchMore` when available. 
It also fixes a logic issue that could cause a wrong dispatch of `initQuery` during pagination.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -576,8 +576,7 @@ client.query(Q('io.cozy.bills'))`)
     // Don't trigger the INIT_QUERY for fetchMore() calls
     if (
       existingQuery.fetchStatus !== 'loaded' ||
-      !queryDefinition.skip ||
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -576,8 +576,7 @@ client.query(Q('io.cozy.bills'))`)
     // Don't trigger the INIT_QUERY for fetchMore() calls
     if (
       existingQuery.fetchStatus !== 'loaded' ||
-      !queryDefinition.skip ||
-      !queryDefinition.bookmark
+      (!queryDefinition.skip && !queryDefinition.bookmark)
```
Crash--: can we have a test for this? ;)

Gara64: Done with https://github.com/cozy/cozy-client/pull/670/commits/ffd4843a75e79bdb386d3d48ac942e6b31fdd007 ;)

PR #675, opened 2020-03-23
feat: Progress events for upload

ptbrowne:
Since fetch does not support upload progress event, we have to use
XMLHttpRequest

### DIFF
--- a/packages/cozy-stack-client/src/xhrFetch.js
+++ b/packages/cozy-stack-client/src/xhrFetch.js
@@ -1,0 +1,63 @@
+import memoize from 'lodash/memoize'
+
+const headersFromString = headerString => {
+  return new Headers(
+    headerString
+      .split('\r\n')
+      .map(x => x.split(':', 2))
+      .filter(x => x.length == 2)
+  )
+}
+
+/**
+ * Returns a `fetch()` like response but uses XHR.
+ * XMLHTTPRequest provides upload progress events unlike fetch.
+ *
+ * @private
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/xhrFetch.js
```diff
@@ -0,0 +1,55 @@
...
+  return {
+    headers: headersFromString(response.getAllResponseHeaders()),
+    ok: response.status >= 200 && response.status < 300,
+    text: async () => response.responseText,
+    json: async () => JSON.parse(response.responseText)
+  }
+}
+
+const shouldXMLHTTPRequestBeUsed = (method, path, options) => {
+  return Boolean(options.onUploadProgress)
```
Crash--: It seems that there is a few interrogation marks on for the iOS supports here : https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest

We should maybe be more strict than just check if there is this option?

Crash--: we use `xhr.onload`. And we have a "?" for safari ios support on "load event". Same thing for `progess`. 

> You mean I should add a user agent check ?

Nop, adding a test like : 
if(xhr.onload !== undefined && xhr.onprogress !== undefined)

to test if the method we used are available

ptbrowne: I added a navigator user agent check, and made the event binding more robust (by checking if xhr.upload exists).

PR #676, opened 2020-03-27
feat: Handle client errors

edas:
Add a React `<ClientErrors />` component that tracks
cozy-client's errors and trigger the correct behavior in
the display. This first step only deals with quota errors.


packages/cozy-client/src/ClientErrors.jsx
```diff
@@ -0,0 +1,104 @@
...
+ */
+function renderErrors(errorStack, setErrorStack) {
+  return errorStack.map((error, key) => {
+    if (error instanceof Response) {
+      const Component = byHttpStatus[error.status]
+      const dismiss = () => setErrorStack(stack => pull(stack, error))
+      return <Component key={key} error={error} dismiss={dismiss} />
+    } else {
+      return null
+    }
```
ptbrowne: What do you think of re-arranging a bit :

```
const getComponentForError = error => {
	if (error instanceof Response) {
	  return byHttpStatus[error.status] || null
	}
	return null     
}

function renderErrors
   ....
    const Component = getComponentForError(error)
    if (Component) {
      const dismiss = () => setErrorStack(stack => pull(stack, error))
      return <Component key={key} error={error} dismiss={dismiss} />
    } else {
      return null
    }
}
```

packages/cozy-client/src/ClientErrors.jsx
```diff
@@ -0,0 +1,104 @@
...
+   */
+  const handleError = useCallback(
+    error => {
+      // `isErrorManaged` is there to avoid the same error to be added twice
+      // once the error has been added once, the `isErrorManaged`is set to true
+      // and any future push is ignored
+      if (!error.isErrorManaged) {
+        if (error instanceof Response) {
+          if (byHttpStatus[error.status]) {
+            error.isErrorManaged = true
```
ptbrowne: The error instanceof Response and byHttpStatus[error.status] could be refactored into a `getErrorIdentifier` so that support for other errors can be added easily.

I am wondering if the getComponentFromError and getIdentifier could be merged into a single function that returns { identifier, component }. This way, no drift can appear between the two functions.

packages/cozy-client/src/ClientErrors.jsx
```diff
@@ -0,0 +1,104 @@
...
+   *
+   * @param {Error} error -
+   * @returns {boolean}
+   */
+  const handleError = useCallback(
+    error => {
+      // `isErrorManaged` is there to avoid the same error to be added twice
+      // once the error has been added once, the `isErrorManaged`is set to true
+      // and any future push is ignored
+      if (!error.isErrorManaged) {
```
ptbrowne: Can we switch the condition to use early bailout ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -94,13 +94,15 @@ class CozyClient {
     this.handleTokenRefresh = this.handleTokenRefresh.bind(this)
 
     this.createClient()
+    const stackClient = this.getStackClient()
+    stackClient.on('error', (...args) => this.emit('error', ...args))
```
ptbrowne: I see no `off()` for this, is it expected ?

edas: We have no public method to replace or remove the stack client. It should be live as long as the main client is live.

packages/cozy-client/src/ClientErrors.jsx
```diff
@@ -0,0 +1,118 @@
...
+    if (handleExceptions) {
+      client.on('error', handleError)
+      return () => client.removeListener('error', handleError)
+    }
+  }, [client, handleError, handleExceptions])
+
+  return (
+    <>
+      {children}
+      {renderErrors(errorStack, setErrorStack)}
```
ptbrowne: I am wondering if we could change the component not be a Wrapper component = that it should not render its children, I am a bit uneasy with having a deep React tree, and am not sure it is necessary here since we do not use a Provider.

packages/cozy-client/src/ClientErrors.jsx
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Rendering functions for client fetch errors by status code
+ *
+ * @private
+ * @type {object}
+ */
+const byHttpStatus = {
+  413: QuotaError
+}
```
Crash--: I'm wondering if there is a way to make that pass the "grep test Component". I used to search for `<ComponentName` and not only `ComponentName` to find where the component is used.

packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -226,6 +228,15 @@ describe('Stack client initialization', () => {
       client.handleTokenRefresh
     )
   })
+
+  it('receive errors from stackClient', done => {
```
ptbrowne:
```suggestion
  it('proxies errors from stackClient', done => {
```

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+    client.on.mockImplementation(name => name === 'error' && done())
+    renderWrappedHook(undefined, client)
+  })
+
+  describe('renderErrors', () => {
+    it('returns a function', () => {
+      const { result } = renderWrappedHook()
+      const { ClientErrors } = result.current
+      expect(ClientErrors).toBeInstanceOf(Function)
+    })
```
ptbrowne: I am not sure this is necessary to test for this. We expect ClientErrors to be either a function or a React.Component derived class. This would be tested by the following tests when ClientErrors is mounted.

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function createWrapper(client = createCozyClient()) {
+  function Wrapper({ children }) {
+    return <CozyProvider client={client}>{children}</CozyProvider>
+  }
+  return Wrapper
+}
+
+function renderWrappedHook(params, client) {
```
ptbrowne: Params is not used.

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+function wrappedShallow(tree, client) {
+  return shallow(tree, { wrappingComponent: createWrapper(client) })
+}
+
+describe('useClientErrors', () => {
+  it('registers an `error` handler in client', done => {
+    const client = createCozyClient()
+    client.on.mockImplementation(name => name === 'error' && done())
+    renderWrappedHook(undefined, client)
+  })
```
ptbrowne: This seems like testing for an implementation detail.

What is important here is that ClientErrors contains errors if cozy-client has an error, the mechanism through which this happens should not be tested IMHO.

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+    it('displays nothing by default', () => {
+      const { result } = renderWrappedHook()
+      const { ClientErrors } = result.current
+      const node = wrappedShallow(<ClientErrors />)
+      expect(node).toHaveLength(0)
+    })
+
+    describe('for quota errors', () => {
+      it('displays a modal', async () => {
```
ptbrowne:
```suggestion
      it('displays a QuotaAlert', async () => {
```

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+        )
+        act(() => {
+          const handler = client.on.mock.calls[0][1]
+          handler(error)
+        })
+
+        await nextUpdate
+        const { ClientErrors } = result.current
+        const node = wrappedShallow(<ClientErrors />, client)
+        expect(node).toHaveLength(1)
```
ptbrowne: Could we test that QuotaAlert is displayed ?

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+          status: 413,
+          statusText: 'Quota exceeded'
+        })
+        const error = new FetchError(
+          response,
+          `${response.status} ${response.statusText}`
+        )
+        act(() => {
+          const handler = client.on.mock.calls[0][1]
+          handler(error)
```
ptbrowne: I am not sure actually that we should have a fake cozy-client here. With a real client, we could do client.emit('error', error) instead of using `.mock.calls[0][1]`.

packages/cozy-stack-client/src/__tests__/CozyStackClient.spec.js
```diff
@@ -223,6 +223,16 @@ describe('CozyStackClient', () => {
...
     })
+
+    it('should emit on error', async () => {
+      fetch.mockReturnValueOnce(
+        new Response(null, { status: 413, statusText: 'Quota exceeded' })
+      )
+      const handler = jest.fn()
+      client.on('error', handler)
+      await client.fetch('GET', '/data/anyroute')
+      expect(handler).toHaveBeenCalled()
```
ptbrowne: I think we could have a `toHaveBeenCalledWith` here to be a bit more precise.

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
edas: I didn't managed to. The component from UI doesn't keep its displayname. I could `mount` and check if its a modal, but not the actual name of the component

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
edas: that would test the EventEmitter of Cozy Client. Here I only look for input and outputs of the hook.

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
edas: Hum… may be related to https://github.com/cozy/cozy-client/pull/676#discussion_r400305385

I see the cozyClient as an injectable dependency, with an external public interface, not some internal behaviour of the hook.

packages/cozy-client/src/ClientErrors.jsx
edas: Is the comment ok for you ?

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
ptbrowne: expect(mount(<ClientErrors />).find(QuotaAlert).length).toBe(1)

should work no ?

edas: nope, QuotaAlert is seen as "Component". Let's see that tomorow morning

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
ptbrowne: AMHA here it makes more sense that the tests exercise the the hook under test with what will happen in real conditions.

packages/cozy-client/src/hooks/useClientErrors.spec.jsx
ptbrowne: I guess since QuotaAlert is declared with `withClient`, the JSX transpiler cannot infer its name, I think by postponing withClient to the `export default` site, QuotaAlert will have the right name.

edas:
https://github.com/cozy/cozy-ui/pull/1420

packages/cozy-client/src/hooks/useClientErrors.jsx
```diff
@@ -0,0 +1,118 @@
...
+ * @see ClientErrors
+ * @param {Error[]} errorStack - array of errors/exceptions
+ * @param {Function} setErrorStack - mutates the array of errors
+ * @returns {Function} React rendering
+ */
+function renderErrors(errorStack, setErrorStack) {
+  const errors = errorStack.map((error, key) => {
+    const Component = getErrorComponent(error)
+    if (Component) {
+      const dismiss = () => setErrorStack(stack => [...pull(stack, error)])
```
ptbrowne: Nit: I think remove(stack, error) would be clearer.

PR #686, opened 2020-04-17
Add useFetchJSON

ptbrowne:
Ability to use client's fetch json.

- [x] tests
- [x] docs
- [x] same arguments as fetchJSON + dependencies


ptbrowne:
> Passing a dependencies for useEffect is a bit unexpected, 

* Re-executing the request if you know that something could have changed.
* Re-executing a request every tens of seconds.

Since this is eventually some sugar for an effect, it makes to me to be able to re-execute the effect.

Otherwise you can only run the effect at mount time which seems overly restrictive.

PR #695, opened 2020-05-11
feat: Connection to multiple queries with less nesting

ptbrowne:
### Problem

queryConnect entails one nesting level for each query, which leads to a lot of nesting in Banks.

Example:

```jsx
const Component = () => {}

export default queryConnect({
  transactions: { ... }, // query options for transactions
  triggers: { ... }, // query options for triggers
  accounts: { ... }, // query options for accounts
  groups: { ... }, // query options for groups
})
```

When rendered, it will look like:

```jsx
<Query query={ transactionsQuery }>
    <Query query={ accountsQuery }>
        <Query query={ groupsQuery }>
            <Query query={ triggersQuery }>
            </Query>
        </Query>
    </Query>
</Query>
```

This makes it hard to debug in Banks and adds noise.

### Solution

Add a new `useQueries` based on `useQuery` that returns an object
will all the responses.

queryConnect2 uses this hook to provide thoses responses as props
(exactly like queryConnect) but only with 1 nesting level.

### DIFF
--- a/packages/cozy-client/src/hoc.spec.jsx
+++ b/packages/cozy-client/src/hoc.spec.jsx
@@ -44,17 +44,21 @@ describe('with client', () => {
   })
 })
 
+const queryConns = {
+  simpsons: { query: client => Q('io.cozy.simpsons'), as: 'simpsons' },
+  upperSimpsons: {
+    query: client => Q('io.cozy.simpsons-upper'),
+    as: 'upperSimpsons'
+  }
+}
+
 describe('queryConnect', () => {
   const setup = () => {
     const client = setupClient()
     const WithQueries = queryConnect({
-      simpsons: { query: client => Q('io.cozy.simpsons'), as: 'simpsons' },
-      upperSimpsons: {
-        query: client => Q('io.cozy.simpsons-upper'),
-        as: 'upperSimpsons'
-      }
+      simpsons: queryConns.simpsons,
+      upperSimpsons: queryConns.upperSimpsons
     })(Component)
-
     return { WithQueries, client }
   }
 
--- a/packages/cozy-client/src/hoc.spec.jsx
+++ b/packages/cozy-client/src/hoc.spec.jsx
@@ -128,4 +128,18 @@ describe('queryConnectFlat', () => {
         .data.map(x => x.name)
     ).toEqual(['HOMER', 'MARGE'])
   })
+
+  it('should have Component as direct children of queryConnectFlag', () => {
+    const { WithQueries, client } = setup()
+    const uut = mount(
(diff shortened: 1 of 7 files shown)


Crash--:
> queryConnect has a problem, it entails one nesting level for each
query, which leads to a lot of nesting in Banks. 

Can you explain a bit more what you mean by a lot of nesting? What kind of nesting are we talking about? Can you add an exemple to get an idea of what you're trying to solve? 

I don't know if queryConnect2 is a good name, but I'm waiting to have a better understanding before giving a few suggestions.

ptbrowne:
- I added an example hopefully clarifying what I meant by nesting.

- As for the name queryConnect2 : it does the same as queryConnect without the nesting. *But*, I do not want to replace queryConnect as of yet for fear of breaking stuff. I will add tests but since both use different implementations (one based on hooks, the other based on redux connect), I am not sure we should replace one by the other as of yet.

y-lohse:
> But, I do not want to replace queryConnect as of yet for fear of breaking stuff.

That makes sense, no problem for me. But `queryConnect2` is a bit cryptic, maybe `queryConnectFlat`?

Implementation LGTM!

ptbrowne:
Thanks, I changed the name to `queryConnectFlat`.

packages/cozy-client/src/hoc.spec.jsx
```diff
@@ -82,7 +76,56 @@ describe('queryConnect', () => {
...
+    })(Component)
+    return { WithQueries, client }
+  }
+
+  it('should give a display name', () => {
+    const { WithQueries } = setup()
+    expect(WithQueries.displayName).toBe('queryConnectFlat(Component)')
+  })
+
+  it('should pass result of query definitions as props', () => {
```
y-lohse, line 109: Wouldn't this be true for `queryConnect` as well? would it be possible to explore the render tree to make sure there isn't more than one `<Query />` component?

ptbrowne: > Wouldn't this be true for queryConnect as well?

Yes this is true for queryConnect as well.

> would it be possible to explore the render tree to make sure there isn't more than one <Query /> component?

Yes I can write another test to make sure of that.

ptbrowne: To be clear, there will not be any `Query` component there, the data pulling from the store is done with a hook.

ptbrowne: https://github.com/cozy/cozy-client/pull/695/commits/e600e819ff31b1c31744cbf44f32f81086514ba6 :)

PR #697, opened 2020-05-11
Draft: useDocument

ptbrowne:
Hook to retrieve a document from the store.

This is in draft as I realized that this implementation does not observe changes in the store.


ptbrowne:
Closing this as I will not be able to work on it.

Issue #690, opened 2020-04-24
FileUpload Date header


Crash--:
We can now use xhr to upload file in order to get process callback. But it comes with a drawback... For security reason, Chromium based browser, don't let us set the Date header for a file. (https://bugs.chromium.org/p/chromium/issues/detail?id=38382 & https://stackoverflow.com/questions/6973403/chrome-extension-refuse-header-date-in-ajax-request/6974039 ) 

We use it at two places: 
https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/FileCollection.js#L543

https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/FileCollection.js#L753

We should add a QueryString updated_at & created_at in order to avoid using this Date header. 
It needs a modification of cozy-stack to accept this new updated_at QS.

https://github.com/cozy/cozy-stack/blob/master/web/files/files.go#L153-L163

ptbrowne:
Would you be up to take this issue ?

PR #699, opened 2020-05-12
WIP: Use UpdatedAt params instead of Date Header

Crash--:
- Need to look after other places where Date is used 
- Need to write tests

See https://github.com/cozy/cozy-client/issues/690 (stack pr https://github.com/cozy/cozy-stack/pull/2506 )


trollepierre:
fixed by https://github.com/cozy/cozy-client/pull/1114

PR #701, opened 2020-05-15
feat: Expose permission.fetchOwn

y-lohse:
`fetchOwn` was only used in `findPermissionFor`, but it is generally useful to load and parse our own permissions. In https://github.com/cozy/cozy-drive/pull/2017 for example, we don't want to know the permissions for a certain document, we need them for the general permission set.

### DIFF
--- a/packages/cozy-client/src/models/permission.js
+++ b/packages/cozy-client/src/models/permission.js
@@ -40,13 +40,12 @@ export function isReadOnly(perm, options = {}) {
 }
 
 /**
- * Get the list of permissions blocks
+ * Fetches the list of permissions blocks
  *
- * @private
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/models/permission.js
```diff
@@ -46,7 +46,7 @@ export function isReadOnly(perm, options = {}) {
  * @param {CozyClient} client -
  * @returns {PermissionItem[]} list of permissions
  */
-async function fetchOwn(client) {
+export async function fetchOwn(client) {
```
ptbrowne, line 49: Could you remove the @private mention in the docstring ?
Also, could you replace Get by Fetch in the docstring ?

PR #709, opened 2020-06-02
fix: Correctly call callback for tokenRefresh passed in options

ptbrowne:
CozyClient passes the onTokenRefresh callback with options.onTokenRefresh.

This is not a breaking change since application that do set `cozyStackClient.onTokenRefresh` will still work after this change (since we call `this.onTokenRefresh` when receiving the token).

### DIFF
--- a/packages/cozy-stack-client/src/CozyStackClient.js
+++ b/packages/cozy-stack-client/src/CozyStackClient.js
@@ -133,6 +133,12 @@ class CozyStackClient {
     }
   }
 
+  onTokenRefresh(token) {
+    if (this.options && this.options.onTokenRefresh) {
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -133,6 +133,12 @@ class CozyStackClient {
     }
   }
 
+  onTokenRefresh(token) {
```
Crash--, line 136: Should we add the docstring?

ptbrowne: It's supposed to be private so I am OK not adding a docstring here.

PR #713, opened 2020-06-19
File creation + keep query sorted

ptbrowne:
- File creation via client.create('io.cozy.files', { ... })
- Query keep their sort order if it has been configured at
  query definition

### DIFF
--- a/packages/cozy-client/src/queries/dsl.js
+++ b/packages/cozy-client/src/queries/dsl.js
@@ -282,6 +282,8 @@ export const uploadFile = (file, dirPath) => ({
  *
  * Used to sort query results inside the store when creating a file or
  * receiving updates.
+ *
+ * @private
  */
 export const makeSorterFromDefinition = definition => {
   const sort = definition.sort
--- a/packages/cozy-client/src/queries/dsl.js
+++ b/packages/cozy-client/src/queries/dsl.js
@@ -1,4 +1,5 @@
-const isString = require('lodash/isString')
+const isArray = require('lodash/isArray')
+const orderBy = require('lodash/orderBy')
 
 /**
  * typedef QueryDefinition
@@ -100,9 +101,9 @@ class QueryDefinition {
    * @returns {QueryDefinition}  The QueryDefinition object.
    */
   sortBy(sort) {
-    if (isString(sort)) {
+    if (!isArray(sort)) {
       throw new Error(
-        'Invalid sort, should be an object (`{ label: "desc"}`), you passed a string.'
+        `Invalid sort, should be an array ([{ label: "desc"}, { name: "asc"}]), you passed ${JSON.stringify(sort)}.`
       )
     }
     return new QueryDefinition({ ...this.toDefinition(), sort })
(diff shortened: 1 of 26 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -277,6 +277,19 @@ class FileCollection extends DocumentCollection {
     return this.createFile(data, { dirId })
   }
 
+  /**
+   * Creates for both directory or file.
+   * - Used by StackLink to support CozyClient.create('io.cozy.files', options)
+   */
+  async create(attributes) {
```
Crash--, line 286: Can you add the description of `attributes` ? Also, should we mark `createDirectory` and `createFile` private?

packages/cozy-client/src/queries/dsl.js
```diff
@@ -100,9 +101,9 @@ class QueryDefinition {
    * @returns {QueryDefinition}  The QueryDefinition object.
    */
   sortBy(sort) {
-    if (isString(sort)) {
+    if (!isArray(sort)) {
       throw new Error(
-        'Invalid sort, should be an object (`{ label: "desc"}`), you passed a string.'
+        `Invalid sort, should be an array ([{ label: "desc"}, { name: "asc"}]), you passed ${JSON.stringify(sort)}.`
       )
     }
```
Crash--, line 110: Is it a breaking change? Can't we manage both ? Object and array ?

packages/cozy-client/src/queries/dsl.js
```diff
@@ -274,6 +275,35 @@ export const uploadFile = (file, dirPath) => ({
...
 })
 
+/**
+ * Creates a sort function from a definition.
+ *
+ * Used to sort query results inside the store when creating a file or
+ * receiving updates.
+ */
+
+export const makeSorterFromDefinition = definition => {
```
Crash--, line 287: can we have a docstring for the definition ?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -302,15 +302,18 @@ class FileCollection extends DocumentCollection {
    */
   async createFile(
     data,
-    { name, dirId = '', executable, metadata, ...options } = {}
+    { name, dirId, executable, metadata, ...options } = {}
```
Crash--, line 307: I agree that this is confusing that sometime we pass a dirId and sometime not. But I'm not confortable with this kind of BC without previous notice. We should warn the dev, and set the default value to `''` if needed. And in a near future remove this warning and really break the stuff.

packages/cozy-stack-client/src/errors.js
```diff
@@ -7,3 +7,24 @@ export default {
   CLIENT_NOT_FOUND,
   INVALID_TOKEN
 }
+
+export class FetchError extends Error {
```
Crash--, line 11: it's more a `refactor` than a `chore`

packages/cozy-stack-client/src/FileCollection.js
ptbrowne: Done, and done.

packages/cozy-client/src/queries/dsl.js
ptbrowne: No this is not a BC, just a documentation change.

packages/cozy-client/src/queries/dsl.js
ptbrowne: The docstring is just above, but was separated from the function with an empty line, I removed it.

packages/cozy-client/src/queries/dsl.js
Crash--: sortBy({name: 'asc'}) => Now returns an Error when before no.

Crash--: OK just read the docstring, and we're already expecting an array `   * @param {Array} sort The list of field name and direction pairs.`

packages/cozy-stack-client/src/FileCollection.js
ptbrowne: Yes, I should have checked the doc, I was not sure that without folderId this would work but it works.

https://docs.cozy.io/en/cozy-stack/files/#post-filesdir-id

>  When it’s not given, the directory is created at the root of the virtual file system.

I removed the commit.

ptbrowne:
You're right, it seems better to put it in queries. I've moved it in https://github.com/cozy/cozy-client/pull/713/commits/d4c5f2ba30ebb86f3930433efbc8ea9801538724, thanks.

PR #717, opened 2020-06-26
fix: Support multiple ids in store query updates

y-lohse:
When receiving a query result, queries using `getByIds` would always be affected by new results, even if they had different ids.

### DIFF
--- a/packages/cozy-client/src/store/queries.spec.js
+++ b/packages/cozy-client/src/store/queries.spec.js
@@ -89,6 +89,25 @@ describe('queries reducer', () => {
       expect(state).toMatchSnapshot()
     })
 
+    it('should correctly update a query with several ids', () => {
+      const query = new QueryDefinition({
+        doctype: 'io.cozy.todos'
+      })
+      applyAction(initQuery('b', query.getByIds([TODO_1._id, TODO_2._id])))
+      applyAction(
+        receiveQueryResult('b', {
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/store/queries.spec.js
```diff
@@ -89,6 +89,25 @@ describe('queries reducer', () => {
...
+          data: [TODO_1, TODO_2]
+        })
+      )
+      expect(state.b.data).toEqual([TODO_1._id, TODO_2._id])
+      applyAction(
+        receiveQueryResult('a', {
+          data: [TODO_3]
+        })
+      )
+      expect(state.b.data).toEqual([TODO_1._id, TODO_2._id])
```
ptbrowne, line 108: I am not following why this test would not work before the change ? What was state.b.data equal to before the change ?

y-lohse: It was `[TODO_1._id, TODO_2._id, TODO_3._id]`, because we  didn't return a `selectorFilterFn` so [this code](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/queries.js#L141-L147) returned true no matter the document in the query result.

ptbrowne: Ah, I think that TODO_3 would be part of b.data.

Nit: I think a comment would be a little helpful.

PR #718, opened 2020-06-26
fix: Client-side query sort is case insensitive

y-lohse:
`lodash.sortBy` is case-sensitive when ordering strings, whereas the stack is not. This PR removes the case-sensitivity from our sort.


packages/cozy-client/src/store/queries.js
```diff
@@ -166,7 +172,10 @@ export const makeSorterFromDefinition = definition => {
     return docs => docs
   } else {
     const attributeOrders = sort.map(x => Object.entries(x)[0])
-    const attrs = attributeOrders.map(x => x[0])
+
```
ptbrowne, line 175: Nit: Line can be removed I think

y-lohse:
Yeah, we ended up with folders named `aloha` being displayed after  folders named `Zebra` for example.

PR #728, opened 2020-07-03
fix: Use query support fetchMore without bookmark + fetchMore return

y-lohse:
- fetchMore returning a promise is in line with ObservableQuery.fetchMore
- queries don't always have a bookmark to fetch more data, especially when coming from pouch

### DIFF
--- a/packages/cozy-client/src/hooks/useQuery.js
+++ b/packages/cozy-client/src/hooks/useQuery.js
@@ -7,8 +7,10 @@ const resolveToValue = fnOrValue => {
   return typeof resolveToValue === 'function' ? fnOrValue() : fnOrValue
 }
 
-const generateFetchMoreQueryDefinition = collection => {
-  return collection.definition.offsetBookmark(collection.bookmark)
+const generateFetchMoreQueryDefinition = queryResult => {
+  return queryResult.bookmark
+    ? queryResult.definition.offsetBookmark(queryResult.bookmark)
+    : queryResult.definition.offset(queryResult.data.length)
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/cozy-client/src/hooks/useQuery.spec.jsx
```diff
@@ -43,6 +44,55 @@ describe('use query', () => {
...
+    } = setupQuery({
+      queryOptions: {
+        as: 'simpsonsBookmarked'
+      },
+      queryDefinition: () => queryDefinition,
+      storeQueries: {
+        simpsonsBookmarked: {
+          data: simpsonsFixture,
+          doctype: 'io.cozy.simpsons',
+          definition: queryDefinition
```
ptbrowne, line 65: There is no bookmark here ? It seemed to me that we'd use off with bookmark only if the query result contained a bookmark ?

y-lohse: Yep, there is a book mark in the definition on l. 50!

packages/cozy-client/src/hooks/useQuery.js
```diff
@@ -7,8 +7,10 @@ const resolveToValue = fnOrValue => {
   return typeof resolveToValue === 'function' ? fnOrValue() : fnOrValue
 }
 
-const generateFetchMoreQueryDefinition = collection => {
-  return collection.definition.offsetBookmark(collection.bookmark)
+const generateFetchMoreQueryDefinition = query => {
+  return query.bookmark
+    ? query.definition.offsetBookmark(query.bookmark)
+    : query.definition.offset(query.data.length)
```
ptbrowne, line 13: s/query/queryResult would make it a bit more clear, the naming is always a bit difficult between query result and queries.

packages/cozy-client/src/hooks/useQuery.spec.jsx
y-lohse: After investigating, we found it's actually using `skip` in this test too, but *also* the bookmark from the definition. I'm going to patch up the test.

PR #729, opened 2020-07-03
PouchLink improvements

y-lohse:
PouchLink can now be configured to replicate certain doctypes only in one direction, which is useful for `io.cozy.files` for example that can't be replicated to the stack, but we'd still like to have the content in the local database. 

The API is debatable; fo the main use case (bi-directional replication), I think the current API is great (`new PouchLink({
  doctypes: ['io.cozy.whatever'] })`) and doesn't need to move. I added support for a second, more advanced format, which is `new PouchLink({
  doctypes: { readWrite: ['io.cozy.whatever'], readOnly: ['io.cozy.files'] } })`. It's flexible, duplication free and extandable with a `writeOnly` if we ever need that.

This PR also adds detection for next pages when results are paginated with `skip` and `limit`, which is how the `pouch.find` API uses.

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -276,7 +276,16 @@ class PouchLink extends CozyLink {
     }
   }
 
-  async executeQuery({ doctype, selector, sort, fields, limit, id, ids }) {
+  async executeQuery({
+    doctype,
+    selector,
+    sort,
+    fields,
+    limit,
+    id,
+    ids,
+    skip
+  }) {
     const db = this.getPouch(doctype)
     let res, withRows
     if (id) {
@@ -295,10 +304,13 @@ class PouchLink extends CozyLink {
         sort,
         selector,
         fields,
-        limit
+        limit,
+        skip
       }
       await this.ensureIndex(doctype, findOpts)
       res = await find(db, findOpts)
+      res.offset = skip
+      res.limit = limit
       withRows = true
     }
     return jsonapi.fromPouchResult(res, withRows, doctype)
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -49,10 +49,21 @@ const normalizeAll = (docs, doctype) => {
  * to respond to queries and mutations.
  */
 class PouchLink extends CozyLink {
+  /**
+   * constructor - Initializes a new PouchLink
+   *
+   * @param {object} [opts={}]
+   * @param {number} [opts.replicationInterval] Milliseconds between replications
+   * @param {string[]} opts.doctypes Doctypes to replicate
+   * @param {object[]} opts.doctypesReplicationOptions A mapping from doctypes to replication options. All pouch replication options can be used, as well as the "strategy" option that determines which way the replication is done (can be "sync", "fromRemote" or "toRemote")
+   *
+   * @returns {object} The PouchLink instance
+   */
+
   constructor(opts = {}) {
     const options = defaults({}, opts, DEFAULT_OPTIONS)
     super(options)
-    const { doctypes } = options
+    const { doctypes, doctypesReplicationOptions } = options
     this.options = options
     if (!doctypes) {
       throw new Error(
@@ -60,6 +71,7 @@ class PouchLink extends CozyLink {
       )
     }
     this.doctypes = doctypes
+    this.doctypesReplicationOptions = doctypesReplicationOptions
     this.indexes = {}
   }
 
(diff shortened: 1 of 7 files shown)


Crash--:
which is useful for io.cozy.doctypes > io.cozy.files

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -31,14 +31,21 @@ const humanTimeDelta = timeMs => {
 }
 
 /* Create a cancellable promise for replication with default options */
-const startReplication = (pouch, getReplicationURL) => {
+const startReplication = (
```
Crash--, line 34: can we have a docstring here ?

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -31,14 +31,21 @@ const humanTimeDelta = timeMs => {
...
+  pouch,
+  replicateFromRemoteOnly,
```
Crash--, line 36: I'm not a big fan of rewording the concept. One time we use : readOnly, and other ReplicateFromRemoteOnly. Can't we use one for both?

packages/cozy-pouch-link/src/PouchManager.spec.js
```diff
@@ -69,13 +69,28 @@ describe('PouchManager', () => {
...
 
-  it('should periodically call replicate', async () => {
+  it('should periodically call sync', async () => {
     manager.startReplicationLoop()
     await sleep(1000)
     const pouch = manager.getPouch('io.cozy.todos')
     expect(pouch.sync.mock.calls.length).toBeGreaterThan(5)
   })
 
+  it('should only replicate in one direction for some doctypes', async () => {
```
Crash--, line 79:
```suggestion
  it('should only replicate from the remote for some doctypes', async () => {
```

packages/cozy-pouch-link/src/PouchManager.js
y-lohse: Fair, i've renamed the variables to `ReplicateFromRemoteOnly`. However for the public API i'd rather keep `readOnly`/`readWrite` since I think `ReplicateFromRemoteOnly`/Replicate...BothWays?`is too wordy and the distinction between bidirectional and the others is hard to express.

Crash--: 👍

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -30,15 +30,30 @@ const humanTimeDelta = timeMs => {
...
 
-/* Create a cancellable promise for replication with default options */
-const startReplication = (pouch, getReplicationURL) => {
+/**
+ * startReplication - Create a cancellable promise for replication with default options
+ *
+ * @param {object} pouch                 Pouch database instance
+ * @param {boolean} replicateFromRemoteOnly Whether the replication should come from the remote only or be bidirectional
+ * @param {Function} getReplicationURL A function that should return the remote replication URL
+ *
```
ptbrowne, line 39: I think we should add @private here.

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -30,15 +30,30 @@ const humanTimeDelta = timeMs => {
...
   const promise = new Promise((resolve, reject) => {
     const url = getReplicationURL()
-    replication = pouch.sync(url, {
-      batch_size: 1000 // we have mostly small documents
-    })
+    replication = replicateFromRemoteOnly
+      ? pouch.replicate.from(url)
+      : pouch.sync(url, {
+          batch_size: 1000 // we have mostly small documents
+        })
```
ptbrowne, line 55: batchSize options should be passed to replicate.from .

y-lohse:
After a chat with @ptbrowne I reworked the PR a little bit:

- The API has changed and I added a blip about it in the documentation
- CouchLink doesn't touch the replication options at all, it passes them to CouchManager
- CouchManager now supports all 3 types of replication, as well as receiving any other replication option

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -49,17 +49,29 @@ const normalizeAll = (docs, doctype) => {
  * to respond to queries and mutations.
  */
 class PouchLink extends CozyLink {
+  /**
+   * constructor - Initializes a new PouchLink
+   *
+   * @param {object} [opts={}]
+   * @param {number} [opts.replicationInterval] Milliseconds between replications
+   * @param {(object)} [opts.doctypes] List of doctypes to replicate
```
ptbrowne, line 57:
```suggestion
   * @param {Array<string>} [opts.doctypes] List of doctypes to replicate
```

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -49,17 +49,29 @@ const normalizeAll = (docs, doctype) => {
...
+   * @param {(object)} [opts.doctypesReplicationOptions] A dictionnary of doctypes and associated configuration options. All pouch replication options can be used, as well as the "strategy" option that determines which way the replication is done (can be "sync", "fromRemote" or "toRemote")
```
ptbrowne, line 58:
```suggestion
   * @param {object[]} opts.doctypesReplicationOptions A mapping from doctypes to replication options. All pouch replication options can be used, as well as the "strategy" option that determines which way the replication is done (can be "sync", "fromRemote" or "toRemote")
```

PR #730, opened 2020-07-07
Aditionnal pouch fixes

y-lohse:
Two more little couch nits that end up causing errors. Commits & tests describe what's going on!

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -307,6 +307,7 @@ class PouchLink extends CozyLink {
     } else if (ids) {
       res = await allDocs(db, { include_docs: true, keys: ids })
       res = withoutDesignDocuments(res)
+      res.total_rows = null // pouch indicates the total number of docs in res.total_rows, even though we use "keys". Setting it to null avoids cozy-client thinking there are more docs to fetch.
       withRows = true
     } else if (!selector && !fields && !sort) {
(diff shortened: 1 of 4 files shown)


packages/cozy-pouch-link/src/jsonapi.js
```diff
@@ -20,7 +20,9 @@ export const normalizeDoc = (doc, doctype) => {
 
 export const fromPouchResult = (res, withRows, doctype) => {
   if (withRows) {
-    const docs = res.rows ? res.rows.map(row => row.doc) : res.docs
+    const docs = res.rows
+      ? res.rows.map(row => row.doc).filter(Boolean)
```
Crash--, line 24: I'm wondering if we will remember what this line is in a few months. Maybe we can create a method with an explicit name?

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -307,6 +307,7 @@ class PouchLink extends CozyLink {
     } else if (ids) {
       res = await allDocs(db, { include_docs: true, keys: ids })
       res = withoutDesignDocuments(res)
+      res.total_rows = null // pouch indicates the total number of docs in res.total_rows, even though we uuse "keys". Setting it to null avoids cozy-client thinking there are more docs to fetch.
```
Crash--, line 310:
```suggestion
      res.total_rows = null // pouch indicates the total number of docs in res.total_rows, even though we use "keys". Setting it to null avoids cozy-client thinking there are more docs to fetch.
```

PR #733, opened 2020-07-08
feat: Can call Create on Settings

Crash--:
For now only supports /synchronized. This is useful to set the lastSync
date.
Not supported: /settings/passphrase

### DIFF
--- a/docs/api/cozy-stack-client.md
+++ b/docs/api/cozy-stack-client.md
@@ -31,9 +31,6 @@ through OAuth.</p>
 <dt><a href="#PermissionCollection">PermissionCollection</a></dt>
 <dd><p>Implements <code>DocumentCollection</code> API along with specific methods for <code>io.cozy.permissions</code>.</p>
 </dd>
-<dt><a href="#SettingsCollection">SettingsCollection</a></dt>
-<dd><p>Implements <code>DocumentCollection</code> API to interact with the /settings endpoint of the stack</p>
-</dd>
 <dt><a href="#SharingCollection">SharingCollection</a></dt>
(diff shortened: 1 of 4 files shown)


ptbrowne:
Mmh the `path` attribute seems a bit off here. I think when designing this API, we should also take other settings route in mind. The fact that this API does not support `passphrase` is a red flag to me.
And idea here : could we use instead of `path` the `_id` attribute ? I think it would make more sense.
The _id could be removed from the data sent to the stack, and the rest of the `attributes` parameters would be sent. I think this would support the passphrase case.

```
const col = client.collection('io.cozy.settings')
col.create({
    "_id": "io.cozy.settings.passphrase", // or "passphrase"
    "register_token": "37cddf40d7724988860fa0e03efd30fe",
    "passphrase": "4f58133ea0f415424d0a856e0d3d2e0cd28e4358fce7e333cb524729796b2791",
    "hint": "a hint to help me remember my passphrase",
    "key": "0.uRcMe+Mc2nmOet4yWx9BwA==|PGQhpYUlTUq/vBEDj1KOHVMlTIH1eecMl0j80+Zu0VRVfFa7X/MWKdVM6OM/NfSZicFEwaLWqpyBlOrBXhR+trkX/dPRnfwJD2B93hnLNGQ=",
    "public_key": "MIIBIjANBgkqhkiG9w...AQAB",
    "private_key": "2.wZuKkufLV31Cpw1v1TQUDA==|u6bUNTaaGxu...y7s=",
    "iterations": 10000
})

col.create({
    "_id": "io.cozy.settings.synchronized", // or "synchronized"
})
```

I think here it would be more coherent with the rest of our APIs.

With this PR, would `client.save('io.cozy.settings', 'synchronized', {})` work ?

PR #735, opened 2020-07-09
feat: Emit event during PouchLifeCycle

Crash--:
Pouch replication can be : 
- time consuming 
- resource consuming

Dispatching those events give the opportunity to the app to react differently. For instance, I don't want to launch the "photo backup process" on Drive if Pouch is doing its replication since both are very expensive

### DIFF
--- a/docs/api/cozy-pouch-link.md
+++ b/docs/api/cozy-pouch-link.md
@@ -61,12 +61,16 @@ Receives PouchDB updates (documents grouped by doctype).
 Normalizes the data (.id -> ._id, .rev -> _rev).
 Passes the data to the client and to the onSync handler.
 
+Emits an event (pouchlink:sync:end) when the sync (all doctypes) is done
+
 **Kind**: instance method of [<code>PouchLink</code>](#PouchLink)  
 <a name="PouchLink+startReplication"></a>
 
 ### pouchLink.startReplication() ⇒ <code>void</code>
 User of the link can call this to start ongoing replications.
 Typically, it can be used when the application regains focus.
+
+Emits pouchlink:sync:start event when the replication begins
 
 **Kind**: instance method of [<code>PouchLink</code>](#PouchLink)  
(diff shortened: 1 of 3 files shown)


docs/api/cozy-pouch-link.md
```diff
@@ -61,13 +61,17 @@ Receives PouchDB updates (documents grouped by doctype).
 Normalizes the data (.id -> ._id, .rev -> _rev).
 Passes the data to the client and to the onSync handler.
 
+Emit an event (pouchlink:sync:ended) when the sync (all doctypes) is done
```
ptbrowne, line 64:
```suggestion
Emits an event (pouchlink:sync:end) when the sync (all doctypes) is done
```

docs/api/cozy-pouch-link.md
```diff
@@ -61,13 +61,17 @@ Receives PouchDB updates (documents grouped by doctype).
...
+
 **Kind**: instance method of [<code>PouchLink</code>](#PouchLink)  
 <a name="PouchLink+startReplication"></a>
 
 ### pouchLink.startReplication() ⇒ <code>void</code>
 User of the link can call this to start ongoing replications.
 Typically, it can be used when the application regains focus.
 
+Emit pouchlink:sync:started event when the replication begins
```
ptbrowne, line 73:
```suggestion
Emits pouchlink:sync:start event when the replication begins
```

docs/api/cozy-pouch-link.md
```diff
@@ -76,6 +80,8 @@ Typically, it can be used when the application regains focus.
 User of the link can call this to stop ongoing replications.
 Typically, it can be used when the applications loses focus.
 
+Emit pouchlink:sync:stoped event
```
ptbrowne, line 83:
```suggestion
Emits pouchlink:sync:stop event
```

If you want to keep the "ed", it should be "stopped".

PR #738, opened 2020-07-15
feat: Add methods

ptbrowne:
- trigger.hasActionableError: returns whether the user can solve the error from the
  trigger
- getCreatedByApp: returns the app that created the document

### DIFF
--- a/packages/cozy-client/src/models/index.js
+++ b/packages/cozy-client/src/models/index.js
@@ -22,5 +22,6 @@ export {
   note,
   account,
   accounts,
-  permission
+  permission,
+  utils
 }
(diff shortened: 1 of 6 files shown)


packages/cozy-client/src/models/index.js
```diff
@@ -6,6 +6,7 @@ import * as folder from './folder'
 import * as account from './account'
 import * as note from './note'
 import * as permission from './permission'
+import * as utils from './utils'
```
y-lohse, line 9: I don't think there is a utils model?

ptbrowne: https://github.com/cozy/cozy-client/pull/738/files#diff-3be6cca06b5c09545418c6c909dc9d87 ?

y-lohse: I missed the second commit 🤦‍♂️

PR #743, opened 2020-07-22
feat: Add ContactCollection

ptbrowne:
To find the myself contact, we need to use a special route on
the stack. This routes creates the contact if it does not
exist

### DIFF
--- a/packages/cozy-stack-client/src/ContactsCollection.spec.js
+++ b/packages/cozy-stack-client/src/ContactsCollection.spec.js
@@ -1,0 +1,51 @@
+import CozyStackClient from './CozyStackClient'
+import ContactsCollection from './ContactsCollection'
+
+const stackMyselfResponse = {
+  data: {
+    type: 'io.cozy.contacts',
+    id: 'bf91cce0-ef48-0137-2638-543d7eb8149c',
+    attributes: {
+      fullname: 'Alice',
+      email: [{ address: 'alice@example.com', primary: true }]
+    },
+    meta: {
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/ContactsCollection.spec.js
```diff
@@ -0,0 +1,51 @@
...
+          return stackMyselfResponse
+        } else {
+          return []
+        }
+      })
+    const col = new ContactsCollection('io.cozy.contacts', stackClient)
+    return { stackClient, col }
+  }
+
+  it('should use a special route for the myself contact', async () => {
```
y-lohse, line 34: Maybe we should also check that the opposite is true, eg. a normal find call uses the DocumentCollection?

ptbrowne: Done, in fact I had removed the test since during tests I had a V8 crash that I did not want to understand further. Turns out that I needed to mock the _index route for the test not to run in a loop. I'll try to dig further into the implementation of DocumentCollection.find to see if there is something that we could do to remove the risk of an infinite loop (ContactCollection.find -> DocumentCollection.find -> ContactCollection.find due to DocumentCollection::createIndex recalling this.find if index creation fails). I think this can be done in another PR.

ptbrowne: I opened https://github.com/cozy/cozy-client/issues/744 to track this issue.

PR #760, opened 2020-08-17
fix: Method to get contact index and contact initials

JF-Cozy:
- The method to get index by FamilyNameGivenNameEmailCozyUrl returns now an empty object instead of empty string when empty. This is a little trick to get empty value at the end, according to https://docs.couchdb.org/en/2.3.1/ddocs/views/collation.html#collation-specification
- Besides, the method capitalize the returned value to simplify things for sorting contacts in the application
- getInitials() now deals with cozy url and empty name object

### DIFF
--- a/packages/cozy-client/src/models/contact.spec.js
+++ b/packages/cozy-client/src/models/contact.spec.js
@@ -81,6 +81,34 @@ describe('getInitials', () => {
     expect(result).toEqual('A')
   })
 
+  it('should return the first letter of the primary email if contact has empty name', () => {
+    const contact = {
+      name: {},
+      email: [
+        {
+          address: 'arya.stark@thenorth.westeros',
+          primary: true
+        }
+      ]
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/models/contact.spec.js
```diff
@@ -81,6 +81,34 @@ describe('getInitials', () => {
...
+          address: 'arya.stark@thenorth.westeros',
+          primary: true
+        }
+      ]
+    }
+    const result = getInitials(contact)
+    expect(result).toEqual('A')
+  })
+
+  it('should return the first letter of the cozy domain if contact has empty name', () => {
```
Crash--, line 98:
```suggestion
  it('should return the first letter of the cozy domain if contact has empty name, an empty email but has a cozy url', () => {
```

Crash-- (approved):

PR #765, opened 2020-08-27
feat: We can pass a codes argument at the perm creation

Crash--:
Use Case: We want to be able to create permission for all kind of doctype. Today, we can do that with createSharingLink() but this method is too much paired to Files. 

We want here, to be able to create a permission with want we want, for instance : 
```
 permissions: { groups: { type: GROUPS_DOCTYPE, verbs: ['GET'] }, contacts: { type: CONTACTS_DOCTYPE, verbs: ['GET'] } }
```

Before the change, we're able to do that, but we can't specify the `codes` we want to produced. Now we can create myCode1 and myCode2 : 
```
create({codes: 'myCode1,myCode2', permissions: { groups: { type: GROUPS_DOCTYPE, verbs: ['GET'] }, contacts: { type: CONTACTS_DOCTYPE, verbs: ['GET'] } } }
==> response 
shortcodes: {"myCode1": "adcdef", "myCode2": "abfdgds"}.
```

BREAKING CHANGE: By default, we defaulted the codes values
to `code` for PermissionsCollection.create. Without we
receive an object without key in response
shortcodes: {"": "abcdfef"}. Now by default we have :
shortcodes: {"code": "adcdef"}.

It's a BC, but this method is not used in our code base for
now.

### DIFF
--- a/packages/cozy-stack-client/src/PermissionCollection.js
+++ b/packages/cozy-stack-client/src/PermissionCollection.js
@@ -18,13 +18,27 @@ class PermissionCollection extends DocumentCollection {
     }
   }
 
-  async create({ _id, _type, ...attributes }) {
-    const resp = await this.stackClient.fetchJSON('POST', uri`/permissions`, {
-      data: {
-        type: 'io.cozy.permissions',
-        attributes
+  /**
+   * Create a new set of permissions
+   * It can also associates one or more codes to it, via the codes parameter
+   *
+   * @param {object} permission
+   * @param {string} permission.codes A comma separed list of values (defaulted to code)
+   *
+   * @see https://docs.cozy.io/en/cozy-stack/permissions/#post-permissions
+   *
+   */
+  async create({ _id, _type, codes = 'code', ...attributes }) {
+    const resp = await this.stackClient.fetchJSON(
+      'POST',
+      uri`/permissions?codes=${codes}`,
+      {
+        data: {
+          type: 'io.cozy.permissions',
+          attributes
+        }
       }
-    })
+    )
     return {
       data: normalizePermission(resp.data)
     }
--- a/packages/cozy-stack-client/src/PermissionCollection.spec.js
+++ b/packages/cozy-stack-client/src/PermissionCollection.spec.js
@@ -21,7 +21,30 @@ describe('PermissionCollection', () => {
       client.fetch.mockReset()
       client.fetchJSON.mockReturnValue(Promise.resolve({ data: [] }))
     })
+    describe('create', () => {
+      it('calls with the right args', async () => {
+        await collection.create({
+          _type: 'io.cozy.permissions',
+          _id: 'a340d5e0d64711e6b66c5fc9ce1e17c6',
+          codes: 'a,b'
+        })
+        expect(client.fetchJSON).toHaveBeenCalledWith(
+          'POST',
+          '/permissions?codes=a%2Cb',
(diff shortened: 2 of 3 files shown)


packages/cozy-stack-client/src/PermissionCollection.spec.js
```diff
@@ -21,7 +21,30 @@ describe('PermissionCollection', () => {
       client.fetch.mockReset()
       client.fetchJSON.mockReturnValue(Promise.resolve({ data: [] }))
     })
+    describe('create', () => {
+      it('calls with the right agrs', async () => {
```
JF-Cozy, line 25: maybe something more explicit like `should be called with 'code' argument` ?

Anyway there is a typo on 'agrs'

JF-Cozy:
Nit: just a little typo in description : "We want here, to be able to create a permission with **what** we want, for instance" and some line breaks at the end

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -18,13 +18,27 @@ class PermissionCollection extends DocumentCollection {
...
+   * @param {object} permission
+   * @param {string} permission.codes A comma separed list of values (defaulted to code)
+   *
+   * @see https://docs.cozy.io/en/cozy-stack/permissions/#post-permissions
+   *
+   */
+  async create({ _id, _type, codes = 'code', ...attributes }) {
+    const resp = await this.stackClient.fetchJSON(
+      'POST',
+      `/permissions?codes=${codes}`,
```
y-lohse, line 34: We should call the `uri` template/function here to properly encode `codes`.

y-lohse (commented):
Ok for me with the suggested changes, but there is a pattern emerging where we insert what should be separate parameters into the `attributes` of the objects that we pass to the stack, in order to be able to use queries. IMO we need to look for a better way to do this.

packages/cozy-stack-client/src/PermissionCollection.js
Crash--: 👍

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -18,13 +18,27 @@ class PermissionCollection extends DocumentCollection {
...
+   * @param {object} permission
+   * @param {string} permission.codes A comma separed list of values (defaulted to code)
+   *
+   * @see https://docs.cozy.io/en/cozy-stack/permissions/#post-permissions
+   *
+   */
+  async create({ _id, _type, codes = 'code', ...attributes }) {
+    const resp = await this.stackClient.fetchJSON(
+      'POST',
+      `/permissions?codes=${uri`${codes}`}`,
```
y-lohse, line 34: I think we usually do 

```suggestion
      uri`/permissions?codes=${codes}`,
```

😁

y-lohse (approved):

packages/cozy-stack-client/src/PermissionCollection.js
Crash--: I don't want to espace the `?` since uri`` will return %3F  and the stack will not be able to decode it.

y-lohse: I think it wouldn't encode it, it only encodes [the interpolated variables](https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/utils.js#L17). It's kinda cool tbh.

PR #772, opened 2020-09-02
feat: Add hasQueryMoreToFetch()

JF-Cozy:
- Returns true|false if query has more data to fetch

### DIFF
--- a/packages/cozy-client/src/utils.js
+++ b/packages/cozy-client/src/utils.js
@@ -3,7 +3,7 @@
  *
  * Rejects with canceled: true as soon as cancel is called
  *
- * @param  {Promise} promise
+ * @param {Promise} promise
  * @returns {AugmentedPromise} - Promise with .cancel method
  */
 const cancelable = promise => {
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/utils.js
```diff
@@ -41,3 +41,14 @@ export const isQueryLoading = col => {
...
+
+/**
+ * Returns whether a query has more data to fetch
+ */
+export const hasQueryMoreToFetch = col => {
+  if (!col) {
+    console.warn('hasQueryMoreToFetch called on falsy value.') // eslint-disable-line no-console
+    return false
+  }
+  return col.fetchStatus === 'loaded' && col.hasMore
```
ptbrowne, line 53: Why is there a check on the fetchStatus here ?

JF-Cozy: Because if there is more data to fetch, `hasMore` is `false` while `fetchStatus` is not `loaded` (so `pending` or `loading`).

This function is used to fetch more data after the first batch (whose number of results depends on `limitBy`), so I thought we need the first result before fetching more.

But you're right it's not necessary

y-lohse: But in that case… is this function really necessary? Checking `result.hasMore` seems easier to me than `hasQueryMoreToFetch(result)` + the import that goes with it.

JF-Cozy: It's the same approach as for `hasQueryBeenLoaded` I think it brings a better readability of the code, isn't it ?

JF-Cozy: I'm going to save everyone a little time by shutting down that PR. If it hasn't been done so far, it's probably for the reasons you mention :)

PR #775, opened 2020-09-03
Fix query props

ptbrowne:
If queryConnect had one of its collection whose queryOpts was a
function, for the "query" and "as" to change based on props, there
was a problem : the queryDefinition would not change based on props,
and it would keep the 1st queryDefinition. This was due to reassignment
to function params.

After fixing the bug, and adding a test, I activated the eslint rule
to have it as errors, and fixed the other cases in all the repo.

Here is a simple example of the bug:

```
const hoc = queryOpts => Component => {
  return props => {
    queryOpts = typeof queryOpts === "function" ? queryOpts(props) : queryOpts;
    Component(queryOpts)
  }
}

const queryOpts = props => ({ docId: props.id })

const Component = queryOpts => {
  console.log('Rendering the component', queryOpts)
} 

const hComponent = hoc(queryOpts)(Component)

hComponent({ id: 1 })
hComponent({ id: 2 })

// Rendering the component { docId: 1 }
// Rendering the component { docId: 1 }
```

### DIFF
--- a/packages/cozy-client/src/hoc.jsx
+++ b/packages/cozy-client/src/hoc.jsx
@@ -27,6 +27,7 @@ const withQuery = (dest, queryOpts, Original) => {
       `withQuery has no options for ${dest} (wrapping ${Original.name})`
     )
   }
+
   return Component => {
     const Wrapper = (props, context) => {
       if (!context.client) {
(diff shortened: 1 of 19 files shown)


packages/cozy-client/src/hoc.jsx
```diff
@@ -35,14 +36,16 @@ const withQuery = (dest, queryOpts, Original) => {
...
-        return <Component {...{ [dest]: queryOpts.doc, ...props }} />
+        return <Component {...{ [dest]: queryOptsRes.doc, ...props }} />
       }
 
       return (
-        <Query {...queryOpts}>
+        <Query {...queryOptsRes}>
           {result => <Component {...{ [dest]: result, ...props }} />}
         </Query>
       )
```
Crash--, line 51: trop biiien ! 👏

Crash-- (approved):

PR #776, opened 2020-09-03
fix: Remove deleted document from store

ptbrowne:
Previously, deleted documents were removed from queries, but the doc
was staying in the documents slice. This is problematic in the case
of relationships since the "parent" document linked to a document
that has been deleted, will still see the deleted document.

This is also problematic if we simply rely on the documents slice to
know if a document is deleted or not.

This PR fixes both situations.

### DIFF
--- a/packages/cozy-client/src/CozyClient.spec.js
+++ b/packages/cozy-client/src/CozyClient.spec.js
@@ -607,6 +607,34 @@ describe('CozyClient', () => {
     })
   })
 
+  describe('destroy', () => {
+    let doc
+    beforeEach(() => {
+      client.setData(
+        normalizeData({
+          'io.cozy.todos': [TODO_1]
+        })
+      )
+      doc = { ...TODO_1, label: 'Buy croissants' }
+      client.store.dispatch.mockReset()
+    })
+
+    it('should remove the document from the store', async () => {
+      await client.destroy(doc)
+      expect(client.store.dispatch).toHaveBeenCalledWith(
+        expect.objectContaining({
+          definition: {
+            document: expect.objectContaining({
+              _id: 'todo_1'
+            }),
+            mutationType: 'DELETE_DOCUMENT'
+          },
+          mutationId: 1
+        })
+      )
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/__tests__/store.spec.js
```diff
@@ -228,6 +229,64 @@ describe('Store', () => {
...
+
+      afterEach(() => {
+        console.warn = originalWarn
+      })
+
+      it('should remove a deleted doc received in a mutation result', () => {
+        store.dispatch(
+          receiveQueryResult('allTodos', {
+            data: [TODO_1, TODO_2],
+            meta: { count: 1 },
```
y-lohse, line 258:
```suggestion
            meta: { count: 2 },
```

packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -607,6 +607,30 @@ describe('CozyClient', () => {
...
 
+  describe('destroy', () => {
+    it('should remove the document from the store', async () => {
+      client.setData(
+        normalizeData({
+          'io.cozy.todos': [TODO_1]
+        })
+      )
+      const doc = { ...TODO_1, label: 'Buy croissants' }
+      client.store.dispatch.mockReset()
```
y-lohse, line 618: Nit: I find it a bit odd to reset a mock in the middle of a test, I think it's usually preferable to do those in `beforeEach/afterEach`.

y-lohse (approved):
All good, makes sense!

packages/cozy-client/src/CozyClient.spec.js
ptbrowne: It's because I am not interested in the dispatch that the setData has done. Would it be clearer for you if I moved the above lines in a beforeEach ?

ptbrowne: I've made the change to put lines in a beforeEach, I agree that it's clearer.

Issue #526, opened 2019-09-24
fetchStatus goes to loading and does not change to loaded


ptbrowne:
It seems that sometimes, a query can be marked as `loading` when in fact it is not loading. This is made more visible when using fetchPolicies since there is no `fetch` whose return callback would set the fetchStatus back to `loaded`.

---

It seems the condition to re-init a query should be changed.

```patch
     // Don't trigger the INIT_QUERY for fetchMore() calls
-    if (existingQuery.fetchStatus !== 'loaded' || !queryDefinition.skip) {
+    if (!existingQuery.lastFetch && existingQuery.fetchStatus !== 'loading') {
```

This way we do not overwrite the status from `loaded` to `loading` when we want to only make sure the query exists in the store.

Crash--:
I'm closing this issue since you fixed the bug in https://github.com/cozy/cozy-client/pull/777

PR #777, opened 2020-09-04
Fix: initQuery doesn't set fetchStatus: loading on queries for noninitial calls

ptbrowne:
This fixes a long standing bug : #526

Previously, after an initial load of data through query(), subsequent calls
to query() would set the fetchStatus to loading. It causes a problem in
case of fetchPolicy : if the fetchPolicy returns false, the
receiveQueryResult will never been dispatched and queryState.fetchStatus 
will stay to false.

To prevent the problem, an additional loadQuery call is introduced which 
sets the fetchStatus to loading. initQuery now does not set the fetchStatus
to loading if the query already has been fetched.

### DIFF
--- a/packages/cozy-client/src/store/queries.spec.js
+++ b/packages/cozy-client/src/store/queries.spec.js
@@ -22,15 +22,26 @@ describe('queries reducer', () => {
   })
 
   it('should init correctly', () => {
-    applyAction(
-      initQuery(
-        'a',
-        new QueryDefinition({
-          doctype: 'io.cozy.todos'
-        })
-      )
-    )
+    const qdef = new QueryDefinition({
+      doctype: 'io.cozy.todos'
+    })
+    applyAction(initQuery('a', qdef))
     expect(state).toMatchSnapshot()
+  })
+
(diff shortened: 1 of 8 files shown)


packages/cozy-client/src/store/queries.spec.js
```diff
@@ -22,15 +22,40 @@ describe('queries reducer', () => {
...
 
   it('should init correctly', () => {
+    const qdef = new QueryDefinition({
+      doctype: 'io.cozy.todos'
+    })
+    applyAction(initQuery('a', qdef))
+    expect(state).toMatchSnapshot()
+  })
+
+  test('init should have no effect after receiveQueryResult', () => {
```
y-lohse, line 32: Very minor, but it's strange to use `test` instead of `it` all of a sudden.

y-lohse (approved):
LGTM

packages/cozy-client/src/store/queries.spec.js
ptbrowne: It's because the sentence cannot be read with "it" :)

PR #781, opened 2020-09-11
Add static helpers for HasMany relationship

ptbrowne:
Main feature is the addition of static helpers to deal with
HasMany relationship data.

There are smaller commits whose description should be enough.


Crash-- (approved):
👍

PR #785, opened 2020-09-15
Client-side query sort can use deep objects properties

JF-Cozy:
**Problem to solve :** 
Before that, if we used nested attributes in `sortBy()` it was ignored and no sorting was done in cozy-client. As a consequence, after editing a document, the sorting was not done again if a new request was not launched. The documents remained in their original position.

**Improvements :** 
- We can now use `sortBy()` with nested attributes like `sortBy([{ 'deep.property': 'asc' }])` as we do here https://github.com/cozy/cozy-contacts/blob/master/src/helpers/queries.js#L57
- getIndexByFamilyNameGivenNameEmailCozyUrl() now tranforms the index in lower case (not capitalize) to match this :  https://github.com/cozy/cozy-client/blob/fix/queries/packages/cozy-client/src/store/queries.js#L176
- getIndexByFamilyNameGivenNameEmailCozyUrl() now return `null` (instead of `empty object`) to have empty index at the end of the sort.

**Explanations for changing the `empty object` to `null` :** 
We thought using the sort order of couchDB (see https://docs.couchdb.org/en/2.3.1/ddocs/views/collation.html#collation-specification) during a `sortBy()` but in fine we sort the documents in cozy-client with an `Array.sort()`. Empty objects are at the end in couchDB but before not empty lowercase string with `Array.sort()`

**Details :** 
The entry point in cozy-client: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/queries.js#L46
Then we use `orderBy()` from lodash (https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/queries.js#L202) which is based on `_baseOrderBy()` itself based on `_baseSortBy()`, which finally makes an `Array.sort()` https://github.com/lodash/lodash/blob/master/.internal/baseSortBy.js#L14

The sort order is therefore that of `Array.sort()` (see https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/sort and https://fr.wikipedia.org/wiki/Table_des_caract%C3%A8res_Unicode/U0000)

**Consequences :** 
Since couchDB and cozy-client do not sort the same way, paging will be a problem.

Edit : documentation about queries https://docs.cozy.io/en/tutorials/data/queries/ will be updated according to this

### DIFF
--- a/packages/cozy-client/src/models/contact.js
+++ b/packages/cozy-client/src/models/contact.js
@@ -1,4 +1,5 @@
-import { get, capitalize, isEmpty } from 'lodash'
+import get from 'lodash/get'
+import isEmpty from 'lodash/isEmpty'
 
 export const getPrimaryOrFirst = property => obj =>
   !obj[property] || obj[property].length === 0
@@ -134,22 +135,25 @@ export const getDisplayName = contact =>
  * @returns {string} - the contact's 'byFamilyNameGivenNameEmailCozyUrl' index
  */
 export const getIndexByFamilyNameGivenNameEmailCozyUrl = contact => {
-  const indexByFamilyNameGivenNameEmailCozyUrl = capitalize(
-    [
-      get(contact, 'name.familyName', ''),
-      get(contact, 'name.givenName', ''),
-      getPrimaryEmail(contact),
-      getPrimaryCozyDomain(contact)
-    ]
-      .join('')
-      .trim()
-  )
+  if (get(contact, 'indexes.byFamilyNameGivenNameEmailCozyUrl')) {
+    if (isEmpty(contact.indexes.byFamilyNameGivenNameEmailCozyUrl)) {
+      return null
+    }
+    return contact.indexes.byFamilyNameGivenNameEmailCozyUrl
+  }
 
-  return get(
-    contact,
-    'indexes.byFamilyNameGivenNameEmailCozyUrl',
-    indexByFamilyNameGivenNameEmailCozyUrl.length === 0
-      ? {}
-      : indexByFamilyNameGivenNameEmailCozyUrl
-  )
+  const createIndexByFamilyNameGivenNameEmailCozyUrl = [
+    get(contact, 'name.familyName', ''),
+    get(contact, 'name.givenName', ''),
+    getPrimaryEmail(contact),
+    getPrimaryCozyDomain(contact)
+  ]
+    .join('')
+    .trim()
+    .toLowerCase()
+
+  if (createIndexByFamilyNameGivenNameEmailCozyUrl.length === 0) {
+    return null
+  }
+  return createIndexByFamilyNameGivenNameEmailCozyUrl
 }
--- a/packages/cozy-client/src/store/queries.js
+++ b/packages/cozy-client/src/store/queries.js
@@ -172,7 +172,7 @@ const getQueryDocumentsChecker = query => {
 }
 
 const makeCaseInsensitiveStringSorter = attrName => item => {
-  const attrValue = item[attrName]
+  const attrValue = get(item, attrName)
   return isString(attrValue) ? attrValue.toLowerCase() : attrValue
 }
 
--- a/packages/cozy-client/src/store/queries.spec.js
+++ b/packages/cozy-client/src/store/queries.spec.js
@@ -256,4 +256,18 @@ describe('makeSorterFromDefinition', () => {
     const sorted = sorter(files)
     expect(sorted.map(x => x._id)).toEqual(['4', '5', '3', '2', '1'])
   })
+
+  it('should handle objects with deep properties', () => {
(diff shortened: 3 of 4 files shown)


packages/cozy-client/src/store/queries.js
```diff
@@ -172,7 +172,7 @@ const getQueryDocumentsChecker = query => {
 }
 
 const makeCaseInsensitiveStringSorter = attrName => item => {
-  const attrValue = item[attrName]
+  const attrValue = get(item, attrName)
   return isString(attrValue) ? attrValue.toLowerCase() : attrValue
 }
```
ptbrowne, line 177: I'd label this commit as a feat.

feat: Client-side query sort handles nested attributes

packages/cozy-client/src/store/queries.spec.js
```diff
@@ -256,4 +256,40 @@ describe('makeSorterFromDefinition', () => {
...
+      { _id: '4', name: { label: 'A' } },
+      { _id: '5', name: { label: 'b' } },
+      { _id: '6', name: { label: [] } },
+      { _id: '7', name: { label: '' } },
+      { _id: '8', name: { label: 1 } },
+      { _id: '9', name: { label: {} } },
+      { _id: '10', name: { label: false } },
+      { _id: '11', name: { label: null } },
+      { _id: '12', name: { label: true } },
+      { _id: '13', name: { label: undefined } }
```
ptbrowne, line 276: Why do we have all those cases here ? Aren't we testing the behavior of Array.sort from _id 6 to _id 13 ?

packages/cozy-client/src/models/contact.js
```diff
@@ -134,22 +135,21 @@ export const getDisplayName = contact =>
...
-  )
+  const indexByFamilyNameGivenNameEmailCozyUrl = [
+    get(contact, 'name.familyName', ''),
+    get(contact, 'name.givenName', ''),
+    getPrimaryEmail(contact),
+    getPrimaryCozyDomain(contact)
+  ]
+    .join('')
+    .trim()
+    .toLowerCase()
```
ptbrowne, line 146: I haven't understood well why it was necessary to lowercase here ? I thought that makeCaseInsensitiveStringSorter would already do this. Is it to fix a bug ? What was the behavior ?

ptbrowne (approved):

packages/cozy-client/src/models/contact.js
JF-Cozy: This is to be consistent between the data in couchDB and the data in cozy-client, in order to have as much as possible an identical sorting behavior.

packages/cozy-client/src/store/queries.spec.js
JF-Cozy: Yes, that's exactly right.

packages/cozy-client/src/models/contact.js
ptbrowne: Is it to fix a bug ? What was the behavior ?

packages/cozy-client/src/models/contact.js
```diff
@@ -134,22 +135,21 @@ export const getDisplayName = contact =>
...
 
   return get(
     contact,
     'indexes.byFamilyNameGivenNameEmailCozyUrl',
     indexByFamilyNameGivenNameEmailCozyUrl.length === 0
-      ? {}
+      ? null
```
Crash--, line 152: Since we have a few cozies with this {} format, we should test if indexByFamilyNameGivenNameEmailCozyUrl === {} and set it to null if needed?

packages/cozy-client/src/store/queries.spec.js
JF-Cozy: I'm will delete that to stay within an application perimeter.

packages/cozy-client/src/store/queries.spec.js
```diff
@@ -256,4 +256,18 @@ describe('makeSorterFromDefinition', () => {
...
+    const sorter = makeSorterFromDefinition(q)
+    const files = [
+      { _id: '1', name: { label: 'C' } },
+      { _id: '2', name: { label: 'B' } },
+      { _id: '3', name: { label: 'a' } },
+      { _id: '4', name: { label: 'A' } },
+      { _id: '5', name: { label: 'b' } }
+    ]
+    const sorted = sorter(files)
+    expect(sorted.map(x => x._id)).toEqual(['3', '4', '2', '5', '1'])
```
Crash--, line 271: je trouverai ça plus facile à lire de tester le label plutôt que les id.

packages/cozy-client/src/models/contact.js
JF-Cozy: Description updated to be more explicit

PR #787, opened 2020-09-17
Feat/warmed up queries

Crash--:
Since first Pouch request can be slow (Pouch creates and populates the
index before returning the result), we need in a some cases (
like the first sync), to be able to warmed up a few queries to
create the index before making the real call.

Until those warmupQueries are done, we forward the request to the next link.

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -136,7 +136,8 @@ class PouchLink extends CozyLink {
       doctypesReplicationOptions: this.doctypesReplicationOptions,
       onError: err => this.onSyncError(err),
       onSync: this.handleOnSync.bind(this),
-      prefix
+      prefix,
+      executeQuery: this.executeQuery.bind(this)
     })
 
     if (this.client && this.options.initialSync) {
@@ -259,6 +260,15 @@ class PouchLink extends CozyLink {
       return forward(operation)
     }
 
+    if (this.needsToWaitWarmup(doctype)) {
+      if (process.env.NODE_ENV !== 'production') {
+        logger.info(
+          `Tried to access local ${doctype} but not warmuped yet. Forwarding the operation to next link`
+        )
+      }
+      return forward(operation)
+    }
+
     // Forwards if doctype not supported
     if (!this.supportsOperation(operation)) {
       if (process.env.NODE_ENV !== 'production') {
@@ -274,6 +284,26 @@ class PouchLink extends CozyLink {
     } else {
       return this.executeQuery(operation)
     }
+  }
+  /**
+   *
+   * @param {string} doctype
+   * @returns {boolean} the need to wait for the warmup
+   * Check if there is warmup queries for this doctype
+   * and return if those queries are already warmedup or not
+   */
+  needsToWaitWarmup(doctype) {
+    if (
+      this.doctypesReplicationOptions &&
+      this.doctypesReplicationOptions[doctype] &&
+      this.doctypesReplicationOptions[doctype].warmupQueries
+    ) {
+      return !this.pouches.areWarmedUpQueries(
+        doctype,
+        this.doctypesReplicationOptions[doctype].warmupQueries
+      )
+    }
+    return false
   }
 
   hasIndex(name) {
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -298,7 +298,7 @@ class PouchLink extends CozyLink {
       this.doctypesReplicationOptions[doctype] &&
       this.doctypesReplicationOptions[doctype].warmupQueries
     ) {
-      return !this.pouches.areWarmedUpQueries(
+      return !this.pouches.areQueriesWarmedUp(
         doctype,
         this.doctypesReplicationOptions[doctype].warmupQueries
       )
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -287,10 +287,11 @@ class PouchLink extends CozyLink {
   }
   /**
    *
-   * @param {string} doctype
-   * @returns {boolean} the need to wait for the warmup
    * Check if there is warmup queries for this doctype
    * and return if those queries are already warmedup or not
+   *
+   * @param {string} doctype
+   * @returns {boolean} the need to wait for the warmup
    */
   needsToWaitWarmup(doctype) {
     if (
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -311,7 +311,7 @@ class PouchLink extends CozyLink {
   }
 
   async ensureIndex(doctype, query) {
-    const fields = getIndexFields(query)
+    const fields = query.indexedFields || getIndexFields(query)
     const name = getIndexNameFromFields(fields)
     const absName = `${doctype}/${name}`
     const db = this.pouches.getPouch(doctype)
@@ -336,7 +336,8 @@ class PouchLink extends CozyLink {
     limit,
     id,
     ids,
-    skip
+    skip,
+    indexedFields
   }) {
     const db = this.getPouch(doctype)
     let res, withRows
@@ -360,7 +361,7 @@ class PouchLink extends CozyLink {
         limit,
         skip
       }
-      await this.ensureIndex(doctype, findOpts)
+      await this.ensureIndex(doctype, { ...findOpts, indexedFields })
       res = await find(db, findOpts)
       res.offset = skip
       res.limit = limit
--- a/packages/cozy-pouch-link/src/CozyPouchLink.spec.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.spec.js
@@ -98,7 +98,7 @@ describe('CozyPouchLink', () => {
       })
     })
 
-    it('test if the pouch is synced and no warmup queries for this doctype, it should not forward', async () => {
+    it('should not forward if the pouch is synced and there is no warmup queries for this doctype', async () => {
       await setup({
         doctypesReplicationOptions: {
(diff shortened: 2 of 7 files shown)


packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -311,7 +311,7 @@ class PouchLink extends CozyLink {
   }
 
   async ensureIndex(doctype, query) {
-    const fields = getIndexFields(query)
+    const fields = query.indexedFields || getIndexFields(query)
```
ptbrowne, line 314: Can you add a test ? It'd be a little easier to understand what was not working before.

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -5,103 +5,22 @@ import get from 'lodash/get'
...
-  })
-
-  const cancel = () => {
-    if (replication) {
-      replication.cancel()
-    }
-  }
-
-  promise.cancel = cancel
-  return promise
```
ptbrowne, line 100: Can you split the commit refactoring startReplication out ? It's difficult to know if something has changed in this function.

Crash--: I'll 👍 . Nothing change

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -275,6 +285,26 @@ class PouchLink extends CozyLink {
       return this.executeQuery(operation)
     }
   }
+  /**
+   *
+   * @param {string} doctype
+   * @returns {boolean} the need to wait for the warmup
+   * Check if there is warmup queries for this doctype
+   * and return if those queries are already warmedup or not
```
ptbrowne, line 293: Please put the description the docstring above the definition of the params, with a line break between params and description.

packages/cozy-pouch-link/src/CozyPouchLink.spec.js
```diff
@@ -72,6 +85,34 @@ describe('CozyPouchLink', () => {
...
+        }
+      })
+      const query = Q(TODO_DOCTYPE)
+      expect.assertions(1)
+      await link.request(query, null, () => {
+        expect(true).toBe(true)
+      })
+    })
+
+    it('test if the pouch is synced and no warmup queries for this doctype, it should not forward', async () => {
```
ptbrowne, line 101:
```suggestion
     it('should not forward if the pouch is synced and there no warmup queries for this doctype', async () => {
```

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -165,9 +176,16 @@ class PouchManager {
         getReplicationURL
       ).then(res => {
         this.addSyncedDoctype(doctype)
-
         logger.log('PouchManager: Replication for ' + doctype + ' ended', res)
-
+        //We want to warmupQueries only once by "start". aka we want to warmup
+        //every time we launch the client (app) but not every time a replication
+        //is done (aka every 30s at least)
```
ptbrowne, line 182: I think it could be clearer.

What do you think of :

```
// Queries are warmed up only once per instantiation of the PouchManager. Since
// the PouchManager lives during the complete lifecycle of the app, warm up
// happens only on app start / restart.
```

Please take the habit of putting a space after the double slash 👼 .

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -165,9 +176,16 @@ class PouchManager {
...
+        if (
+          !this.warmedUpQueries[doctype] &&
+          replicationOptions.warmupQueries
+        ) {
+          this.warmupQueries(doctype, replicationOptions.warmupQueries)
+        }
```
ptbrowne, line 188: Maybe it would be clearer if this was extracted into `this.checkToWarmupDoctype`, this way, we could put the comment above as the docstring of the function, I think it would be clearer.

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -263,6 +281,54 @@ class PouchManager {
...
+
+  async warmupQueries(doctype, queries) {
+    if (!this.warmedUpQueries[doctype]) this.warmedUpQueries[doctype] = []
+    await Promise.all(
+      queries.map(query => {
+        const def = getQueryAlias(query)
+        if (!this.warmedUpQueries[doctype].includes(def)) {
+          this.warmedUpQueries[doctype].push(def)
+          try {
+            return this.executeQuery(query.definition().toDefinition())
```
ptbrowne, line 293: I think we want to await the end of the executeQuery. Otherwise, in case of async error, we will not delete the entry in this.warmedUpQueries.

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -263,6 +281,54 @@ class PouchManager {
...
+
+  persistwarmedUpQueries() {
+    window.localStorage.setItem(
+      LOCALSTORAGE_WARMUPEDQUERIES_KEY,
+      JSON.stringify(this.warmedUpQueries)
+    )
+  }
+
+  areWarmedUpQueries(doctype, queries) {
+    const persistwarmedUpQueries = this.getPersistedWarmedUpQueries()
```
ptbrowne, line 312: s/areWarmedUpQueries/areQueriesWarmedUp
s/persistwarmedUpQueries/persistWarmedUpQueries

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -263,6 +281,54 @@ class PouchManager {
...
+
+  getPersistedWarmedUpQueries() {
+    const item = window.localStorage.getItem(LOCALSTORAGE_WARMUPEDQUERIES_KEY)
+    if (!item) {
+      return {}
+    }
+    return JSON.parse(item)
+  }
+
+  destroyWarmedUpQueries() {
```
ptbrowne, line 328: I think it's better to keep destroy for async operations ? clear sounds more harmless here and I think is better suited for a cache.

packages/cozy-pouch-link/src/PouchManager.spec.js
```diff
@@ -266,6 +303,106 @@ describe('PouchManager', () => {
...
+      expect(manager.syncedDoctypes).toEqual(['io.cozy.todos'])
+    })
+
+    it('Should persist the new synced doctypes list', () => {
+      const manager = new PouchManager(['io.cozy.todos'], managerOptions)
+      manager.persistSyncedDoctypes = jest.fn()
+      manager.addSyncedDoctype('io.cozy.todos')
+
+      expect(manager.persistSyncedDoctypes).toHaveBeenCalledTimes(1)
+    }) */
```
ptbrowne, line 350: Why are both tests commented ?

packages/cozy-pouch-link/src/PouchManager.spec.js
```diff
@@ -334,4 +471,51 @@ describe('PouchManager', () => {
...
+          'io.cozy.todos': {
+            strategy: 'fromRemote',
+            warmupQueries: [query1(), query2()]
+          }
+        }
+      }
+      manager = new PouchManager(['io.cozy.todos'], newManagerOptions)
+    })
+
+    it('should executes warmeupQueries on the first replicationLoop only', async () => {
```
ptbrowne, line 492: Can you add test for the clearing of the this.warmupQueries entry if there is an (a)sync error in client.executeQuery ?

ptbrowne (commented):
I think we need to test what happens in case of error in executeQuery and I've got some small remarks.

packages/cozy-pouch-link/src/PouchManager.js
Crash--: spaced comment https://github.com/cozy/cozy-libs/pull/1107 

Thanks 👍

packages/cozy-pouch-link/src/PouchManager.js
Crash--: Hmm, I think I've this commit somewhere, let me find it 👍

packages/cozy-pouch-link/src/PouchManager.spec.js
Crash--: mishandling, no need for those tests

packages/cozy-pouch-link/src/CozyPouchLink.spec.js
```diff
@@ -323,4 +363,32 @@ describe('CozyPouchLink', () => {
...
   })
+
+  describe('index creation', () => {
+    let spy
+
+    afterEach(() => {
+      spy.mockRestore()
+    })
+
+    it('uses the default index aka the from the sort', async () => {
```
ptbrowne, line 374:
```suggestion
    it('uses the default index, the one from the sort', async () => {
```

I do not think that aka was OK to use here.

packages/cozy-pouch-link/src/PouchManager.spec.js
```diff
@@ -334,4 +454,63 @@ describe('PouchManager', () => {
...
+      expect(manager.getPersistedWarmedUpQueries()).toEqual({
+        'io.cozy.todos': ['query1', 'query2']
+      })
+      //Simulation of a loop. Let's replicate again
+      await manager.replicateOnce()
+      await sleep(10)
+      expect(executeMock).toHaveBeenCalledTimes(2)
+    })
+
+    it('should not persist or cache warmedup queries if one have failled', async () => {
```
ptbrowne, line 504:
```suggestion
    it('should not persist or cache warmedup queries if one has failed', async () => {
```

packages/cozy-pouch-link/src/PouchManager.spec.js
```diff
@@ -334,4 +454,63 @@ describe('PouchManager', () => {
...
+      jest
+        .spyOn(rep, 'startReplication')
+        .mockImplementation(() => Promise.resolve({}))
+      executeMock.mockRejectedValueOnce('error').mockResolvedValue({})
+
+      await manager.replicateOnce()
+      await sleep(10)
```
ptbrowne, line 511: Why do we need to sleep here ? Don't we have access to a promise that we could await instead ?

ptbrowne (approved):
Thanks !

Issue #786, opened 2020-09-17
Share code logic between StackLink and PouchLink


Crash--:
ATM, we do not share code logic between those 2 links. 

For instance we implement in both side the method to generate the name and the fields for an instance : 
https://github.com/cozy/cozy-client/blob/master/packages/cozy-pouch-link/src/CozyPouchLink.js#L284-L285

https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/DocumentCollection.js#L337-L339

And if we look closer to https://github.com/cozy/cozy-client/blob/cdf570a6dad2ec484c5a0537510bd68559272f4b/packages/cozy-stack-client/src/DocumentCollection.js#L428 we see that we're not using `indexedField` at all and it's problematic. (PR coming to fix that specific issue). 

I'm wondering if we have more logic than just "indexedField" we can share between Stack and PouchLink.

Crash--:
I like the idea submitted by @paultranvan (come on, you to read the PouchDB log to much) to create a `cozy-query-plan` module. 

Let's see how & when we can do that :)

PR #793, opened 2020-09-24
fix: Do not sort fields for index name

paultranvan:
BREAKING CHANGE: in the absence of an `indexFields` directive, the
attributes order in a `where` selector will matter to create the index.
To force an attributes order for an index, use `indexFields`.

We used to sort the index fields to build its name, e.g.
"by_date_and_type", which allows to keep the same order independently
from their position in the `where` selector.
However, this breaks the `indexedFields` behavior: because of the sort,
the fields might be indexed in a wrong order, e.g. `[type, date]` will
actually produce a `[date, type]` index. As the order is important
in Mango, this can break some queries.

We remove this sort to be consistent with
[cozy-stack-client](https://github.com/cozy/cozy-client/blob/ea768adecaf0dd1c52e9876bf68e86d099579a9b/packages/cozy-stack-client/src/DocumentCollection.js#L416-L418).

Eventually, this would be nice to refactor the code in a common lib, see https://github.com/cozy/cozy-client/issues/786

### DIFF
--- a/docs/api/cozy-client.md
+++ b/docs/api/cozy-client.md
@@ -1180,6 +1180,7 @@ from a Cozy. `QueryDefinition`s are sent to links.
 
 * [QueryDefinition](#QueryDefinition)
     * [new QueryDefinition(options)](#new_QueryDefinition_new)
+    * [.checkSortOrder(sort|selector|indexedFields)](#QueryDefinition+checkSortOrder)
     * [.getById(id)](#QueryDefinition+getById) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
     * [.getByIds(ids)](#QueryDefinition+getByIds) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
     * [.where(selector)](#QueryDefinition+where) ⇒ [<code>QueryDefinition</code>](#QueryDefinition)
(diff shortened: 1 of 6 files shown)


ptbrowne:
Cool ! How is it a breaking change though ?

paultranvan:
@ptbrowne If you have this query:
```
const query = client.find('io.cozy.banks.operations').where({date: {$gt: null}, amount: {$gt: 100}}).sortBy([{amount: 'asc', date: 'asc'}])
```
It will no longer works, because it used to create a `[amount, date]` index, but now will create a `[date, amount]` index, which won't work with pouchdb, because of the specified sort order. You will receive an error from PouchDb like `Error: Cannot sort on field(s) when using the default index`.
Now, you need to explicitly pass a `indexFields(['amount', 'date']` to be sure of the fields order in the index.

Crash--:
> Now, you need to explicitly pass a indexFields(['amount', 'date'] to be sure of the fields order in the index.

Or having the same order in the select and order, no?

paultranvan:
@Crash-- Right, I edited my comment

ptbrowne:
Mmh ok, this breaking change is problematic as we will only see the problem when working with PouchDB  right ? To help the developer with this, I think it'd be interesting to add a check in the query builder that checks if the query is valid (here that the sort is the same) after applying the sort/where ? We could emit a warning if order is not the same. In banks (and another lib I cannot remember which), we crash the tests when there is a warning, this could help the developer with the migration.

What do you think ?

paultranvan:
> we will only see the problem when working with PouchDB right?

Yes, but a badly designed CouchDB query might be run, but without actually using the index (it will return a warning: no index found). So the check you are mentioning should actually be for both couchdb and pouchdb, and should be shared between both.
But I guess we can add it just for PouchDB in the meantime

ptbrowne (approved):

ptbrowne:
💯

PR #796, opened 2020-09-25
fix: Use index for PouchDB mango queries

paultranvan:
We need to pass the index to the query so pouch knows which index to use. As a side effect, it was creating each mrview 2 times, which is very costful for instances with many documents, e.g. thousands.
Note this was already done in [stack-client](https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/DocumentCollection.js#L359), which is again related to https://github.com/cozy/cozy-client/issues/786 ;)

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -362,7 +362,11 @@ class PouchLink extends CozyLink {
         limit,
         skip
       }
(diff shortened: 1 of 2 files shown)


Crash--:
can you add a test for it?

ptbrowne (approved):

ptbrowne:
I was also going to ask for a small test :)

Crash-- (approved):

PR #797, opened 2020-09-28
Fix/operation on doctypes

Crash--:
Since fromRemote is a readOnly strategy, we don't want to apply
current mutation on it since this mutation will never be synched.

So, we forward it to the next link

### DIFF
--- a/docs/mobile-guide.md
+++ b/docs/mobile-guide.md
@@ -185,11 +185,6 @@ example, the `StackLink`).
 
 Additionally, it is possible to replicate some doctypes only in a specific direction:
 
-Since making the first query can be long (PouchDB will create the index first), you can 
-specify the queries you want to be "warmed up". It means that, those queries will be 
-executed by CozyClient during the PouchLink's instanciation, but CozyClient will use 
-PouchDB only if those queries have been resolved at least one time.
-
 ```js
 
 const pouchLink = new PouchLink({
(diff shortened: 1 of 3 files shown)


packages/cozy-pouch-link/src/CozyPouchLink.spec.js
```diff
@@ -124,6 +124,42 @@ describe('CozyPouchLink', () => {
         expect(true).toBe(true)
       })
     })
+
+    it('test if supportsOperation return false if the doctype is sync with a read only strategy', async () => {
```
ptbrowne, line 128:
```suggestion
    test('supportsOperation returns false if the doctype is sync with a read only strategy', async () => {
```

`it` and `test` are named so that we can easily read what the test is about, let's use the correct one here ;)

docs/mobile-guide.md
```diff
@@ -208,3 +203,44 @@ const pouchLink = new PouchLink({
...
+        type: 'file',
+        trashed: false,
+        updated_at: {
+          $gt: null
+        }
+      })
+  options: {
+    as: 'recent-view-query',
+    fetchPolicy: defaultFetchPolicy,
+    singleDocData: false
```
ptbrowne, line 230: Why is singleDocData set to false here ?

ptbrowne (approved):
Thanks :+1:

ptbrowne:
What was the behavior here for my culture ?

Issue #799, opened 2020-09-28
Create a static fromDOM method to cozy-client


doubleface:
This would allow to get a cozy-client instance using dataset injected by the Stack in the DOM, like cozyClient.fromEnv

ptbrowne:
We could maybe leverage / replace / reuse the loadInstanceOptionsFromDOM.

Crash--:
done in https://github.com/cozy/cozy-client/pull/800

PR #800, opened 2020-09-28
feat: Add fromDOM method

doubleface:
To get a CozyClient instance from data injected by the stack in the DOM
in the main page

fixes issue #799

### DIFF
--- a/docs/api/cozy-client.md
+++ b/docs/api/cozy-client.md
@@ -734,6 +734,7 @@ Responsible for
         * [.fromOldClient()](#CozyClient.fromOldClient)
         * [.fromOldOAuthClient()](#CozyClient.fromOldOAuthClient) ⇒ [<code>CozyClient</code>](#CozyClient)
         * [.fromEnv()](#CozyClient.fromEnv)
+        * [.fromDOM(selector, options)](#CozyClient.fromDOM) ⇒ <code>object</code>
         * [.registerHook(doctype, name, fn)](#CozyClient.registerHook)
 
 <a name="new_CozyClient_new"></a>
@@ -1147,6 +1148,20 @@ Warning: unlike other instantiators, this one needs to be awaited.
 In konnector/service context, CozyClient can be instantiated from environment variables
 
 **Kind**: static method of [<code>CozyClient</code>](#CozyClient)  
+<a name="CozyClient.fromDOM"></a>
+
+### CozyClient.fromDOM(selector, options) ⇒ <code>object</code>
+When used from an app, CozyClient can be instantiated from the data injected by the stack in
+the DOM.
+
+**Kind**: static method of [<code>CozyClient</code>](#CozyClient)  
+**Returns**: <code>object</code> - - CozyClient instance  
+
+| Param | Type | Default | Description |
+| --- | --- | --- | --- |
+| selector | <code>string</code> | <code>&quot;[role&#x3D;application]&quot;</code> | Options |
+| options | <code>object</code> |  | CozyClient constructor options |
+
 <a name="CozyClient.registerHook"></a>
 
 ### CozyClient.registerHook(doctype, name, fn)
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -153,6 +153,15 @@ describe('CozyClient initialization', () => {
     expect(client.stackClient.token.token).toBe(token)
   })
 
+  it('can be instantiated from dataset injected by the Stack in DOM', async () => {
+      const options = { cozyDomain: 'cozy.tools', cozyToken: 'abc123' }
+      const globalQuerySelectorBefore = document.querySelector
+      document.querySelector = jest.fn().mockReturnValue({ dataset: options })
```
ptbrowne, line 159: I think you want to use jest.spyOn to temporarily replace document.querySelector.

```
beforeEach(() => {
  jest.spyOn(document, 'querySelector').mockReturnValue({ dataset: options })
})

afterEach(() => {
  document.querySelector.mockRestore()
})
```

packages/cozy-client/src/CozyClient.js
```diff
@@ -239,6 +239,36 @@ class CozyClient {
...
+   */
+  static fromDOM(selector = '[role=application]', options = {}) {
+    const root = document.querySelector(selector)
+    if (!root || !root.dataset) {
+      throw new Error(
+        `Found no data in ${selector} to instantiate cozyClient`
+      )
+    }
+
+    const {cozyDomain, cozyToken} = root.dataset
```
Crash--, line 257: We should supports the both way the stack injects the content : 
- in the dataset 
- but also in JSON format in data-cozy 

```suggestion
const data = root.dataset.cozy
      ? JSON.parse(root.dataset.cozy)
      : { ...root.dataset }
const {cozyDomain, cozyToken} = data
```

docs/api/cozy-client.md
```diff
@@ -1147,6 +1148,19 @@ Warning: unlike other instantiators, this one needs to be awaited.
 In konnector/service context, CozyClient can be instantiated from environment variables
 
 **Kind**: static method of [<code>CozyClient</code>](#CozyClient)  
+<a name="CozyClient.fromDOM"></a>
+
+### CozyClient.fromDOM(selector, options) ⇒ <code>object</code>
+In cozy app context, CozyClient can be instantiated from data injected by the stack
```
ptbrowne, line 1154:
```suggestion
When used from an app, CozyClient can be instantiated from the data injected by the stack in
the DOM.
```

Crash--: When used from an web app ? ;)

packages/cozy-client/src/CozyClient.js
```diff
@@ -239,6 +239,38 @@ class CozyClient {
...
   }
 
+  /**
+   * In cozy app context, CozyClient can be instantiated from data injected by the stack
+   *
+   * @param  {string}   selector - Options
+   * @param  {object}   options  - CozyClient constructor options
+   * @returns {object} - CozyClient instance
+   */
+  static fromDOM(selector = '[role=application]', options = {}) {
```
ptbrowne, line 249: I am hesitant on the signature : I think selector will not be used as frequently as the options imho. The problem is that to define options, signature has to be passed, it is a bit inconvenient. I wonder if we should exchange the two args ? It the same problem for fromEnv though.

packages/cozy-client/src/CozyClient.js
```diff
@@ -239,6 +239,38 @@ class CozyClient {
     })
   }
 
+  /**
+   * In cozy app context, CozyClient can be instantiated from data injected by the stack
```
ptbrowne, line 243:
```suggestion
   * When used from an app, CozyClient can be instantiated from the data injected by the stack    in
   * the DOM.
```

ptbrowne (approved):

packages/cozy-client/src/CozyClient.js
doubleface: I agree, but I did it this way for consistency. 
I don't have any use case for the option parameter though.

ptbrowne: Passing the schema, the oauth options, the app name.

PR #801, opened 2020-09-30
fix: PouchLink should not forward if this is a read only operation

Crash--:
PouchLink should not forward if this is a read only operation when the
sync strategy is read only.

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.spec.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.spec.js
@@ -125,9 +125,11 @@ describe('CozyPouchLink', () => {
       })
     })
 
-    test('supportsOperation returns false if the doctype is sync with a read only strategy', async () => {
+    test('supportsOperation returns false if the doctype is synched with a read only strategy and the operation is a mutation', async () => {
       const operationTODO = {
-        doctype: TODO_DOCTYPE
+        doctype: TODO_DOCTYPE,
+        mutationType: 'CREATE_DOCUMENT',
+        document: { _type: 'io.cozy.todos' }
       }
 
       await setup({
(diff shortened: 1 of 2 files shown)


packages/cozy-pouch-link/src/CozyPouchLink.spec.js
```diff
@@ -148,17 +164,38 @@ describe('CozyPouchLink', () => {
...
         doctypesReplicationOptions: {
           'io.cozy.todos': { strategy: 'fromRemote' }
         }
       })
-      const query = Q(TODO_DOCTYPE)
-      expect.assertions(1)
-      await link.request(query, null, () => {
-        expect(true).toBe(true)
+      link.pouches.isSynced = jest.fn().mockReturnValue(true)
+      link.request(
```
ptbrowne, line 174: Nit: I am a bit surprised that there is not await here ? I think the `forward` argument should be async, so that we can put an await here.

packages/cozy-pouch-link/src/CozyPouchLink.spec.js
```diff
@@ -148,17 +164,38 @@ describe('CozyPouchLink', () => {
...
+
+    it('should not forward if the doctype is synched only for read access and the query is not a mutation', async () => {
+      await setup({
+        doctypesReplicationOptions: {
+          'io.cozy.todos': { strategy: 'fromRemote' }
+        }
+      })
+      link.pouches.isSynced = jest.fn().mockReturnValue(true)
+      const mock = jest.fn()
+      await link.request(Q(TODO_DOCTYPE), null, () => {
```
ptbrowne, line 195:
```suggestion
      await link.request(Q(TODO_DOCTYPE), null, mock)
```

?

ptbrowne (approved):

packages/cozy-pouch-link/src/CozyPouchLink.spec.js
Crash--: Since we can pass what we want in the forward argument, I do not need to await because here I want to test that the forward argument is called and all the async work in `request` is done later. But let's add the await in order to be consistent

PR #813, opened 2020-10-14
fix: Rename fetchFileContent to fetchFilecontentById

Crash--:
Deprecate fetchFileContent and replace it by fetchFileContentById()

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -494,6 +494,13 @@ class FileCollection extends DocumentCollection {
     forceFileDownload(`${href}?Dl=1`, filenameToUse)
   }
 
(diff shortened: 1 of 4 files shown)


JF-Cozy (approved):

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -494,6 +494,13 @@ class FileCollection extends DocumentCollection {
     forceFileDownload(`${href}?Dl=1`, filenameToUse)
   }
 
+  async fetchFileContent(id) {
+    console.info(
```
ptbrowne, line 498: Please use console.warn. This way if we use it in an app that has configured its tests to break on warn, they will get notified and will be able to change the call.

PR #816, opened 2020-10-20
feat: New Sharing API

Crash--:
Comment tout a commencé... 

On a récemment ajouté dans Drive / Notes (les app utilisant Cozy-Sharing), la possibilité de créer des partages cozy à cozy tout en mettant des participants à ce partage en read only. 

Donc dans la première implémentation, nous avons été assez naif. A la création d'un partage, on regardait si le membre était en read only ou pas, et on mettait les "rules" (et plus précisément la partie "policy") du partage en mode read only. 

Seulement après quand on voulait ajouter un membre à ce partage mais en read / write, ce membre se retrouvait en lecture seule... On a découvert qu'on ne pouvait pas changer la "policy" de la rules une fois le partage créé (pour des histoires de trigger côté stack). Qu'on pouvait, réduire la voilure de la "policy" en mettant le flag read_only sur le membre, mais pas la changer en elle même.

Donc on a décidé de créer par défaut le partage avec la "policy" la plus ouverte possible. Ainsi, si on veut que des membres en read_only, on ajoute ce flag à chaque membre, mais si on veut ajouter des membres en read / write, on peut aussi le faire (car la policy le permet). 

En même temps, j'en ai profité pour ajouter une nouvelle méthode create() qui colle à l'API de la stack et qui retire la notion de sharingType qui a certainement était introduite pour éviter une certaine complexité à un instant T mais qui devenait compliquée à maintenir car elle influait sur d'autres arguments.


packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -28,28 +28,54 @@ class SharingCollection extends DocumentCollection {
    * @param  {string} sharingType - If "two-way", will set the open_sharing attribute to true
    * @param  {string} description - Describes the sharing
    * @param  {string=} previewPath Relative URL of the sharings preview page
+   * @param  {boolean} open_sharing If the sharing is open or not
```
ptbrowne, line 31: I can't understand from here or from the doc what an open sharing mean :/ I checked this doc: https://docs.cozy.io/en/cozy-stack/sharing/#post-sharings

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -28,28 +28,54 @@ class SharingCollection extends DocumentCollection {
...
    */
   async share(
     document,
     recipients,
     sharingType,
     description,
-    previewPath = null
+    previewPath = null,
+    open_sharing,
+    forceOneWay = false
```
ptbrowne, line 45: I have trouble with this terminology "one-way", "two-way"; we use the read-only terminology elsewhere.

Furthermore, I think it is not very understandable to have:

```
share(
  document,
  recipients,
  'one-way',
  description,
   null,
   true,
   true
)
```

Even when using an option object, I think it is hard to understand what is the intention here:

```
share({
  document,
  recipients,
  sharingType: 'one-way',
  description,
  openSharing: true,
  forceOneWay: true
})
```

I am not sure but I think this would produce an invalid sharing ?

packages/cozy-stack-client/src/SharingCollection.js
Crash--: Not sure that the stack is the right doc to understand the details of the doctype. You can check here : https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.sharings.md

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -28,28 +28,54 @@ class SharingCollection extends DocumentCollection {
    * @param  {string} sharingType - If "two-way", will set the open_sharing attribute to true
    * @param  {string} description - Describes the sharing
    * @param  {string=} previewPath Relative URL of the sharings preview page
+   * @param  {boolean} open_sharing If the sharing is open or not
+   * @param  {boolean} forceOneWay By default, even if the sharing is `one-way`(aka read-only) we set the rules like
+   * a `two-way` (aka read/write) sharing since the stack is not able to update the rule later. By default, we still can
+   * make the sharing acting as a `one-way` for recipients by addind them to `read_only_recipients` attributes.
+   * If you really want to have a `one-way` sharing and you are sure you will not need to update it later, then you can use this
+   * `forceOneWay` attribue.
```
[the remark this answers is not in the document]
Crash--: ATM for Cozy, no. I need to check with the GL, but we will need it some day when we'll share banks accounts or else.

packages/cozy-stack-client/src/SharingCollection.js
Crash--: > I am not sure but I think this would produce an invalid sharing ?

Nop. I think we can have an openSharing with a `one-way` sharingType and a forceOneWay

Crash--:
@ptbrowne I updated (without any test or else), just to be sure we're on the same page before going to deep in ;)

packages/cozy-stack-client/src/SharingCollection.js
ptbrowne: I added a link from one doc to the other, I also found that searching with an underscore on the doc does not work correctly :/ https://github.com/cozy/cozy-stack/pull/2715

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -142,7 +172,10 @@ SharingCollection.normalizeDoctype = DocumentCollection.normalizeDoctypeJsonApi
 
 // Rules determine the behavior of the sharing when changes are made to the shared document
 // See https://github.com/cozy/cozy-stack/blob/master/docs/sharing-design.md#description-of-a-sharing
-const getSharingRules = (document, sharingType) => {
+export const getSharingRules = (document, sharingType) => {
+  console.warn(
+    'sharingType is deprecated. We now set default policy. If we need to set a custom policy use the full `rules` object of the SharingCollection.create() method'
```
ptbrowne, line 177: Since the default policy is dependent on the document_type, I think it could be useful to output it here ?

`sharingType is deprecated. The default policy for ${document._type} is now ${policy}. To set a custom policy, use the `rules` object of the SharingCollection.create method.` 

I see below that one of the cases io.cozy.files/"items" does not use getSharingPolicy ? Is it wanted ?

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -19,8 +19,48 @@ class SharingCollection extends DocumentCollection {
       data: resp.data.map(normalizeSharing)
     }
   }
+  /**
+   *
+   * Creates a new Sharing. See https://docs.cozy.io/en/cozy-stack/sharing/#post-sharings
+   *
+   * @param {object} params
+   * @param {object} params.document The document to share
+   * @param {string} params.description Description of the sharin
```
ptbrowne, line 28:
```suggestion
   * @param {string} params.description Description of the sharing
```

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -19,8 +19,48 @@ class SharingCollection extends DocumentCollection {
...
+   * @param {string} params.previewPath The path (route) to see the preview
```
ptbrowne, line 29: I don't think that adding route helps to the comprehension, an example would be more helpful IMHO.

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -19,8 +19,48 @@ class SharingCollection extends DocumentCollection {
...
+   * @param {Array} params.rules The rules defined to the sharing. See https://docs.cozy.io/en/cozy-stack/sharing-design/#description-of-a-sharing
```
ptbrowne, line 30: I think it'd be helpful to have a `@typedef` here. This way we do not have to go look at a different file, and when we'll go to typescript it will be easier to migrate automatically the jsdoc to typescript definitions.


```jsx
@typedef Rule - A sharing rule
@property ...

* @param {Array<Rule>} params.rules ....
```

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -19,8 +19,48 @@ class SharingCollection extends DocumentCollection {
...
+   * @param {Array} params.recipients Recipients to add to the sharings (will have the same permissions given by the rules defined by the sharing )
+   * @param {Array} params.readOnlyRecipients Recipients to add to the sharings with only read only access
```
ptbrowne, line 32: @typedef Recipient ?

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -19,8 +19,48 @@ class SharingCollection extends DocumentCollection {
...
+      data: {
+        type: 'io.cozy.sharings',
+        attributes: {
+          description,
+          preview_path: previewPath,
+          open_sharing: openSharing,
+          rules: rules ? rules : getSharingRules(document)
+        },
+        relationships: {
+          recipients: getRecipientsPayload(recipients),
```
ptbrowne, line 54: I think we could have:

```
{
  relationships: {
    recipients: {
      data: recipients.map(toRelationshipItem)
    }
  }
}
```

The `getRecipientsPayload` does not do anything specific for recipients and the `toRelationshipItem` can be useful elsewhere.

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -141,46 +179,85 @@ class SharingCollection extends DocumentCollection {
...
+// See https://docs.cozy.io/en/cozy-stack/sharing-design/#description-of-a-sharing
+export const getSharingRules = (document, sharingType) => {
+  if (sharingType) {
+    console.warn(
+      `sharingType is deprecated and will be removed. We now set this default policy : ${
+        isFile(document)
+          ? getSharingRulesForFile(document)
+          : getSharingRulesForPhotosAlbum(document)
+      }} \n      
+      If this default policy doesn't follow your need, we have to set a custom policy
```
ptbrowne, line 191:
```suggestion
      If this default policy doesn't follow your need, please set custom rules
```

* IMHO the use of "your" then "we" makes it confusing
* I do no think that adding the "policy" concept is useful ? Is there a difference with the rules ? It seems that a "policy" is a subset of a "rule" ? `getSharingPolicyForReferencedFiles` -> `getRuleAttributesForReferencedFiles` ?

ptbrowne (approved):
Some small remarks, LGTM.

packages/cozy-stack-client/src/SharingCollection.js
Crash--: It seems we can't import `typedef` from other file. Since `Recipient` is an `io.cozy.contacts` document I wanted to create a typedef Contact in ContactCollection or ContactModel and import it there but I can't? 

So do you have an idea of how we can re-use typedef definition ? Or do you have an advice on how to do?

packages/cozy-stack-client/src/SharingCollection.js
Crash--: 11e5f8f10d5c8ce780b964cb280488e072b44f58

packages/cozy-stack-client/src/SharingCollection.js
Crash--: 30f5bb8bd5496a226c3938fdc03fc74976fc7a79

packages/cozy-stack-client/src/SharingCollection.js
ptbrowne: It seems that importing a type is only possible in a typescript context. I have also seen https://www.npmjs.com/package/jsdoc-import-typedef that could be good to check out. In the context of this PR, I have no proper solution right now, maybe just writing `@param {Array<Recipient>}` right now is sufficient. 

For the future, I think we definitely have to check if we could use typescript tools to generate our doc instead of jsdoc.

PR #826, opened 2020-11-06
fix: Use correct var in create

paultranvan:
https://github.com/cozy/cozy-client/commit/03e46d12574b897a76af925917b8d944fec4d929 introduced a bug that makes any call to `client.create` fails with `ReferenceError: document is not defined`


JF-Cozy (approved):
nit: `const ret`shouldn't be `const res` (for result/response) ?

paultranvan:
@JF-Cozy `ret` could be shorthand for "returned value/object/whatever" so I think it's ok ;)

ptbrowne (commented):
I think the fact that this kind of changed was not catched in a test shows that a test is missing here. Can you add one ?

ptbrowne (approved):

PR #827, opened 2020-11-12
feat: Add setHasManyItems, removeHasManyItem & getHasManyItems  helpers

Merkur39:
Hello, I propose to add three static helpers for HasMany relationship
cf: #781

### DIFF
--- a/packages/cozy-client/src/associations/HasMany.js
+++ b/packages/cozy-client/src/associations/HasMany.js
@@ -4,6 +4,18 @@ import Association from './Association'
 import { QueryDefinition } from '../queries/dsl'
 import { receiveQueryResult, getDocumentFromState } from '../store'
 
+/**
+ * @typedef {object} Relationship
+ * @property {string} relName - name of the relationship
+ * @property {string} relItemId - id of the relation
+ * @property {Relation} relItemAttrs - Attributes to be set (at least _id and _type)
+ */
+/**
+ * @typedef {object} Relation
+ * @property {string} _id - id of the relation
+ * @property {string} _type - doctype of the relation
+ */
+
 const empty = () => ({
   data: [],
   next: true,
@@ -233,6 +245,10 @@ export const getHasManyItem = (HasMany.getHasManyItem = (
   return relData.find(rel => rel._id == relItemId)
 })
 
+export const getHasManyItems = (HasMany.getHasManyItems = (doc, relName) => {
+  return get(doc, `relationships.${relName}.data`, [])
+})
+
 /**
  * Sets a relationship item with the relationship name and id
  *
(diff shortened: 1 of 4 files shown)


Crash--:
@Merkur39 can you run `yarn docs` commit and push again plz ? ;)

packages/cozy-client/src/associations/HasMany.js
```diff
@@ -256,6 +272,50 @@ export const setHasManyItem = (HasMany.setHasManyItem = (
...
+  return updatedDocument
+})
+
+/**
+ * Sets multiple relationships with their respective name and ID
+ *
+ * @param {object} doc - Document to be updated
+ * @param {Array.<Relationship>} relationships - Array of relationships to be sets
+ */
+export const setHasManyItems = (HasMany.setHasManyItems = (
```
ptbrowne, line 285: I'm wondering about the utility and naming of this function. 

* It seems like the symetric setter of getHasManyItems but in fact it sets *multiple relationships* whereas getHasManyItems returns multiple *items* of a single relationship. This makes it asymetrical to getHasManyItems and this can be a source of confusion.
To me, the symetrical of getHasManyItems would be setHasManyItems(doc, relName, items).

* I think it is unnecessary to create a copy of updatedDocument to be coherent with the single setHasManyItem function.

* It creates a new API in the form of the { relName, relItemId, relItemAttrs } type. 

* I am not sure about the utility of abstracting a forEach here.

At this point, I am not sure that this function should be included.

ptbrowne (commented):
Thanks @Merkur39 ! I would be OK for getHasManyItems and removeHasManyItem but at this point, I am not sure that setHasManyItems has the right name or is necessary.

packages/cozy-client/src/associations/HasMany.spec.js
```diff
@@ -195,4 +195,62 @@ describe('HasMany static helpers', () => {
...
+    ]
+
+    const docWithRelationships = HasMany.setHasManyItems(
+      docWithoutRelationships,
+      relationships
+    )
+    expect(docWithRelationships).toMatchSnapshot()
+  })
+
+  it('should remove one relationship', () => {
```
ptbrowne, line 228:
```suggestion
  it('should remove an item from a relationship data', () => {
```

packages/cozy-client/src/associations/HasMany.js
```diff
@@ -256,6 +272,50 @@ export const setHasManyItem = (HasMany.setHasManyItem = (
...
+      item.relName,
+      item.relItemId,
+      item.relItemAttrs
+    )
+  })
+  return updatedDocument
+})
+
+/**
+ * Remove one relationship
```
ptbrowne, line 302:
```suggestion
 * Remove one relationship item
```

packages/cozy-client/src/associations/HasMany.js
Merkur39: Yes indeed, I had created it for a personal need, but I'm not sure it's pertinant to integrate it into Cozy-Client
It has no real added value

ptbrowne (approved):
Thanks !

PR #829, opened 2020-11-16
feat: Sharing get & permission check shortcut

Crash--:
- Add get on sharing doctype to be able to client.query(client.get('io.cozy.sharings', sharingId))
- Add isShortcutCreatedOnTheRecipientCozy (from Drive, but we'll need it in Notes and others apps using sharing / permission)

### DIFF
--- a/docs/api/cozy-client.md
+++ b/docs/api/cozy-client.md
@@ -326,6 +326,16 @@ we have in the store.</p>
 <dd></dd>
 <dt><a href="#PermissionItem">PermissionItem</a> : <code>object</code></dt>
 <dd></dd>
+<dt><a href="#Permission">Permission</a> : <code>object</code></dt>
+<dd><p>When a cozy to cozy sharing is created Cozy&#39;s stack creates a
+shortcut in <code>/Inbox of sharing</code> on the recipient&#39;s cozy to have a
+quick access even when the sharing is not accepted yet.</p>
+<p>However, this file is created only if the stack knows the URL of the cozy.
+This is not always the case.</p>
+<p>This method is here to tell us if the shortcut&#39;s file is created
+on the recipient&#39;s cozy. It can be used to make an UI distinction between the
+both situation.</p>
+</dd>
 <dt><a href="#HydratedQueryState">HydratedQueryState</a> ⇒ <code><a href="#HydratedQueryState">HydratedQueryState</a></code></dt>
 <dd><p>Returns the query from the store with hydrated documents.</p>
 </dd>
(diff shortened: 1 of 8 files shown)


ptbrowne:
I see that in the description, you advertise client.get but we should use Q over client.get. I realised we had not put a deprecation message there, so there they are : https://github.com/cozy/cozy-client/pull/830.

docs/api/cozy-client.md
```diff
@@ -2217,6 +2226,25 @@ Fetch and build an URL to open a note.
...
+## isShortcutCreatedOnTheRecipientCozy
+Returns if the shortcut for the sharing is created on the recipient's cozy
+
+To know if the shortcut is created on the recipient's cozy, we need to check
+the presence of the `instance` attribute on the sharing member. (since we
+are in a cozy to cozy sharing, we have only one member per "token")
+
+If this field is setted, it means the stack knows the url of the recipient's
+cozy. And if the stack knows that, we're sure that the stack has already created
+the shortcut on the recipient's cozy.
```
ptbrowne, line 2240: This doc is an explanation of the implementation. I am not sure this is the best information to put here; when reading the doc I would be more interested to know where this information (whether the shortcut is created on the recipient cozy) can be used.

s/setted/set

packages/cozy-client/src/models/permission.js
```diff
@@ -139,3 +139,26 @@ export async function isDocumentReadOnly(args) {
...
+ * the presence of the `instance` attribute on the sharing member. (since we
+ * are in a cozy to cozy sharing, we have only one member per "token")
+ *
+ * If this field is setted, it means the stack knows the url of the recipient's
+ * cozy. And if the stack knows that, we're sure that the stack has already created
+ * the shortcut on the recipient's cozy.
+ *
+ * @param {Array} included JSON API included fields. From getOwnPermissions response
+ */
+export const isShortcutCreatedOnTheRecipientCozy = included => {
```
ptbrowne, line 156: I wonder why it is "included" that is passed here instead of a shortcut document ?

docs/api/cozy-client.md
Crash--: Is it clearer now?

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -35,6 +36,19 @@ class SharingCollection extends DocumentCollection {
       data: resp.data.map(normalizeSharing)
     }
   }
+  /**
+   *
```
ptbrowne, line 40:
```suggestion
   * Fetches a sharing by id
```

packages/cozy-stack-client/src/SharingCollection.spec.js
```diff
@@ -31,6 +31,13 @@ describe('SharingCollection', () => {
   const client = new CozyStackClient()
   const collection = new SharingCollection('io.cozy.sharings', client)
 
+  describe('get', () => {
+    it('tests if it calls the right route', async () => {
```
ptbrowne, line 35:
```suggestion
    it('should call the right route', async () => {
```

"it" should refer to the thing under test, not to the test :)

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -192,7 +192,9 @@ class PermissionCollection extends DocumentCollection {
   /**
    * async getOwnPermissions - Gets the permission for the current token
```
ptbrowne, line 193:
```suggestion
   * async getOwnPermissions - Fetches permissions for the current token
```

ptbrowne (approved):
It's clearer now, thank you.

packages/cozy-stack-client/src/PermissionCollection.js
Crash--: 8571e954

Crash--: (renamed and deprecated ;))

PR #830, opened 2020-11-16
feat: Deprecate query helpers on the client

ptbrowne:
By separating query building and result fetching, it is clearer which
method build queries, and which methods retrieve results.

By deprecating those two methods, we make it clear that for query
building, Q is the recommended way, and for executing a query,
client.query is to be used.

- [x] Updated the documentation
- [x] Updated tests

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -5,6 +5,9 @@
 - [Creating a client](#creating-a-client)
 - [How to query with Cozy Client](#how-to-query-with-cozy-client)
 - [How to mutate the data](#how-to-mutate-the-data)
+- [Learn more about queries](#learn-more-about-queries)
+- [Connecting with you React components](#connecting-with-you-react-components)
+- [Different entrypoints for node/browser](#different-entrypoints-for-nodebrowser)
 
 <!-- /MarkdownTOC -->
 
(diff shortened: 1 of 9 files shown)


docs/getting-started.md
```diff
@@ -68,18 +71,19 @@ Every doctype accessed in `cozy-client` needs to be declared in the schema secti
 
 ## How to query with Cozy Client
 
-The two main methods to query data with cozyClient are `find()` and `get()`. Both take an doctype as first argument and return a `QueryDefinition` object: 
+To run a query, we first build a `QueryDefinition` with `Q` and then we execute it.
```
Crash--, line 74:
```suggestion
To run a query, we first build a `QueryDefinition` with `Q`: 
```

docs/getting-started.md
```diff
@@ -123,19 +124,11 @@ const client = new CozyClient({
...
-    const doc = data[0]
-    // modify existing io.cozy.todo
-    await client.save({...doc, checked: true})
-
-  }
-
-)
+const qdef = Q('io.cozy.todos').where({ checked: false })
+const { data: todos } = await client.query(qdef)
+const doc = data[0]
```
Crash--, line 129:
```suggestion
const doc = todos[0]
```

Crash-- (approved):
👍

PR #832, opened 2020-11-19
feat: Use node-fetch in place of isomorphic-fetch

doubleface:
We found that when compiled with webpack, isomorphic-fetch would cause
issues like :
"TypeError: realFetch.call is not a function"

Since, isomorphic-fetch is only used in node context, the isomorphic
part of this package is not needed and we can use node-fetch directly, 
which does not cause the same issue.

To be able to install node-fetch, I also had to upgrade
all babel related dependencies

also related to: This issue was fixed in connectors with https://github.com/konnectors/libs/pull/724


Crash--:
Can you add some informations in the commit message plz? In 1 year, we will remove this resolve without knowing why it was there in the first place

Crash-- (commented):
don't we have to add a dep to `node-fetch` since we're using it directly?

doubleface:
> don't we have to add a dep to `node-fetch` since we're using it directly?

Normally, we should, but this is what causes the error in the first place. This is a "quick and dirty" solution.

https://github.com/cozy/cozy-libs/pull/1164 is the associated mid-term, cleaner solution, but with a higher risk of regression I think.

Crash--:
Can we add a test for this behavior? Since the solution is not sustainable we should be sure to have something working. What do you think?

Crash-- (approved):
Looking good now! Thank you

ptbrowne:
> Can we add a test for this behavior? Since the solution is not sustainable we should be sure to have something working. What do you think?

What has been done for that ?

Crash--:
> What has been done for that ?

ATM nothing. But since the solution is much more sustainable (& logical) than the previous one, I'm ok with it (or at least having a published version without test that we can use in production since we broke few services)

PR #838, opened 2020-11-25
Add partial index support

paultranvan:
[Partial indexes](https://docs.couchdb.org/en/stable/api/database/find.html#find-partial-indexes) allows to create filter in the index definition.
This filter can be useful in several cases:
- When using `$exists: false`: the filter excludes docs without the attribute. For instance, this will be useful for this [Contacts query](https://github.com/cozy/cozy-contacts/blob/master/src/helpers/queries.js#L65-L89) as a partial index will avoid to do a full-scan of all the documents at each query (only docs without the attribute will be indexed).
- When querying non-contiguous rows. CouchDB indexes are designed to be efficient only on equality/contiguous range queries (see [here to understand why](https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/queries.js#L34-L59) Typically a `$ne` operator such as with this [Drive query](https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/queries.js#L34-L59) will be performed in-memory. A partial index can be used to exclude the doc `_id: { $ne: TRASH_DIR_ID }` from the query.

Note PouchDB does not support partial indexes yet. So, as a workaround, we force any partial index definition directly into the selector when using PouchDB.

Our [partial index](https://docs.cozy.io/en/tutorials/data/advanced/#partial-indexes) documentation should be updated afterwards


docs/api/cozy-client.md
```diff
@@ -1431,6 +1434,19 @@ See https://docs.cozy.io/en/tutorials/data/queries/#sort-data-with-mango
...
+Check the selector predicates.
+
+It is useful to warn the developer when a partial index might be used.
+
+**Kind**: instance method of [<code>QueryDefinition</code>](#QueryDefinition)  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| selector | <code>object</code> | The selector definition |
+
```
ptbrowne, line 1449: When reading this I am wondering what is returned by checkSelector ? Is it a boolean value ? Can it be precised ?

docs/api/cozy-pouch-link.md
```diff
@@ -105,6 +106,24 @@ and return if those queries are already warmedup or not
...
+performances.
+
+**Kind**: instance method of [<code>PouchLink</code>](#PouchLink)  
+**Returns**: <code>object</code> - The merge between the selector and the partial filter  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| selector | <code>object</code> | The query selector |
+| partialFilter | <code>object</code> | The partial filter |
+
```
ptbrowne, line 126: I am wondering if this should be exposed in the doc as it seems like a private method ?

packages/cozy-client/examples/albums-relationships.js
```diff
@@ -1,4 +1,8 @@
-const { QueryDefinition, HasMany, default: CozyClient } = require('../dist')
+const {
+  QueryDefinition,
+  HasMany,
+  createClientInteractive
+} = require('cozy-client')
```
ptbrowne, line 5: Mmh, I think the dist was useful while developing so that you do not have to reinstall the package ? How does cozy-client is resolved since we are in the package ? Is there a symlink from cozy-client/node_modules/cozy-client to cozy-client/packages/cozy-client/dist ?

packages/cozy-client/examples/albums-relationships.js
```diff
@@ -43,12 +47,14 @@ const schema = {
...
-  }
-  const client = new CozyClient({ uri, token, schema })
+  const client = await createClientInteractive({
+    scope: ['io.cozy.contacts'],
+    uri: process.env.COZY_URL || 'http://cozy.tools:8080',
+    schema,
+    oauth: {
+      softwareID: 'io.cozy.client.cli'
+    }
+  })
```
ptbrowne, line 57: :+1:

packages/cozy-client/examples/partial-index.js
```diff
@@ -0,0 +1,52 @@
...
+  getContactsByIndexesAttribute,
+  getContactsTest
+} = require('./contacts')
+
+global.fetch = require('node-fetch') // in the browser we have native fetch
+
+const main = async _args => {
+  const client = await createClientInteractive({
+    scope: ['io.cozy.contacts'],
+    uri: process.env.COZY_URL || 'http://cozy.tools:8080',
```
ptbrowne, line 16: Instead of process.env.COZY_URL, could it comes directly from _args ? Why is args with an underscore ?

packages/cozy-client/examples/partial-index.js
```diff
@@ -0,0 +1,52 @@
...
+  await getContactsByIndexesAttributePartialIndex(client, {
+    hasIndexes: false
+  })
+  console.timeEnd('0 contact with partial index')
+
+  const contacts = await getContactsTest(client)
+  console.log(contacts.length)
+}
+
+main(process.argv).catch(e => console.error(e))
```
ptbrowne, line 52: Please use argparse to parse argv, it is useful as it generates --help automatically.

```
import { ArgumentParser } from 'argparse'

const parser = new ArgumentParser()
parse.addArgument('--url', { defaultValue: 'http://cozy.tools:8080' })
const args = parser.parseArgs()
```

packages/cozy-client/examples/relationships.js
```diff
@@ -46,15 +46,17 @@ const schema = {
...
+    scope: [TRANSACTION_DOCTYPE, BILLS_DOCTYPE],
+    uri: process.env.COZY_URL || 'http://cozy.tools:8080',
+    schema,
+    oauth: {
+      softwareID: 'io.cozy.client.cli'
+    }
+  })
   const args = minimist(_args.slice(2), {
     string: ['selector']
   })
```
ptbrowne, line 59: Oups, minimist is used, here, if you have the time to put argparse, it would be good to replace it also here.

packages/cozy-client/src/queries/dsl.js
```diff
@@ -78,6 +80,28 @@ class QueryDefinition {
     }
   }
 
+  /**
+   * Check the selector predicates.
```
ptbrowne, line 84: s/Check/Checks

EDIT: Saw that it was at least coherent with checkSortOrder.

Nit: When reading checkSomething(), I tend to think that the result will be boolean or that the method will throw. Here we just warn in case of a problem. I checked a bit my thesaurus and could find a better verb unfortunately.

packages/cozy-client/src/queries/dsl.js
```diff
@@ -110,6 +134,7 @@ class QueryDefinition {
    */
   where(selector) {
     this.checkSortOrder({ selector })
+    this.checkSelector(selector)
```
ptbrowne, line 137: Mmh at least we are coherent with checkSortOrder 👍

ptbrowne (approved):
Looking good to me, do you have any perf reports ?

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -321,6 +321,30 @@ class PouchLink extends CozyLink {
...
+   * @param {object} partialFilter - The partial filter
+   * @returns {object} The merge between the selector and the partial filter
+   */
+  mergePartialIndexInSelector(selector, partialFilter) {
+    if (partialFilter) {
+      logger.info(
+        `PouchLink: The query contains a partial index but PouchDB does not support it. ` +
+          `Hence, the partial index definition is used in the selector for in-memory evaluation, ` +
+          `which might impact expected performances.`
+      )
```
Crash--, line 342: It's great to log it, but can the developer do something about this info? Do we recommend to not use partialFilter when using Pouch? Or do we ask the developer to donate / vote or something else?

Crash-- (approved):
Looks good! 👏

packages/cozy-client/examples/albums-relationships.js
paultranvan: Yes there is a simlink. With the dist import I had a syntax error error coming from cozy-ui on `import React from 'react';` I'm interested if you have a better workaround

packages/cozy-client/examples/partial-index.js
paultranvan: I actually didn't notice the underscore, but yes it could comes from it

packages/cozy-client/examples/partial-index.js
paultranvan: Ok, I'll update all the examples then

packages/cozy-client/src/queries/dsl.js
paultranvan: I see your point, but I guess we can rely on the jsdoc here

packages/cozy-client/src/queries/dsl.js
```diff
@@ -1,4 +1,4 @@
-const isArray = require('lodash/isArray')
+import { isArray, findKey } from 'lodash'
```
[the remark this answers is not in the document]
paultranvan: Ok, does it change something? I assumed the result was the same in the build

packages/cozy-pouch-link/src/CozyPouchLink.js
paultranvan: Ideally, for me the developer should not care about using Couch or Pouch when designing queries, so partialFilter should be used when it's possible. I guess we can suggest here the developer to at least tell PouchDB that this feature would be useful

Crash--: "plz sponsor my free time by donating on ... to implement it on PouchDB" :D

packages/cozy-client/examples/albums-relationships.js
ptbrowne: Mmh this is interesting, can you log an issue ? I think the primary reason is that the `import` syntax is not supported by node. You could for example use babel-node to run the example. Another solution would be for us to ship a version with `import` and a version with `requires` so that it can run either on node or via webpack. For information we do not ship versions with `require` since we typically bundle our libs through webpack and having imports is important to beneficiate from tree shaking.

https://webpack.js.org/guides/tree-shaking/

Another thing that is interesting is that cozy-clients depend on cozy-ui. I think this should not be the case, cozy-ui should not be in the import chain. I think this is because of useClientErrors ? I am not sure that its place should be cozy-client, do you think that we could move it to cozy-ui @Crash-- ?

Crash--: Moving it to UI, I'm not sure. Having it as a cozy-client plugin strongly agree!

paultranvan: https://github.com/cozy/cozy-client/issues/839

PR #840, opened 2020-12-07
feat: Allow registry collection fetch to have a limit parameter

doubleface:
This will allow a cleaner solution to the growing number of connectors
and apps in the store application.
The next step is to add proper pagination handling for this collection.

See https://github.com/cozy/cozy-store/pull/705#issuecomment-739244759
for more context

### DIFF
--- a/packages/cozy-client/src/registry.js
+++ b/packages/cozy-client/src/registry.js
@@ -72,13 +72,14 @@ class Registry {
    * @param  {string} params - Fetching parameters
    * @param  {string} params.type - "webapp" or "konnector"
    * @param  {string} params.channel - "dev"/"beta"/"stable"
+   * @param  {string} params.limit - maximum number of fetched apps - defaults to 200
    *
    * @returns {Array<RegistryApp>}
    */
   async fetchApps(params) {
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/registry.js
```diff
@@ -72,13 +72,14 @@ class Registry {
    * @param  {string} params - Fetching parameters
    * @param  {string} params.type - "webapp" or "konnector"
    * @param  {string} params.channel - "dev"/"beta"/"stable"
+   * @param  {string} params.limit - maximum number of fetched apps - defaults to 200
    *
    * @returns {Array<RegistryApp>}
    */
   async fetchApps(params) {
-    const { channel, type } = params
+    const { channel, type, limit = 200 } = params
```
Crash--, line 80: nit: Since we fetch the first 300 for the store. Should we upgrade our default to 300 ?

Crash-- (approved):

packages/cozy-client/src/registry.js
doubleface: I prefer to have a lower default value to not forget to implement the pagination :)

PR #852, opened 2021-01-06
feat: Support for perfs.execution_stats flag 

paultranvan:
Add  a `perfs.execution_stats` flag that enables an `execution_stats` flag for find queries, to get insights about query execution: https://docs.couchdb.org/en/2.3.1/api/database/find.html#execution-statistics

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -748,6 +748,13 @@ client.query(Q('io.cozy.bills'))`)
     try {
       this.dispatch(loadQuery(queryId))
       const response = await this.requestQuery(queryDefinition)
+      if (response.execution_stats) {
+        console.log(
+          `Execution stats for query ${queryId}: ${JSON.stringify(
+            response.execution_stats
+          )}`
+        )
+      }
       this.dispatch(
         receiveQueryResult(queryId, response, {
           update
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -137,6 +137,11 @@ Both `create()` and `save()` will return a Promise with a `data` attribute conta
 
 You can check out our dedicated [query documentation](https://docs.cozy.io/en/tutorials/data/queries/) to know more about querying with cozy-client and avoid common traps that can dramatically impact your app performances.
 
+Note you can audit your queries performance by setting a `{ perfs.execution_stats: true }` flag. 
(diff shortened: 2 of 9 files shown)


Crash--:
No console.log() with the stats?

ptbrowne:
Can you add documentation on how to use the feature ? If I understood correctly, we'll have execution stats inside the response and we should use the browser dev tools to view them ?

paultranvan:
@Crash-- I assumed the object returned in the response was enough as we can see it in the network dev tools. However, a console.log might be easier, and we could add some info, such as a warning when a full scan is detected, so I'll add this

paultranvan:
@ptbrowne Yes that what I had in mind, but I'll also add some logs (see my other comment). For the documentation, definitely, but where? Do we centralize the flags documentation?

ptbrowne:
I think putting this in the getting started guide would be good : v

packages/cozy-client/src/CozyClient.js
```diff
@@ -748,6 +748,13 @@ client.query(Q('io.cozy.bills'))`)
     try {
       this.dispatch(loadQuery(queryId))
       const response = await this.requestQuery(queryDefinition)
+      if (response.execution_stats) {
+        console.log(
+          `Execution stats for query ${queryId}: ${JSON.stringify(
+            response.execution_stats
+          )}`
+        )
+      }
```
ptbrowne, line 757: This is particular to the stack client right ? I don't think the log should be here. I wonder if it would be better suited to put this in the CozyStackLink ou CozyStackClient ?

paultranvan: I prefered to avoid CozyStackClient as it would need to add the log for all find methods. 
It is for now particular to the stack client, because pouch does not implement execution_stats, so CozyStackLink makes sense, but I assumed it could be the case in the future, hence the log here, as it could be relevant for any link

paultranvan:
> I think putting this in the getting started guide would be good : https://github.com/cozy/cozy-client/blob/master/docs/getting-started.md

Done with 6302bfe

paultranvan:
> No console.log() with the stats?

Done with e688f79

packages/cozy-client/src/CozyClient.js
ptbrowne: OK.

ptbrowne (approved):

PR #855, opened 2021-01-18
feat: Add helper to get files by qualification rules

paultranvan:
This was previously done by cozy-doctypes, but seems more relevant in cozy-client that handles the qualification model: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/administrativeProcedures/AdministrativeProcedure.js#L57

### DIFF
--- a/packages/cozy-client/src/models/file.js
+++ b/packages/cozy-client/src/models/file.js
@@ -1,6 +1,7 @@
 import get from 'lodash/get'
 import isString from 'lodash/isString'
 import { setQualification } from './document'
+import { Q } from '../queries/dsl'
 
 const FILE_TYPE = 'file'
 const DIR_TYPE = 'directory'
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/models/file.js
```diff
@@ -198,3 +199,27 @@ export const saveFileQualification = async (client, file, qualification) => {
...
 }
+
+/**
+ * Helper to query files based on qualification rules
+ *
+ * @param {object} client - The CozyClient instance
+ * @param {object} docRules - the rules containing the searched qualification and the count
+ * @returns {object} - The files found by the rules
+ */
+export const getFilesByQualificationRules = async (client, docRules) => {
```
Crash--, line 210: Let's use fetch instead of get plz

Crash-- (approved):
just a rename needed 👍

PR #856, opened 2021-01-19
fix: Do not filter on class

paultranvan:
The query was filtering 'file' to exclude directories, but it was on a wrong field: `class` is used to differentiate file type, e.g. 'pdf', 'image', etc, as stated [in the doc](https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.files.md#files).
And, in any case, this filtering is useless, as we index on the `metadata.qualification` field, which will exclude directories.

### DIFF
--- a/packages/cozy-client/src/models/file.js
+++ b/packages/cozy-client/src/models/file.js
@@ -214,7 +214,6 @@ export const fetchFilesByQualificationRules = async (client, docRules) => {
       ...rules
     })
     .partialIndex({
-      class: 'file',
       trashed: false
     })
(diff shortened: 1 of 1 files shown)


ptbrowne (approved):
Thanks !

Nit: Prefer to put the description you put in the PR inside the commit, it makes it easier to understand the change while working with `git history`. Also, here the commit does not state what has been fixed, more what has been done, it is preferable to put `fix: Fetch qualifications by rules` and put in the description the details on the implementation.

paultranvan:
@ptbrowne Totally agreed, I'll be more cautious from now ;)

Issue #719, opened 2020-06-29
Perf: Switch from pessimistic index creation to optimistic


ptbrowne:
When launching a query, we need an index. At the moment, we
try to create the index and then we make the query. In the
vast majority of cases, the index is already created so
the first query is useless. We could switch the query logic
to try the query first, and only try to create the index if
the query fails.

paultranvan:
Solved with https://github.com/cozy/cozy-client/pull/857

ptbrowne:
Hurray 🎉

PR #857, opened 2021-01-20
feat: Optimistic index creation

paultranvan:
We used to have a pessimistic index creation strategy: for each find
query, a index creation query was sent to the database, and ignored if
the index was already created.
Most of the time, these index creation queries are useless, as the index
is created when the app is first used.
Therefore, we now adopt an optimistic strategy, that is, running the
find query, and create the index afterwards only if a missing index is
returned.
It also now creates a named index, i.e. `_design/by_selector_field` and
catch index creation conflicts.

This fixes  https://github.com/cozy/cozy-client/issues/719

EDIT 04/02/20
As mentioned [here](https://github.com/cozy/cozy-client/pull/857#issuecomment-766918762), the first draft of this PR had the drawback to force the re-creation of all existing indexes. So, to overcome this, we implemented this behaviour:

- If the index is missing, query the existing design_docs and check if there is one with the same index definition
- If there is one, [copy this design doc](https://github.com/cozy/cozy-stack/blob/master/docs/data-system.md#copy-a-design-document) with the new index name. The actual index (the b-tree) won't be re-computed.
  - Delete the old design doc
- If there is none, creat the index

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -1,16 +1,16 @@
 import flag from 'cozy-flags'
 import { uri, attempt, sleep } from './utils'
 import uniq from 'lodash/uniq'
-import transform from 'lodash/transform'
-import head from 'lodash/head'
 import omit from 'lodash/omit'
+import head from 'lodash/head'
 import startsWith from 'lodash/startsWith'
 import qs from 'qs'
 
 import Collection, {
   dontThrowNotFoundError,
   isIndexNotFoundError,
-  isIndexConflictError
+  isIndexConflictError,
+  isDocumentUpdateConflict
 } from './Collection'
 import {
   getIndexNameFromFields,
@@ -168,8 +168,44 @@ class DocumentCollection {
     return this.stackClient.fetchJSON(
       'POST',
       path,
-      await this.toMangoOptions(selector, options)
+      this.toMangoOptions(selector, options)
     )
+  }
+
+  /**
+   * Handle index creation if it is missing.
+   *
+   * When an index is missing, we first check if there is one with a different
+   * name but the same definition. If yes, we copy it to re-use the same index
+   * and remove the old one.
+   * If there is none, we create the new index.
+   *
+   *
+   * @param {object} selector - The mango selector
+   * @param {object} options - The find options
+   * @private
+   */
+  async handleMissingIndex(selector, options) {
+    const { indexedFields, partialFilter } = options
+    const existingIndex = await this.findExistingIndex(selector, options)
+    const indexName = getIndexNameFromFields(indexedFields)
+    if (!existingIndex) {
+      await this.createIndex(indexedFields, {
+        partialFilter,
+        indexName
+      })
+    } else {
+      try {
+        await this.copyIndex(existingIndex, indexName)
+        await this.destroyIndex(existingIndex)
+      } catch (error) {
+        // There might be a conflict error when 2 queries using the same index
+        // are run in parallel
+        if (!isDocumentUpdateConflict(error)) {
+          throw error
+        }
+      }
+    }
   }
 
   /**
@@ -185,9 +221,9 @@ class DocumentCollection {
    * @param {object} selector - The mango selector
    * @param {object} options - The find options
    * @returns {object} - The find response
+   * @private
    */
   async findWithMango(path, selector, options = {}) {
-    const { indexedFields, partialFilter } = options
     let resp
     try {
       resp = await this.fetchDocumentsWithMango(path, selector, options)
@@ -195,10 +231,7 @@ class DocumentCollection {
       if (!isIndexNotFoundError(error)) {
         throw error
       } else {
-        await this.createIndex(indexedFields, {
-          partialFilter,
-          indexName: this.getIndexNameFromFields(indexedFields)
-        })
+        await this.handleMissingIndex(selector, options)
         resp = await this.fetchDocumentsWithMango(path, selector, options)
       }
     }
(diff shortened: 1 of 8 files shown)


packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -153,6 +157,47 @@ class DocumentCollection {
...
+
+  /**
+   * Find documents with the mango selector and create index
+   * if missing.
+   *
+   * We adopt an optimistic approach for index creation:
+   * we run the query first, and only if an index missing
+   * error is returned, the index is created and
+   * the query run again.
+   *
```
ptbrowne, line 176: Can you set `@private` ?

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -322,7 +364,7 @@ class DocumentCollection {
   }
 
   async toMangoOptions(selector, options = {}) {
```
ptbrowne, line 366: This method is now sync right ? 👏

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -99,6 +100,14 @@ class FileCollection extends DocumentCollection {
     return this.statById(id)
   }
 
+  async fetchFindFiles(selector, options) {
+    return this.stackClient.fetchJSON(
+      'POST',
+      '/files/_find',
+      await this.toMangoOptions(selector, options)
```
ptbrowne, line 107: If toMangoOptions is sync, we can remove the await here.

ptbrowne (approved):
LGTM, 👏

ptbrowne:
What are the next steps for this to be merged ?

Crash--:
There is a need to prioritize efforts because we've work to do in order to not recreate the existing indexes. We need to implement the COPY endpoint for couchdb to be able to copy the index with the new filename (stack) and use it on cozy-client side. You can have more information in the associated trello card.

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -156,6 +164,80 @@ class DocumentCollection {
...
+  }
+
+  /**
+   * Handle index creation if it is missing.
+   *
+   * When an index is missing, we first check if there is one with a different
+   * name but the same definition. If yes, we copy it to re-use the same index
+   * and remove the old one.
+   * If there is none, we create the new index.
+   *
```
ptbrowne, line 182: Two lines is too much ;)

ptbrowne (commented):
Wow quite some work there ! I've started to review, I've seen nothing blocking.

I want to re-read it tomorrow to be sure I am not missing something functionally.

Great work @paultranvan 🚀

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -168,10 +168,46 @@ class DocumentCollection {
...
+      try {
+        await this.copyIndex(existingIndex, indexName)
+        await this.destroyIndex(existingIndex)
+      } catch (error) {
+        // There might be a conflict error when 2 queries using the same index
+        // are run in parallel
+        if (!isDocumentUpdateConflict(error)) {
+          throw error
+        }
+      }
```
Crash--, line 207: I'm wondering if we should not put this work in a `migrateIndex(from, to)` method. This way, it'll be clear that we're doing this work for the purpose of a migration. 

It's clear ATM because we have the use case in mind, I don't know if in 6 months or 1 year, we'll remember that we have done that because we changed the way we named the index for the sake of the optimistic approach. 

I'm wondering if we should not explain "With the optimistic index creation approach, we are going to create new design documents." what / how is the optimistic approach in the commit description?

packages/cozy-stack-client/src/mangoIndex.js
```diff
@@ -37,3 +42,32 @@ export const getIndexFields = ({ selector, sort = [] }) => {
...
+      const partialFilterInIndex = get(
+        index,
+        `views.${ddoc}.map.partial_filter_selector`
+      )
+      if (isEqual(partialFilter, partialFilterInIndex)) {
+        return true
+      }
+    }
+    return false
+  })
```
Crash--, line 72: Is it useful to test this method?

Crash-- (commented):
Do you have an idea of the estimated time needed to migrate an index?

paultranvan:
> Do you have an idea of the estimated time needed to migrate an index?

The migration will include the following steps:

- 1st query that will fail
- Query to get all design docs
- COPY of the old design doc,  which is the time to create a regular doc - there is no index recomputation
- DELETE of the old design doc
- 2nd query on the new index (already computed)

So, I'd say about few hundreds of ms depending on the network

packages/cozy-stack-client/src/DocumentCollection.js
paultranvan: Really? It seems to me it's sometimes useful to be a bit more verbose to clarify things

packages/cozy-stack-client/src/mangoIndex.js
paultranvan: It might be, yes, I'll add it

packages/cozy-stack-client/src/DocumentCollection.js
paultranvan: Yes, it seems indeed clearer to add a `migrateIndex` method :+1:

paultranvan: Method added with dc1c684 and commit description available in 72b64bf

packages/cozy-stack-client/src/mangoIndex.js
paultranvan: There it is 484139f

Crash-- (approved):

PR #858, opened 2021-01-20
feat: Handle index conflicts

paultranvan:
Note this is temporary workaround as we realized that https://github.com/cozy/cozy-client/pull/857 might require extra work to avoid recomputing all indexes.

Sometimes, an app might run 2 concurrent queries on the same index. If
this index happens to be not created yet, it might result into a
creation conflict. In cozy-drive, this was causing an interface error,
which is a pity as this only occur for new users leading to a bad
first impression.
Here, we fix this by simply retrying the index creation when a
conflict is detected.
Furthermore, we plan to improve the index creation in a nearby
future to avoid unecessary index creation requests.


ptbrowne (approved):
👍

PR #870, opened 2021-02-08
fix: Add type to path

flohhhh:
When query definition had a type, /data/io.cozy.triggers was fetched
instead of '/jobs/triggers'.

### DIFF
--- a/packages/cozy-stack-client/src/TriggerCollection.js
+++ b/packages/cozy-stack-client/src/TriggerCollection.js
@@ -27,6 +27,22 @@ export const isForAccount = (triggerAttrs, accountId) => {
   return triggerAttrs.message.account == accountId
 }
 
+const buildParamsUrl = (worker, type) => {
+  const urlParams = new URLSearchParams()
+
+  if (worker) {
+    urlParams.set('Worker', worker)
+  }
+  if (type) {
+    if (Array.isArray(type.$in)) {
+      urlParams.set('Type', type.$in.join(','))
+    } else {
+      urlParams.set('Type', type)
+    }
+  }
+  return urlParams.toString()
+}
+
 /**
  * Implements `DocumentCollection` API along with specific methods for `io.cozy.triggers`.
  */
@@ -107,9 +123,11 @@ class TriggerCollection extends DocumentCollection {
    * @throws {FetchError}
    */
   async find(selector = {}, options = {}) {
-    if (Object.keys(selector).length > 0 && selector.worker && selector.type) {
+    const { worker, type, ...rest } = selector
+    const hasOnlyWorkerAndType = Object.keys(rest).length === 0
+    if (hasOnlyWorkerAndType) {
       // @see https://github.com/cozy/cozy-stack/blob/master/docs/jobs.md#get-jobstriggers
-      const url = `/jobs/triggers?Worker=${selector.worker}&Type=${selector.type}`
+      const url = `/jobs/triggers?${buildParamsUrl(worker, type)}`
 
       try {
         const resp = await this.stackClient.fetchJSON('GET', url)
--- a/packages/cozy-stack-client/src/TriggerCollection.js
+++ b/packages/cozy-stack-client/src/TriggerCollection.js
@@ -107,9 +107,9 @@ class TriggerCollection extends DocumentCollection {
    * @throws {FetchError}
    */
   async find(selector = {}, options = {}) {
-    if (Object.keys(selector).length === 1 && selector.worker) {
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/TriggerCollection.js
```diff
@@ -107,9 +107,9 @@ class TriggerCollection extends DocumentCollection {
    * @throws {FetchError}
    */
   async find(selector = {}, options = {}) {
-    if (Object.keys(selector).length === 1 && selector.worker) {
+    if (Object.keys(selector).length > 0 && selector.worker && selector.type) {
```
Crash--, line 110: This change is a breaking change, and I'm not sure we really want to force the use of the `type`. In some case, it is wanted to not select on the type. 

We should handle those cases: 
- only worker
- only type
- both of them 

WDYT?

ptbrowne: I agree. Moreover the stack implements a way to filter through multiple types (/jobs/triggers?Type=@cron,@webhook). Since Mango already supports this via $in, I propose that we support:

```
Q('io.cozy.triggers').where({
  type: {
    $in: ['@cron', '@webhook']
  }
})

Q('io.cozy.triggers').where({
  type: '@cron'
})

Q('io.cozy.triggers').where({
  type: '@webhook'
})
```

ptbrowne: cc @nono, are there other APIs from the stack that have the same filtering ability ? Would it make sense for those APIs to support the mango selector so that it is not done client-side ? It would make the client-side simpler.

nono: > are there other APIs from the stack that have the same filtering ability ?

I think we can find 2 or 3 endpoints with filtering like that. Scanning the full docs should answer that.

> Would it make sense for those APIs to support the mango selector so that it is not done client-side ?

Meh, I don't like that. It means creating a new endpoint with `POST`, and parsing the mango selectors, and making a response compatible with the expected format. And it is probably a good way to have silent errors (like putting a limit in the mango filter that will be ignored).

flohhhh: I made changes taking into account your comments

packages/cozy-stack-client/src/TriggerCollection.js
```diff
@@ -27,6 +27,22 @@ export const isForAccount = (triggerAttrs, accountId) => {
...
+    const suffix = type ? '&' : ''
+    url += `Worker=${worker}${suffix}`
+  }
+  if (type) {
+    if (Array.isArray(type.$in)) {
+      url += `Type=${type.$in.join(',')}`
+    } else {
+      url += `Type=${type}`
+    }
+  }
```
ptbrowne, line 42: Could the URLSearchParams API  be used instead of building the params manually ?

https://developer.mozilla.org/fr/docs/Web/API/URLSearchParams

```
sp = new URLSearchParams()
URLSearchParams {  }

sp.append('Worker', 'toto')
sp.toString()
"Worker=toto"
sp.append('Type', '@cron')
sp.toString()
"Worker=toto&Type=%40cron"
```

@nono Will the "%40" be correctly parsed by the stack ? I see that in the doc, the "@" is not url-encoded. cf https://github.com/cozy/cozy-stack/pull/2884/files

packages/cozy-stack-client/src/TriggerCollection.js
```diff
@@ -107,12 +123,17 @@ class TriggerCollection extends DocumentCollection {
    * @throws {FetchError}
    */
   async find(selector = {}, options = {}) {
-    if (Object.keys(selector).length === 1 && selector.worker) {
+    const selectorLen = Object.keys(selector).length
+    const { worker, type } = selector
+    const hasOneKey = selectorLen === 1
+    const hasWorkerAndType = worker && type && selectorLen === 2
```
ptbrowne, line 129: Mmh, I think we could simplify a bit the logic:

```
const { worker, type, ...rest } = selector
const hasOnlyWorkerAndType = Object.keys(rest).length === 0
if (hasOnlyWorkerAndType) {
  // use the /jobs/triggers URL.
}
```

ptbrowne (approved):
I'd prefer if the URLSearchParams could be used since it is more robust than manually joining parameters via & but I am not sure that the stack will parse url-encoded "@".

LGTM.

ptbrowne:
Thanks for the changes, LGTM, it seems that the PR is red because of linting issues.

PR #877, opened 2021-02-15
Fixes for optimist index

paultranvan:
This fixes 2 things introduced with #857

* For some queries, CouchDB returns a `no_usable_index` error, instead of a `no_index` error when the index is missing. 
Therefore, we handle this case to create the index after this error 
* For indexes with a partial index, we sort the fields to ease the comparison when looking for existing indexes. This sort was changing the original `indexedFields`, leading to potentially wrong index definition

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -10,6 +10,7 @@ import Collection, {
   dontThrowNotFoundError,
   isIndexNotFoundError,
   isIndexConflictError,
+  isNoUsableIndexError,
   isDocumentUpdateConflict
 } from './Collection'
 import {
@@ -241,7 +242,7 @@ class DocumentCollection {
     try {
       resp = await this.fetchDocumentsWithMango(path, selector, options)
     } catch (error) {
-      if (!isIndexNotFoundError(error)) {
+      if (!isIndexNotFoundError(error) && !isNoUsableIndexError(error)) {
         throw error
       } else {
         await this.handleMissingIndex(selector, options)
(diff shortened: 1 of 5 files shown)


packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -241,7 +242,7 @@ class DocumentCollection {
     try {
       resp = await this.fetchDocumentsWithMango(path, selector, options)
     } catch (error) {
-      if (!isIndexNotFoundError(error)) {
+      if (!isIndexNotFoundError(error) && !isNoUsableIndexError(error)) {
         throw error
       } else {
         await this.handleMissingIndex(selector, options)
```
JF-Cozy, line 248: Est-ce qu'on rajouterait pas un test ou deux pour tester ces différents cas d'erreur ?

ptbrowne:
Code looks good, I agree with @JF-Cozy that we should have tests for fix commits.

packages/cozy-stack-client/src/DocumentCollection.spec.js
```diff
@@ -512,6 +512,51 @@ describe('DocumentCollection', () => {
...
     })
 
+    it('should not throw an error if it is a missing index', async () => {
+      client.fetchJSON.mockRestore()
+      client.fetchJSON
+        .mockRejectedValueOnce(new Error('no_index'))
+        .mockReturnValueOnce(Promise.resolve({ rows: [] }))
+        .mockReturnValueOnce(Promise.resolve({}))
+        .mockReturnValueOnce(Promise.resolve(FIND_RESPONSE_FIXTURE))
+        .mockReturnValueOnce(Promise.resolve(FIND_RESPONSE_FIXTURE))
```
ptbrowne, line 522: FWIW, you can use mockResolvedValueOnce.

ptbrowne (approved):
Thanks, LGTM !

packages/cozy-stack-client/src/DocumentCollection.spec.js
paultranvan: *mind blown*

packages/cozy-stack-client/src/DocumentCollection.js
paultranvan: Absolutely: e5e7718

JF-Cozy (approved):

PR #878, opened 2021-02-19
Typescript declarations

ptbrowne:
The idea here is to use the JSDoc comments that we already have in a lot of files of cozy-client (JSDoc is used to generate API docs automatically) to generate typescript declarations and to typecheck cozy-client at transpile time.

Gains:

- More confidence for cozy-client's developer
- DX is improved for cozy-client's users (IDE help) 

<details>

<summary>Example</summary>

In:

```js
/**
 * A good dog
 */
class Dog {
  constructor() {
    this.name = 'Hello'
  }
  /**
   * Woofs !
   *
   * @param {String} msg - What to woof
   * @return {void}
   */
  woof(msg) {
    this.name = 0 // PROBLEM Changing string to number
    console.log(msg)
  }
}

const dog = new Dog()
dog.meow() // PROBLEM Absent method
dog.woof() // PROBLEM Missing argument

```

Out:

```ts
/**
 * A good dog
 */
declare class Dog {
    name: string;
    /**
     * Woofs !
     *
     * @param {String} msg - What to woof
     * @return {void}
     */
    woof(msg: string): void;
}
declare const dog: Dog;
```
</details>

It means that before publishing cozy-client, a typecheck pass will be done.

- We continue to use our existing babel transpiler
- Typescript only emits declarations (those declarations are then picked up by IDEs to show contextual helps, see screenshots below) 
  - These declarations are generated from JSDoc comments (See [1][])
  - These declarations are picked up by my editor (SublimeText) if I
  install the Typescript plugin, see screenshots below
- Checks that intra cozy-client, methods are called with the right parameters etc...

- ⚠️ To be able to typecheck, I had to install redux@4 otherwise we had
  an error, I think we should have the redux update as part of another
  PR. See for the problem https://github.com/reduxjs/redux-thunk/issues/205

- ⚠️ I had to remove some errors (13) with `@ts-ignore`. I do not think it's problematic.

### IDE screenshots

image

image

### Documentation

These two can be useful before reading the PR:
https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html
https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns

[1]: https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html

PR #881, opened 2021-02-22
feat: Add TimeSerie helpers

paultranvan:
We introduce a new doctype in Cozy to manipulate time serie ; those helpers are useful to save such data by ensuring the basic format is respected and to fetch it through a generic query. 

For the time series format, see: https://github.com/cozy/cozy-doctypes/pull/170

### DIFF
--- a/docs/api/cozy-client.md
+++ b/docs/api/cozy-client.md
@@ -220,14 +220,6 @@ example.</p>
 <dt><a href="#fetchURL">fetchURL</a> ⇒ <code>string</code></dt>
 <dd><p>Fetch and build an URL to open a note.</p>
 </dd>
-<dt><a href="#saveTimeSerie">saveTimeSerie</a></dt>
-<dd><p>Helper to save a time serie document.</p>
-</dd>
-<dt><a href="#fetchTimeSerieByIntervalAndSource">fetchTimeSerieByIntervalAndSource</a> ⇒ <code>object</code></dt>
-<dd><p>Helper to retrieve time series by their date interval and source.</p>
-<p>The starting date must be greater or equal while the ending date must
-be stricly less than the given startDate and endDate parameters.</p>
-</dd>
 <dt><a href="#triggerStates">triggerStates</a></dt>
 <dd><p>Trigger states come from /jobs/triggers</p>
 </dd>
@@ -366,6 +358,14 @@ This is not always the case.</p>
 <p>This method is here to tell us if the shortcut&#39;s file is created
 on the recipient&#39;s cozy. It can be used to make an UI distinction between the
 both situation.</p>
+</dd>
+<dt><a href="#TimeSeries">TimeSeries</a></dt>
+<dd><p>Helper to save a time series document.</p>
+</dd>
+<dt><a href="#TimeSeries">TimeSeries</a> ⇒ <code><a href="#TimeSeries">TimeSeries</a></code></dt>
+<dd><p>Helper to retrieve time series by their date interval and source.</p>
+<p>The starting date must be greater or equal while the ending date must
+be stricly less than the given startDate and endDate parameters.</p>
 </dd>
 <dt><a href="#HydratedQueryState">HydratedQueryState</a> ⇒ <code><a href="#HydratedQueryState">HydratedQueryState</a></code></dt>
 <dd><p>Returns the query from the store with hydrated documents.</p>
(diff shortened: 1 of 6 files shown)


docs/api/cozy-client.md
```diff
@@ -2349,6 +2357,34 @@ Fetch and build an URL to open a note.
 | client | <code>object</code> | CozyClient instance |
 | file | <code>object</code> | io.cozy.file object |
 
+<a name="saveTimeSerie"></a>
+
+## saveTimeSerie
```
ptbrowne, line 2362: Time series has an S at the end even in singular.

docs/api/cozy-client.md
```diff
@@ -2349,6 +2357,34 @@ Fetch and build an URL to open a note.
...
+<a name="fetchTimeSerieByIntervalAndSource"></a>
+
+## fetchTimeSerieByIntervalAndSource ⇒ <code>object</code>
+Helper to retrieve time series by their date interval and source.
+
+The starting date must be greater or equal while the ending date must
+be stricly less than the given startDate and endDate parameters.
+
+**Kind**: global constant  
+**Returns**: <code>object</code> - The time series found by the query
```
ptbrowne, line 2381: Are we sure it's not an array instead of an object ? Could we be a bit more precise on the type ? It will be useful for typechecking later.

packages/cozy-client/src/models/timeseries.js
```diff
@@ -0,0 +1,74 @@
...
+  if (!timeserie.serie || !Array.isArray(timeserie.serie)) {
+    throw new Error('You must specify a serie array for the time serie')
+  }
+}
+
+/**
+ * Helper to save a time serie document.
+ *
+ * @param {object} client - The CozyClient instance
+ * @param  {{dataType, serie, startDate, endDate, source, theme}} - The time serie attributes
```
ptbrowne, line 24: Please use a typedef to create a new type.

```
@typedef TimeSeries
@attributes dataType {String} - Blabla

@param {TimeSeries}
```

ptbrowne (approved):
Looking good, there is a problem with timeserie needing an "s" everywhere though.

Can you add an example or a guide using a timeseries in the documentation ?

Thanks

packages/cozy-client/src/models/timeseries.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+export const saveTimeSerie = async (
+  client,
+  { dataType, serie, startDate, endDate, source, theme }
+) => {
+  validateTimeSerieFormat({ dataType, serie, startDate, endDate, source })
+
+  const doctype = `io.cozy.timeseries.${dataType}`
```
Crash--, line 32: I'm not a big fan of not checking the value of `dataType`. Without this check, we can create the doctype we want starting with `io.cozy` and I don't think this is what we want at the end. 

Is it already possible to define a list of dataType or is it too soon?

paultranvan: > Without this check, we can create the doctype we want starting with io.cozy
yes, just like the others doctype

> Is it already possible to define a list of dataType
For now, there is just "geojson", but do we want to maintain a predefined list of authorized doctypes in cozy-client ? If we want to restrict the doctypes (and I agree with this part), shoudn't it be enforced by the stack?

docs/api/cozy-client.md
paultranvan: This is actually an object in JSON-API, eg
```
{
  data: [{timeserie1}, {timeserie2}]
}
```
Should we describe the data array?

ptbrowne: You could have a 

```
@typedef TimeSeries
@augments object
@property data {Array<GeoJSON>} 
```

or something like this ?

> Should we describe the data array?

We try now to have docstrings with types so that we can use typescript. So if we can use user definde types instead of only object, it will be helpful. See https://github.com/cozy/cozy-client/pull/878

docs/api/cozy-client.md
paultranvan: Good point, fixed with 58f70b7

docs/api/cozy-client.md
paultranvan: Is f1b3d28 ok?

paultranvan: Should be better with fb3bfb5

ptbrowne:
Thanks, LGTM

packages/cozy-client/src/models/timeseries.js
paultranvan: Update for this: this is something we definitely want in the future, but we merge this PR as it is for now

PR #890, opened 2021-03-02
feat: Improve types for CozyClient.fromEnv

ptbrowne:
Add types and typedefs for initialisation of CozyClient from
node environment.


Crash-- (approved):
👍

PR #902, opened 2021-03-15
Typing playground

ptbrowne:
The goal of this playground is to first explore then implement improved types
for cozy-client.

The main goal is to have the data types flow from query definition to query
results seamlessly, when someone issues an io.cozy.triggers request, we should
know that the results will be Trigger documents.

Here, the goal has been achieved inside the playground and we need to port back
the types to their respective components.

- QueryResult<Doctype>
- QueryDefinition<Doctype>
- Export typed doctypes for usage in apps
  - export <TriggerDoctype>"io.cozy.triggers"
  - export <FileDoctype>"io.cozy.files"
- Import doctypes from cozy-doctypes and transform them to types

### DIFF
--- a/packages/cozy-client/src/playground.ts
+++ b/packages/cozy-client/src/playground.ts
@@ -1,0 +1,62 @@
+import React from 'react'
+
+import useClient from './hooks/useClient'
+
+type TriggerDoctype = 'io.cozy.triggers'
+type FileDoctype = 'io.cozy.files'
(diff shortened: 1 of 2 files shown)


ptbrowne:
Je laisse ça ouvert, ca peut être intéressant pour le futur, mais je ne vais pas pouvoir travailler dessus.

rezk2ll:
@Crash--

PR #904, opened 2021-03-16
Add capabilities inside cozy-client

ptbrowne:
When instantiating a client from the DOM, the client has
the "capabilities" attributes filled.

I also changed the signature of the fromDOM call, to have
the selector in second, it is a breaking change but I checked and we do not use fromDOM much yet:

```
notes
libs
contacts
drive
store
banks
  src/ducks/client/web.js
settings
  src/lib/client.js
home
```

I'll take care of changing the call in Banks.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -1102,7 +1102,7 @@ client.query(Q('io.cozy.bills'))`)
   getAssociation(document, associationName) {
     return createAssociation(
       document,
-      this.schema.getAssociation(document._type, associationName),
+      this.schema.getRelationship(document._type, associationName),
       this.getRelationshipStoreAccessors()
     )
   }
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -1097,6 +1097,17 @@ client.query(Q('io.cozy.bills'))`)
   }
 
   /**
+   * Creates an association that is linked to the store.
+   */
+  getAssociation(document, associationName) {
+    return createAssociation(
+      document,
+      this.schema.getAssociation(document._type, associationName),
+      this.getRelationshipStoreAccessors()
+    )
+  }
+
+  /**
    * Returns the accessors that are given to the relationships for them
    * to deal with the stores.
    *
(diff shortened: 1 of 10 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -340,17 +348,18 @@ class CozyClient {
       ? JSON.parse(root.dataset.cozy)
       : { ...root.dataset }
 
-    const { cozyDomain, cozyToken } = data
+    const { domain, token } = data
```
Crash--, line 351: Hmm, this one is a BC too. To have data.domain & data.token, we need to use `cozy-data={{ }}` in the app. This is not done everywhere yet. Else we have `data-cozy-domain="{{.Domain}}"` resulting of having data.cozyDomain .

Should we check both (cozyDomain && cozyToken) || (domain && token) ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -175,6 +178,11 @@ class CozyClient {
 
     this.schema = new Schema(schema, stackClient)
 
+    /**
+     * @type {ClientCapabilities}
+     */
+    this.capabilities = capabilities || null
```
Crash--, line 184: This is great for webapp, we don't need to fetch on /capabilities now! 👍

packages/cozy-client/src/CozyClient.js
```diff
@@ -1081,6 +1081,17 @@ client.query(Q('io.cozy.bills'))`)
...
+   * Creates an association that is linked to the store.
+   */
+  getAssociation(document, associationName) {
+    return createAssociation(
+      document,
+      this.schema.getAssociation(document._type, associationName),
+      this.getRelationshipStoreAccessors()
+    )
+  }
+
```
Crash--, line 1094: Why this revert?

ptbrowne: Because the cozy bar apparently rely on getAssociation. Instead of removing it, I am fixing it in the next commit.

packages/cozy-client/src/CozyClient.js
ptbrowne: Yep ! fortunately a test also caught the error :) I added support for both cozyToken and token and same for domain.

Crash-- (approved):

Issue #910, opened 2021-03-23
could not find react-redux context value when using useQuery


drazik:
:wave:

I am trying to use the `useQuery` hook to fetch some documents, but it throws the following error:

```
Uncaught Error: could not find react-redux context value; please ensure the component is wrapped in a <Provider>
```

I have a `CozyProvider` in `src/targets/browser/index.jsx`:

```jsx
const renderApp = function(client) {
  const App = require('components/App').default
  render(
    <CozyProvider client={client}>
      <App />
    </CozyProvider>,
    document.querySelector('[role=application]')
  )
}

//...

document.addEventListener("DOMContentLoaded", () => {
  const client = CozyClient.fromDOM({ schema })
  renderApp(client)
})
```

The App component contains routing logic (`src/components/App.jsx`):

```jsx
const App = () => {
  return (
    <HashRouter>
      <Layout>
        <Switch>
          <Route path="/music" component={TracksList} />
          <Redirect from="/" to="/music" />
          <Redirect from="*" to="/music" />
        </Switch>
        <IconSprite />
      </Layout>
    </HashRouter>
  )
}
```

In the TracksList component I try to fetch `io.cozy.files` documents (`src/components/TracksList.jsx`):

```jsx
export const TracksList = () => {
  const query = Q("io.cozy.files")
  const queryResult = useQuery(query, { as: "files" })

  return (
    isQueryLoading(queryResult) ? <p>Loading...</p> : <p>Loaded</p>
  )
}
```

To check if it's specific to `useQuery`, I tried to use the `Query` component instead by doing:

```jsx
export const TracksList = () => {
  const query = Q("io.cozy.files")

  return (
    <Query query={query} as="files">
      {(queryResult) => isQueryLoading(queryResult) ? <h1>Loading...</h1> : <p>Loaded</p>}
    </Query>
  )
}
```

And in this case it works fine. It seems I'm doing something wrong with `useQuery` :thinking: But I didn't find docs about it on https://docs.cozy.io/en/cozy-client/react-integration/ and I searched for usage examples in existing cozy apps codebases but I couldn't find a meaningful difference with my code.

ptbrowne:
Ah yes, useQuery uses `useSelector from `react-redux` and you have to put a ReduxProvider. I guess we could do it by default in CozyProvider and have a prop to deactivate it if the app uses its own ReduxProvider. What do you think ?

```patch
+ import { Provider as ReduxProvider } from 'react-redux'
+    client.ensureStore()
[...]

PR #911, opened 2021-03-23
Remove need for Redux provider

ptbrowne:
The original issue is https://github.com/cozy/cozy-client/issues/910, drazik found out that without ReduxProvider, useQuery failed : it uses the default useSelector from redux that expects to have the redux context provided by redux provider.

Here, the useSelector used in useQuery is created from the client context, meaning that it works by getting the store from the cozy-client context instead of the redux context. An app is no longer required to wrap its app with ReduxProvider to use `useQuery`.

### DIFF
--- a/packages/cozy-client/src/hooks/useQuery.js
+++ b/packages/cozy-client/src/hooks/useQuery.js
@@ -1,9 +1,12 @@
 import { useEffect, useCallback } from 'react'
-import { useSelector } from 'react-redux'
+import { createSelectorHook } from 'react-redux'
 import get from 'lodash/get'
 import useClient from './useClient'
 import logger from '../logger'
 import { UseQueryReturnValue } from '../types'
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/hooks/useQuery.js
```diff
@@ -1,9 +1,12 @@
 import { useEffect, useCallback } from 'react'
-import { useSelector } from 'react-redux'
+import { createSelectorHook } from 'react-redux'
 import get from 'lodash/get'
 import useClient from './useClient'
 import logger from '../logger'
 import { UseQueryReturnValue } from '../types'
+import { clientContext } from '../context'
+
+const useSelector = createSelectorHook(clientContext)
```
Crash--, line 9: interesting, thanks!

Crash-- (approved):

Issue #839, opened 2020-11-26
cozy-ui should not be in the import chain


paultranvan:
cozy-client depends on cozy-ui, which can cause errors on `import React from 'react';` when running cozy-client build from node.
We could move the functions relying on  cozy-ui directly in cozy-ui, or use a cozy-client plugin.

See [this discussion](https://github.com/cozy/cozy-client/pull/838#discussion_r530406495)

PR #912, opened 2021-03-23
feat: Remove useClientErrors from default export

ptbrowne:
useClientErrors uses cozy-ui and cozy-ui uses @material-ui, it is
better if someone can use cozy-client without installing cozy-ui

BREAKING CHANGE: To use useClientErrors, you must import it directly:

```patch
import { useClientErrors } from 'cozy-client'
import useClientErrors from 'cozy-client/dist/hooks/useClientErrors
```

Info: I have seen this error as part of checking if it was easy to create
an app with create-react-app and introduce 'cozy-client'. Apart from
this problem, it is very easy :tada:, but the major blocker is that
the create-react-app does not write its files inside the build
folder while watching, meaning it cannot work in dev mode with the
stack.


Crash--:
fix #839

Crash-- (approved):

PR #913, opened 2021-03-26
feat: Add start for devtools

ptbrowne:
This is the devtools that has been made recently in Banks but I think it should continue its life here.

- Add a devtool panel that can be opened by clicking on a Fab button
that is only displayed when the debug flag is set to true.
- The devtools is made of several "panels". There are deffault panels
and the app can also inject its own adhoc panels.
- The "queries" panel shows the current queries inside ccozy-client
cache.
It is inspired by react-query devtools and should in the
future be improved to allow for action on queries: reset, set to empty,
delete etc... It could also show performance tips for each query (based
on the definition for example).
- The "flags" panel shows all the current flags and allow to modify them.
- The "libraries" panel show library versions based on the
global __VERSIONS__ variable that should be injected by the app. If it
is defined, the panel will be blank.


packages/cozy-client/docs/devtools.md
```diff
@@ -0,0 +1,22 @@
+## Devtools
+
+Cozy-client exposes a devtool that can be injected in an app for debug
+and better developer experience.
+
+- The devtools is made of several "panels". There are default panels and the app can also inject its own adhoc panels.
+- The "queries" panel shows the current queries inside ccozy-client cache. It is inspired by react-query devtools and should in the future be improved to allow for action on queries: reset, set to empty, delete etc... It could also show performance tips for each query (based on the definition for example).
```
Crash--, line 7: s/ccozy-client/cozy-client

packages/cozy-client/package.json
```diff
@@ -41,15 +41,18 @@
...
     "babel-plugin-search-and-replace": "1.1.0",
     "btoa": "1.2.1",
     "cozy-logger": "1.6.0",
-    "cozy-ui": "^34.0.0",
+    "cozy-ui": "48.0.0",
+    "date-fns": "2.19.0",
     "jsdoc-plugin-intersection": "^1.0.4",
     "mockdate": "^2.0.2",
     "react": "16.13.1",
+    "react-inspector": "5.1.0",
```
Crash--, line 55: Since this is not a "cozy" dep, should we add a note about that in  ./devtools.md ?

ptbrowne: I've added a note on necessary dependencies, thanks ! 

See 97383e785fb53d08634b631555eb05b292df5230

Crash-- (approved):

PR #914, opened 2021-03-29
fix: Devtools improvements

ptbrowne:
- Add how to activate VersionPlugin in Libraries panel
- Do not crash if there are no queries yet
- Use the correct object from cozy-flags

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -103,6 +103,7 @@ const DOC_UPDATE = 'update'
  * @property  {object}       [schema] - Schema description for each doctypes
  * @property  {AppMetadata}  [appMetadata] - Metadata about the application that will be used in ensureCozyMetadata
  * @property  {ClientCapabilities} [capabilities] - Capabilities sent by the stack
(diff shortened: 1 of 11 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -103,6 +103,7 @@ const DOC_UPDATE = 'update'
  * @property  {object}       [schema] - Schema description for each doctypes
  * @property  {AppMetadata}  [appMetadata] - Metadata about the application that will be used in ensureCozyMetadata
  * @property  {ClientCapabilities} [capabilities] - Capabilities sent by the stack
+ * @property  {boolean} [store] - If set to false, the client will not instantiate a Redux store automatically. Use this if you want to merge cozy-client's store with your own redux store. See [here](https://docs.cozy.io/en/cozy-client/react-integration/#1b-use-your-own-redux-store) for more information.
```
Crash--, line 106: Is is defaulted to true? 

I'm a bit surprised to see this commit in a "Devtools improvements" PR. I was not expecting a BC for this one

Crash-- (approved):

packages/cozy-client/src/CozyClient.js
ptbrowne: Store creation is not done unless it is `false`.

It is something is discussed in https://github.com/cozy/cozy-client/issues/910. I encountered a problem when using the devtools in the home where I had to manually call ensureStore(), this is why I tackled this issue here.

The breaking change is fairly small, and for an advanced use case, this is why I included it.

PR #916, opened 2021-03-30
Feat/autoname query

### LINKED ISSUE #782: Automated query id for single documents query
When a query is a get by id, it should automatically be
aliased so that every component relying on this query
relies on the same data.

See an example of usage here : https://github.com/cozy/cozy-libs/pull/1091#discussion_r486812866

Crash--:
Since the new dev tools, we can see unnamed queries. Let's fix what we can. 

The idea is the generate the `as` for a `getById()` query 

Should fix #782

### DIFF
--- a/packages/cozy-client/src/queries/dsl.js
+++ b/packages/cozy-client/src/queries/dsl.js
@@ -337,6 +337,20 @@ class QueryDefinition {
  */
 export const Q = doctype => new QueryDefinition({ doctype })
 
+/**
+ * Check if the query is a getById() query
+ *
+ * @param {QueryDefinition} queryDefinition The query definition
+ *
+ * @returns {boolean}
+ */
+export const isAGetByIdQuery = queryDefinition => {
+  // 2 attrs because we check if id and doctype are not undefined
+  return (
+    Object.values(queryDefinition).filter(attr => attr !== undefined).length ===
+      2 && queryDefinition.id !== undefined
+  )
+}
 // Mutations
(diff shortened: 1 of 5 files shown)


packages/cozy-client/src/queries/dsl.js
```diff
@@ -337,6 +337,23 @@ class QueryDefinition {
...
+ *
+ * @returns {boolean}
+ */
+export const isAGetByIdQuery = queryDefinition => {
+  if (
+    queryDefinition &&
+    queryDefinition.id &&
+    !queryDefinition.ids &&
+    !queryDefinition.selector
+  )
```
ptbrowne, line 353: Nit: Wonder why we have `selector` and not other attributes here ? I think the code works but maybe it could be more generic with `Object.keys(queryDefinition).length === 1 && queryDefinition.id !== undefined` ?

ptbrowne (approved):
:+1: Nice

packages/cozy-client/src/queries/dsl.js
Crash--: Ah nice way to detect that! I added selector since there is nothing that block us to do getById('1').where()... I just started by selector and ids and wanted to add other attributes but I forgot. 

Will give a shot to the direction you said!

Crash--: We set all the attributes to undefined even if we don't use them. So I filtered the undefined value and check if the length is 2. 👍

PR #917, opened 2021-03-30
Feat/exec time dev tools

Crash--:
Let's display the execution time in the Queries panel! 

Capture d’écran 2021-03-30 à 19 29 53

### DIFF
--- a/packages/cozy-client/src/devtools/Queries.jsx
+++ b/packages/cozy-client/src/devtools/Queries.jsx
@@ -46,7 +46,7 @@ const TableCell = withStyles({
 const getClassNameForExecutedTimesQuery = time => {
   if (time <= 100) {
     return 'u-valid u-success'
-  } else if (time > 100 && time < 500) {
+  } else if (time > 100 && time < 250) {
     return 'u-warn u-warning'
   } else {
     return 'u-danger u-error'
--- a/packages/cozy-client/src/devtools/Queries.jsx
+++ b/packages/cozy-client/src/devtools/Queries.jsx
@@ -27,7 +27,7 @@ import { NavSecondaryAction, ListGridItem } from './common'
 import PanelContent from './PanelContent'
 
 /**
- * @type {Object.<string, React.CSSProperties>}
+ * @type {object.<string, React.CSSProperties>}
  * @private
  */
 const styles = {
@@ -239,10 +239,10 @@ const QueryListItem = ({ name, selected, onClick }) => {
         primary={name}
         secondary={
           <>
-            {lastUpdate} - {queryState.data.length} docs
             {queryState.execution_stats && (
               <ExecutionTime queryState={queryState} />
             )}
+            {lastUpdate} - {queryState.data.length} docs
           </>
         }
       />
--- a/packages/cozy-client/src/devtools/Queries.jsx
+++ b/packages/cozy-client/src/devtools/Queries.jsx
@@ -115,7 +115,7 @@ const QueryData = ({ data, doctype }) => {
   return (
     <div className="u-pb-3">
       Table:{' '}
-      <input type="checkbox" onClick={handleShowTable} checked={showTable} />
+      <input type="checkbox" onChange={handleShowTable} checked={showTable} />
       <br />
       Search:{' '}
       <input
--- a/packages/cozy-client/src/devtools/Queries.jsx
+++ b/packages/cozy-client/src/devtools/Queries.jsx
@@ -260,7 +260,7 @@ const ExecutionTime = ({ queryState }) => {
     <>
       {' - '}
       <span className={classCSS}>
-        {queryState.execution_stats.execution_time_ms} ms
+        {Math.round(queryState.execution_stats.execution_time_ms)} ms
       </span>
     </>
   )
--- a/packages/cozy-client/src/devtools/Queries.jsx
+++ b/packages/cozy-client/src/devtools/Queries.jsx
@@ -43,6 +43,15 @@ const TableCell = withStyles({
   }
 })(MuiTableCell)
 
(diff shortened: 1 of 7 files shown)


packages/cozy-client/src/devtools/Queries.jsx
```diff
@@ -199,15 +239,32 @@ const QueryListItem = ({ name, selected, onClick }) => {
         primary={name}
         secondary={
           <>
-            {lastUpdate} - {queryState.data.length} docs -
-            {queryState.definition.doctype}
+            {lastUpdate} - {queryState.data.length} docs
+            {queryState.execution_stats && (
+              <ExecutionTime queryState={queryState} />
+            )}
```
ptbrowne, line 245: AMHA you can put the execution time first. lastUpdate is not very helpful.

packages/cozy-client/src/devtools/Queries.jsx
```diff
@@ -199,15 +239,32 @@ const QueryListItem = ({ name, selected, onClick }) => {
...
         }
       />
       <NavSecondaryAction />
     </ListItem>
   )
 }
+
+const ExecutionTime = ({ queryState }) => {
+  if (!queryState.execution_stats) return null
+  const classCSS = getClassNameForExecutedTimesQuery(
```
ptbrowne, line 256: Nit: className is a good name also :D

packages/cozy-client/src/devtools/index.jsx
```diff
@@ -103,7 +103,9 @@ const DevToolsPanel = props => {
           />
         </ListGridItem>
         {panels.map(panelOptions =>
-          currentPanel === panelOptions.id ? <panelOptions.Component /> : null
+          currentPanel === panelOptions.id ? (
+            <panelOptions.Component key={panelOptions.id} />
```
ptbrowne, line 107: Ah there it was, I saw a warning at some point for this key= and could not find the missing key 😅

packages/cozy-client/src/store/queries.js
```diff
@@ -7,18 +7,26 @@ import uniq from 'lodash/uniq'
...
 const INIT_QUERY = 'INIT_QUERY'
 const LOAD_QUERY = 'LOAD_QUERY'
 const RECEIVE_QUERY_RESULT = 'RECEIVE_QUERY_RESULT'
 const RECEIVE_QUERY_ERROR = 'RECEIVE_QUERY_ERROR'
 
+// Read if the devtools are open to store the execution time
+// This is done at runtime to not read the value everytime
+// we receive a result. So you have to refresh your page
+// in order to get the stats
+const executionStatsEnabled = flag('perfs.execution_stats')
```
ptbrowne, line 28: This is really not a problem to read the value everytime you receive a result ;) Flags are stored in memory so that we do not have to check localStorage for a read.

ptbrowne (approved):

ptbrowne:
:tada

packages/cozy-client/src/devtools/Queries.jsx
```diff
@@ -43,6 +43,15 @@ const TableCell = withStyles({
...
 })(MuiTableCell)
 
+const getClassNameForExecutedTimesQuery = time => {
+  if (time <= 100) {
+    return 'u-valid u-success'
+  } else if (time > 100 && time < 500) {
+    return 'u-warn u-warning'
+  } else {
+    return 'u-danger u-error'
+  }
```
ptbrowne, line 53: 500 seems already a bit high :D This execution time is not counting the network time etc.. It is only inside couchdb right ?

packages/cozy-client/src/devtools/Queries.jsx
```diff
@@ -199,15 +239,32 @@ const QueryListItem = ({ name, selected, onClick }) => {
...
+const ExecutionTime = ({ queryState }) => {
+  if (!queryState.execution_stats) return null
+  const classCSS = getClassNameForExecutedTimesQuery(
+    queryState.execution_stats.execution_time_ms
+  )
+  return (
+    <>
+      {' - '}
+      <span className={classCSS}>
+        {queryState.execution_stats.execution_time_ms} ms
```
ptbrowne, line 263: I think we can Math.round the execution time :)

packages/cozy-client/src/store/queries.js
Crash--: I don't like to put a function call in a method we use a lot. I know that modern browsers have a lot of optimizations for that specific use case but still think that if we can avoid it we should 🤷‍♂️

packages/cozy-client/src/devtools/index.jsx
Crash--: I admit, it was not the easiest to find!

packages/cozy-client/src/store/queries.js
ptbrowne: You do not like hooks ? :D

packages/cozy-client/src/devtools/Queries.jsx
Crash--: Yep this is only the time inside couchdb. OK to move it down to 250!

paultranvan: Yes it only includes couchdb time. It is high indeed, but you might have legitimate queries taking some time, for instance because you requested all the docs at once. Typically, I wouldn't be surprised that the query getting all the banks operations might take hundreds of ms for 10k+ docs :)

ptbrowne: I think we should have a warning for the query in Banks, we know this is a problem to retrieve all those docs.

paultranvan: Yes a warning looks good for such case

PR #923, opened 2021-04-01
feat: Add devtools panel resize

ptbrowne:
Hurray, we can resize the devtools panel.
Logic taken from react-query :blush:.

### DIFF
--- a/packages/cozy-client/src/devtools/index.jsx
+++ b/packages/cozy-client/src/devtools/index.jsx
@@ -5,6 +5,7 @@ import Paper from '@material-ui/core/Paper'
 import IconButton from '@material-ui/core/IconButton'
 import Grid from '@material-ui/core/Grid'
 import Box from '@material-ui/core/Box'
+import Slide from '@material-ui/core/Slide'
 import { useTheme, makeStyles } from '@material-ui/core/styles'
 
 import CozyTheme from 'cozy-ui/transpiled/react/CozyTheme'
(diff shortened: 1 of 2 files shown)


packages/cozy-client/types/types.d.ts
```diff
@@ -19,7 +19,7 @@ export type ClientResponse = any;
 export type Manifest = any;
 export type OldCozyClient = any;
 export type NodeEnvironment = any;
-export type QueryFetchStatus = "failed" | "loading" | "pending" | "loaded";
+export type QueryFetchStatus = "loading" | "loaded" | "pending" | "failed";
```
Crash--, line 22: What do you do with this file?! You don't like it? You edit it every time! 🤣

packages/cozy-client/src/devtools/index.jsx
```diff
@@ -78,35 +81,101 @@ const DevToolsNavList = ({ selected, panels, onNav }) => {
...
+    }
+
+    const unsub = () => {
+      document.removeEventListener('mousemove', run)
+      document.removeEventListener('mouseUp', unsub)
+    }
+
+    document.addEventListener('mousemove', run)
+    document.addEventListener('mouseup', unsub)
+  }
```
Crash--, line 145: I had something similar in mind, but this code is much more smaller than I excepted! Well done

Crash-- (approved):

packages/cozy-client/src/devtools/index.jsx
ptbrowne: All credits goes to the initial implementation inside react-query :)

packages/cozy-client/types/types.d.ts
ptbrowne: It's typescript that does that, it's pretty annoying.

PR #925, opened 2021-04-16
feat: Do not fetch the same query is she is already loading

Crash--:
In some case we can have the same query running almost in parallel.
In that case, we juste want only one to be executed. Others will
receive automatically the results.

To do that, I remove a previous optimisation saying that when we
INIT a query we force the fetchStatus to loading. It's true in most
case but not in the case I explained before.

So now, init only init, load only load and we can check that when we
try to load if something is already loading.


packages/cozy-client/src/CozyClient.js
```diff
@@ -804,21 +804,28 @@ client.query(Q('io.cozy.bills'))`)
...
-      const shouldFetch = options.fetchPolicy(existingQuery)
-      if (!shouldFetch) {
-        return
+    const existingQuery = this.getQueryFromState(queryId)
+    // If we already have the same query in loading state, no
+    // need to refrech it
+    if (existingQuery.fetchStatus === 'loading') {
+      return
+    }
+    if (existingQuery) {
```
taratatach, line 813: This check should be done before the `fetchStatus` one to avoid accessing an attribute of `undefined`.

JF-Cozy (approved):

PR #930, opened 2021-04-22
feat: Change files storage location on Android

JF-Cozy:
See Cordova plugin file doc for more information
https://cordova.apache.org/docs/en/7.x/reference/cordova-plugin-file/#where-to-store-files

BREAKING CHANGE: Due to Android 11 release, we have to change the files storage location.


packages/cozy-client/src/models/fsnative.js
```diff
@@ -21,26 +21,19 @@ const COZY_FILES_PATH = isIOS() ? 'CozyDrive' : 'Cozy Drive'
...
+ * Get root path
  *
  * @returns {string}
  */
-export const getRootPath = () => {
-  return isAndroidApp()
-    ? win.cordova.file.externalRootDirectory
-    : win.cordova.file.dataDirectory
-}
+export const getRootPath = () => win.cordova.file.dataDirectory
```
Crash--, line 28: I'm convinced that on Android we should use `externalDataDirectory`. 

externalDataDirectory calls `getExternalFilesDir` https://github.com/apache/cordova-plugin-file/blob/5274cab39e3f96a8682a74ba28663f32611501c8/src/android/FileUtils.java#L161 

If we read https://developer.android.com/training/data-storage/ , I assume we should use the externalDirectory of the app instead of the internal one. 

Same for the cacheDirectory

Crash-- (approved):

PR #935, opened 2021-05-04
feat: Dedupes equal queries fired at the same time

ptbrowne:
- Several components might fire the same query at the same time. In this 
case, the fetchPolicy is not used and both queries are executed. Here, we
use the queryDefinition as a key in a cache that holds pending queries. If
an existing query for a particular query definition is already pending, its
promise is returned to the caller.

- Since we now use this dedup mechanism we reverted partially #925


Crash--:
Hmm, it has been fixed by https://github.com/cozy/cozy-client/pull/925, no?

ptbrowne:
Ah OK, I was still seeing the problem in banks, and had seen your report of the problem. Cool if it's already fixed. I should have asked before to make the change.

The linked PR does have a traps though:

- if the query is already loading, nothing is returned which means we cannot rely on client.query() returning always something
- the query has to be named
- If a query is already loading and a "pagination" request is done by executing fetchMore, I think nothing will happen

Crash--:
> if the query is already loading, nothing is returned which means we cannot rely on client.query() returning always something. This is already the case for the fetch policy though, so the problem already existed before the PR.

I'm currently working on a fix for the fetchPolicy (it's just about returning queryFromState), and yesterday I talked about the bug you mentioned about client.query() to @paultranvan when fixing it. I don't want `client` to be "dependent" on a react re-render stuff. 

I like your way to fix it because:
- client.query() will still work since we can re-attache the previous request. No need to react magic 
- we can revert the revert by setting fetchStatus to loading when initQuery() 

Is there a way to avoid to dispatch the loadQuery() to avoid a redux action?

ptbrowne:
When you say "react" in your previous post, it's in relation to the redux store right ? I totally agree that having such a deep optimization for a render is too much coupling.

> Is there a way to avoid to dispatch the loadQuery() to avoid a redux action?

First, I really want to see if those actions are not batched into 1 render at React level.

Crash--:
@ptbrowne Do you want to go to the merge by yourself for this PR or do you want us to work on it?

Crash--:
I think it's good to review now cc @JF-Cozy @paultranvan @Ldoppea @flohhhh @Merkur39

packages/cozy-client/src/CozyClient.js
```diff
@@ -821,21 +831,22 @@ client.query(Q('io.cozy.bills'))`)
...
 
     try {
       this.dispatch(loadQuery(queryId))
-      const response = await this.requestQuery(queryDefinition)
+
+      const response = await this._promiseCache.exec(
+        () => this.requestQuery(queryDefinition),
+        () => stringify(queryDefinition)
+      )
+
```
ptbrowne, line 849: Below, the returned result is a response. I think it should also be made directly from the state and not be a Response object, this way, both code path return the same thing. I fear that the response and the hydrated query result are not exactly the same thing. Unfortunately, the QueryResult type here is too simple and is not protecting us.

paultranvan (approved):
Neat!

packages/cozy-client/src/CozyClient.js
Crash--: I had the same fear. Especially for relationships. After few tests, I had the same things, but you're right having only one path to return something is the way to go. 

I've reverted this modification, will create an issue and work on it later since it appears to break a lot of tests at first.

PR #946, opened 2021-05-25
Debug flag activates execution statistics

ptbrowne:
I think that it is simpler if the debug flag controlling
the display of the devtool panel also controls the execution
statistics. This way it is even simpler to see performance
execution statistics, you have nothing to do :)


Crash-- (approved):
You're right, thx!

PR #947, opened 2021-05-26
Query improvements

ptbrowne:
The [PR in banks](https://github.com/cozy/cozy-banks/pull/2107.) that switches from "all banking data being fetched" to "query only
the necessary data for each page" is related and can be interesting to read to
understand the context.

## Support for enabled parameter in useQuery/Query

Useful when you have dependent queries:

```jsx
const Component = ({ groupId } ) => {
  const { data: group} = useQuery(Q('io.cozy.bank.groups').getById(groupId), { as: `group/$groupId` })
  const { data: accounts} = useQuery(Q('io.cozy.bank.accounts').getByIds(group ? group.accounts : null), { as: `group/${groupId}/accounts`, enabled: !!group })
  return <span>{ accounts ? accounts.length : null}</span>
}
```

Here, while the group is not there, the second query will not be executed.

Very similar to what is done in react-query : https://react-query.tanstack.com/reference/useQuery#_top (ctrl+f enabled).

## Disable auto updating for some queries

Right now, queries are auto updated when new data comes in: documents that pass
the filter defined by the "selector" part of the query are inserted into the query
state (only their ids, full documents are stored in the `documents` slice)
*even if* the new data was not fetched through this query (for example through
a fetchMore).

This can be inconvenient for sorted queries on a date interval. If I query
bank transactions for March (date <= 31/03/2020) and then query those for January, I do not
want the transactions from January to go into the March query, otherwise
there can be a "gap", since February transactions have not been loaded.
This is gap is inconvenient since when I render transactions for March, it is
difficult to detect than we are missing data and that we need to `fetchMore()`
if there is already the January data to show below the March query.

When setting the auto update to false (specifically if the `add` attribute of the
autoUpdate option object is set to false), documents that are loaded for the January
query will not be inserted in the March query. Then we can detect when we have no
data left for the March query when we are at the bottom of the list and call `fetchMore`;
in a sense January data will not "disturb" the March query.

As for implementation: options passed when a query is executed are stored in
the store. This gives us the ability to check the autoUpdate option and
if set to false to forego the auto updating mechanism. The autoUpdate option
allows for granular control on the type of update that can be allowed,
if `add` is set to false, new documents that have not been fetched through
this query will not be added; but this can still be important that the documents
updates through another query also "trigger" the query so that it re-renders.
In this case, `autoOptions: { add:false, update: true }` will do the job.
Since autoUpdate options are true by default, `autoOptions: { add: false }`
also does the job here.

### Alternative

Without the option, we could have the selector for March be

`date >= 01/03/2020 && date <= 31/03/2020`

the important part here being that the data is selected both with
a lower than threshold AND a greater than threshold.
This has the disadvantage that you only query data for the particular
month and you have to change the selector for every month change when
you scroll. Dealing with month that miss data seems more difficult
whereas with the greater-than-only selector, data will come "naturally"
when doing `fetchMore` and you do not have to deal with month that
miss data.


docs/react-integration.md
```diff
@@ -131,6 +131,10 @@ function App() {
 }
 ```
 
+`useQuery` also supports a "disabled" parameter: if set to `false` the query won't
+be sent. This is useful when you have dependent queries that rely on the results
```
Crash--, line 135: I find that using `sent` is difficult to understand when we talk about a `query`. In my head, `sent` for a query, is the action to fetch the server. 

Here, we do not only fetch the server, but we do not execute anything. Maybe we should use `executed` instead of `sent`?

docs/react-integration.md
```diff
@@ -131,6 +131,10 @@ function App() {
...
+of another query to be sent.
```
Crash--, line 136: of another query before being executed?

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -219,15 +219,16 @@ class DocumentCollection {
...
     if (!existingIndex) {
       await this.createIndex(indexedFields, {
         partialFilter,
         indexName
       })
-    } else {
+    } else if (existingIndex._id !== `_design/${indexName}`) {
       await this.migrateUnamedIndex(existingIndex, indexName)
+    } else {
+      throw new Error(`Index unusable for query, index used: ${indexName}`)
```
Crash--, line 231: should be split off this commit

packages/cozy-client/src/store/queries.js
```diff
@@ -15,6 +15,13 @@ import flag from 'cozy-flags'
 import { getDocumentFromSlice } from './documents'
 import { isReceivingMutationResult } from './mutations'
 import { properId } from './helpers'
+import { isAGetByIdQuery, QueryDefinition } from '../queries/dsl'
+import { QueryState  } from '../types'
+
+/**
+ * @typedef {object} InitQueryOptions
```
Crash--: where is it used? (todo check if it's present in next commits)

packages/cozy-client/src/queries/dsl.js
```diff
@@ -150,10 +150,13 @@ class QueryDefinition {
...
    * @returns {QueryDefinition}  The QueryDefinition object.
    */
-  where(selector) {
+  where(selector, mergeWithExisting) {
     this.checkSortOrder({ selector })
     this.checkSelector(selector)
-    return new QueryDefinition({ ...this.toDefinition(), selector })
+    return new QueryDefinition({
+      ...this.toDefinition(),
+      selector: (mergeWithExisting ? Object.assign({}, this.selector, selector) : selector)
```
Crash--, line 158: I think I need more context on that one and how you consider to use it

ptbrowne: You have a query  that takes care of the relationship, of the default sort, limit by,  and of a default filtering condition. You want to add a filtering condition to the already existing filters.

```
const todoQuery = Q('io.cozy.bank.operations').where({ done: false })
const recentTodoQuery = todoQuery.where({ date: { $gt: '2021-05' } }, true)
```

Without the `true`, we lose the { done: false } selector.

I think it'd be clearer with an options argument { merge: true }.

Crash--: Remembering my young years with Doctrine ORM, it seems to me that there was the possibility to add selectors later on thanks to `andWhere` method. 

```jsx
const todoQuery = Q('io.cozy.bank.operations').where({ done: false })
const recentTodoQuery = todoQuery.andWhere({ date: { $gt: '2021-05' } })
```

What do you think? 

Other questions in mind: 
- what about the indexesFields used? Imagine : 

```
const todoQuery = Q('io.cozy.bank.operations').where({ done: false }).indexFields(['done'])
const recentTodoQuery = todoQuery.where({ date: { $gt: '2021-05' } }, true) // or todoQuery.addWhere({ate: { $gt: '2021-05' } }
```

How can we add the new `date` index fields? Are we going to add this merge/add* everywhere?

ptbrowne: Yes maybe this pattern can be used for other methods of the DSL but

* for indexes I am not sure it is really easy as you have to control precisely the index does and it can be easier to just override the existing indexes.
 * For sort I think it the same story = easier to override than to add
 * limitBy, there is no sense in merging
 * getById, no sense in merging

For andWhere vs where(..., { merge: true }), I do not think there is really a difference, the main one for me would be that it seems that the discoverability of the andWhere vs the where is lower : at some point by typing the where method in your IDE, you'll be able to discover that there is the merge feature, whereas for andWhere, I find that it adds another method. The difference is not that great though, I do not have any a big leaning against one or another. It seems the merge method is a bit more flexible, at some point you could use it to implement a more granular merge strategy by passing a function.

.where(..., { merge: existingWhere => /* do what you want */ })

Anyway, andWhere could also support this via an optional function..

What do you think, do you have any preference ?

docs/react-integration.md
ptbrowne: J'ai pas compris exactement ce que tu voulais dire, j'ai essayé de clarifier en

```patch
-be sent. This is useful when you have dependent queries that rely on the results
-of another query to be sent.
+be executed. This is useful when you have dependent queries that rely on the results
+of another query to have been fetched.
```

Qu'est ce que tu en penses ?

docs/react-integration.md
ptbrowne: Je prend en compte ce commentaire dans la discussion précédente.

packages/cozy-stack-client/src/DocumentCollection.js
ptbrowne: Done.

packages/cozy-client/src/store/queries.js
ptbrowne: I've re-ordered the commits and think this is fixed.

packages/cozy-client/src/queries/dsl.js
paultranvan: If we plan to eventually allow the merge of index/sort/... , I have a slight preference for `mergeWhere`, and eventually `mergeIndex`, etc. The "merge" terminology seems a bit more intuitive to me :)
If we do not plan this, `.where(..., { merge: true}` looks fine to me. 
This being said, the merging of index and sort might be a bit risky, as the order is quite important and might be harder to see by such method.

ptbrowne: Finally, I removed the commit introducing this option. In Banks I ended up copying the selector because when I overrided the sort + indexFields, I had warnings because of the way the checks is done.

```
const aQuery = Q('io.cozy.bank.operations').where({ done: true })
  .indexFields(['done'])
  .sortBy([{ done: 'asc' }])

const bQuery = aQuery
	.where({ done: true, date: { $gt: '2020-01' } })
    .indexFields(['date', 'done']) // warning since data and srot fields are not the same
    .sortBy([{ date: 'asc'}, { done: 'asc'}]) // at this point the selector is OK
```

If the indexFields and sortBy calls are reversed, we still have a warning since the sortBy would tell us that we need to index the fields that we are sorting.

For the future, I am wondering if the query should only be validated at execution time. cc @paultranvan

packages/cozy-client/src/types.js
```diff
@@ -47,8 +47,26 @@ import { QueryDefinition } from './queries/dsl'
...
  * @property {object|Array} data
  * @property {string} bookmark
+ * @property {QueryOptions} options
  */
 
+/**
+ * @typedef {object} AutoUpdateOptions
+ * @param {boolean} update - Should documents be updated in the query (default: true)
+ * @param {boolean} add - Should documents be added to the query (default: true)
+ * @param {boolean} remove - Should documents be added to the query (default: true)
```
Crash--, line 71: be removed

ptbrowne: Done

Crash-- (approved):
Looks good, thanks!

Crash-- (approved):
Looks good, thanks!

PR #953, opened 2021-06-04
feat: Add synchronization exclusions methods

taratatach:
We add methods to list, add and remove synchronization exclusions for
a given OAuth client.
These will be used to configure which directories should be
synchronized with a given Desktop client.

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -40,6 +40,14 @@ import { dontThrowNotFoundError } from './Collection'
  * @typedef {object} Stream
  */
 
+/**
+ * Document representing a io.cozy.oauth.clients
+ *
+ * @typedef {object} OAuthClient
+ * @property {string} _id - Id of the client
+ * @property {string} _type - Doctype of the client (i.e. io.cozy.oauth.clients)
+ */
+
 const ROOT_DIR_ID = 'io.cozy.files.root-dir'
 const CONTENT_TYPE_OCTET_STREAM = 'application/octet-stream'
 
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -891,6 +898,80 @@ class FileCollection extends DocumentCollection {
...
+    oauthClient,
+    { skip = 0, limit, cursor } = {}
+  ) {
+    const params = {
+      include: 'files',
+      'page[limit]': limit,
+      'page[cursor]': cursor,
+      sort: 'id'
+    }
+    const url = uri`/data/${oauthClient._type}/${oauthClient._id}/relationships/not_synchronizing`
```
Crash--, line 922: In fact, since we use the `/data/io.cozy.oauth.clients` route, shouldn't we create an `OAuthClients` collection?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -891,6 +898,80 @@ class FileCollection extends DocumentCollection {
...
+   * @param  {OAuthClient} oauthClient  A JSON representing the OAuth client
+   * @param  {Array}  directories       An array of JSON documents having a `_type` and `_id` field and representing directories.
+   * @returns {object}                  The JSON API conformant response.
+   */
+  removeNotSynchronizedDirectories(oauthClient, directories) {
+    const refs = directories.map(d => ({ id: d._id, type: d._type }))
+    return this.stackClient.fetchJSON(
+      'DELETE',
+      uri`/data/${oauthClient._type}/${oauthClient._id}/relationships/not_synchronizing`,
+      { data: refs }
```
Crash--, line 972: no need to normalizeFile ?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -891,6 +898,80 @@ class FileCollection extends DocumentCollection {
...
+   * @param  {OAuthClient} oauthClient  A JSON representing the OAuth client
+   * @param  {Array}  directories       An array of JSON documents having a `_type` and `_id` fields and representing directories.
+   * @returns {object}                  The JSON API conformant response.
+   */
+  addNotSynchronizedDirectories(oauthClient, directories) {
+    const refs = directories.map(d => ({ id: d._id, type: d._type }))
+    return this.stackClient.fetchJSON(
+      'POST',
+      uri`/data/${oauthClient._type}/${oauthClient._id}/relationships/not_synchronizing`,
+      { data: refs }
```
Crash--, line 951: no need to normalizeFile the response?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -891,6 +898,80 @@ class FileCollection extends DocumentCollection {
...
+   *  Add directory synchronization exclusions to an OAuth client — see https://docs.cozy.io/en/cozy-stack/not-synchronized-vfs/#post-datatypedoc-idrelationshipsnot_synchronizing
+   *
+   *  For example, to exclude directory `/Photos` from `My Computer`'s desktop synchronization:
+   * ```
+   * addNotSynchronizedDirectories({_id: 123, _type: "io.cozy.oauth.clients", clientName: "Cozy Drive (My Computer)", clientKind: "desktop"}, [{_id: 456, _type: "io.cozy.files", name: "Photos", path: "/Photos"}])
+   * ```
+   *
+   * @param  {OAuthClient} oauthClient  A JSON representing the OAuth client
+   * @param  {Array}  directories       An array of JSON documents having a `_type` and `_id` fields and representing directories.
+   * @returns {object}                  The JSON API conformant response.
```
Crash--, line 944: does it return an `IOCozyFile` document? ;)

packages/cozy-stack-client/src/FileCollection.js
taratatach: I thought about it but since this is related to files synchronization and I saw the `references` methods which do not pertain only to files either I finally decided to add it to the existing collection.

I'm OK with moving it out as well. I'll listen to your advice on this one as I don't tinker around Cozy Client that much :).

packages/cozy-stack-client/src/FileCollection.js
taratatach: Well we're not getting files back so it's probably not a good idea?

packages/cozy-stack-client/src/FileCollection.js
taratatach: See https://github.com/cozy/cozy-client/pull/953#discussion_r646556452

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -40,6 +40,13 @@ import { dontThrowNotFoundError } from './Collection'
  * @typedef {object} Stream
  */
 
+/**
+ * Document representing a io.cozy.oauth.clients
+ *
+ * @typedef {object} OAuthClient
+ * @property {string} _id - Id of the client
```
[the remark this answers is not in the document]
taratatach: Right this is probably a good idea. I based this type on the `FileDocument` definition above and didn't think about it that much.
But since I'm actually using the `_type` attribute in the methods it should appear here.

packages/cozy-stack-client/src/FileCollection.js
Crash--: I was tricked by "   * @returns {object}                  The JSON API conformant response." 

Maybe you can say it returns a 204 No-Content?

packages/cozy-stack-client/src/FileCollection.js
Crash--: > as I don't tinker around Cozy Client that much :)

Shame on you! You should

Crash-- (approved):

packages/cozy-stack-client/src/FileCollection.js
Crash--: Let's leave the functions here. We need to work on others issues before handling correctly that case!

PR #955, opened 2021-06-07
feat: Add job setState

doubleface:
This allows to set a job's state.

This will only work for jobs of @client type. For this type of jobs, the client side launcher needs to set the result of the job itself since it is not launched in the stack's
environment.

### DIFF
--- a/packages/cozy-stack-client/src/JobCollection.js
+++ b/packages/cozy-stack-client/src/JobCollection.js
@@ -75,6 +75,10 @@ class JobCollection {
    * @param {JobDocument} job - io.cozy.jobs document
    */
   async update(job) {
+    if (job.worker !== 'client') {
+      throw new Error('JobCollection.update only works for client workers')
+    }
+
     return this.stackClient.fetchJSON('PATCH', `/jobs/${job._id}`, {
       data: {
         type: JOBS_DOCTYPE,
--- a/packages/cozy-stack-client/src/JobCollection.js
+++ b/packages/cozy-stack-client/src/JobCollection.js
@@ -76,7 +76,9 @@ class JobCollection {
    */
   async update(job) {
     if (job.worker !== 'client') {
-      throw new Error('JobCollection.update only works for client workers')
+      throw new Error(
+        `JobCollection.update only works for client workers. ${job.worker} given`
+      )
     }
 
     return this.stackClient.fetchJSON('PATCH', `/jobs/${job._id}`, {
--- a/packages/cozy-stack-client/src/JobCollection.js
+++ b/packages/cozy-stack-client/src/JobCollection.js
@@ -56,6 +56,28 @@ class JobCollection {
   }
 
   /**
+   * Set the job's state
+   *
+   * @param {string} id: job's id
+   * @param {boolean} result: job's result
+   * @param {string} error: error message if any
+   *
+   * This does work only for jobs comming from @client triggers
+   */
+  async setState(id, result, error) {
+    return this.stackClient.fetchJSON('PATCH', `/jobs/${id}`, {
+      data: {
+        type: JOBS_DOCTYPE,
+        id,
+        attributes: {
+          state: result ? 'done' : 'errored',
+          error
+        }
+      }
+    })
+  }
+
+  /**
    * Polls a job state until it is finished
    *
    * `options.until` can be used to tweak when to stop waiting. It will be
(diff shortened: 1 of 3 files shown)


JF-Cozy (approved):

packages/cozy-stack-client/src/JobCollection.js
```diff
@@ -55,6 +55,28 @@ class JobCollection {
...
+      data: {
+        type: JOBS_DOCTYPE,
+        id,
+        attributes: {
+          state: result ? 'done' : 'errored',
+          error
+        }
+      }
+    })
+  }
```
Crash--, line 78: We should rely on the CozyClient DSL to be able to write: 

```jsx 
const existingJob = await client.query(Q('io.cozy.jobs').getById('1'));
const updatedJob = {
    ...existingJob,
    attributes: {
      ...existingJob.attributes,
      state: 'MyState',
      error: MyError
    }
}
client.save(updatedJob)
```

We should not create custom function in the collection, we should not bypass our CC mechanisms. Or at least, we should try to avoid it as much as we can. For a simple update, it's easy.

Here is the path: 
client.save => https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L592-L596

getDocumentSavePlan => https://github.com/cozy/cozy-client/blob/6b59f8d6752d44afb8a10827af2cd2a9ca151d72/packages/cozy-client/src/CozyClient.js#L678

will make the call to create or update : https://github.com/cozy/cozy-client/blob/6b59f8d6752d44afb8a10827af2cd2a9ca151d72/packages/cozy-client/src/CozyClient.js#L684-L685 

updateDocument will send an `UPDATE_DOCUMENT` action: https://github.com/cozy/cozy-client/blob/90c7abb13fd71267cb44a1af6bca7b03b03a7645/packages/cozy-client/src/queries/dsl.js#L372

Links (in that case StackClient but can also be PouchLink) will call the update() method of the collection: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/StackLink.js#L66-L69

To resume: 
This method should be called `update` and accept an io.cozy.jobs as argument.

Crash-- (changes requested):
see my comment

packages/cozy-stack-client/src/JobCollection.js
doubleface: Thanks for the detailed comment

packages/cozy-stack-client/src/JobCollection.js
```diff
@@ -48,13 +57,36 @@ class JobCollection {
...
+      data: {
+        type: JOBS_DOCTYPE,
+        id: job._id,
+        attributes: {
+          state: job.attributes.state,
+          error: job.attributes.error
+        }
+      }
+    })
+  }
```
Crash--, line 88: the response should be normalized

packages/cozy-stack-client/src/JobCollection.js
```diff
@@ -48,13 +57,36 @@ class JobCollection {
...
     return Collection.get(this.stackClient, uri`/jobs/${id}`, {
       normalize: normalizeJob
     })
   }
 
+  /**
+   * Update the job's state
+   * This does work only for jobs comming from @client triggers
+   *
+   * @param {JobDocument} job - io.cozy.jobs document
```
Crash--, line 75:
```suggestion
   * @returns {JobDocument} 
```

packages/cozy-stack-client/src/JobCollection.js
```diff
@@ -48,13 +57,36 @@ class JobCollection {
...
+   */
+  async update(job) {
```
Crash--, line 77: I'm wondering if this method should not check if `job.attributes.worker === client` before making the call. Like that, it's explicit that only this worker is supported. Should throw an error if not. WDYT?

Crash-- (approved):
2 comments, but ok to merge after that :)

packages/cozy-stack-client/src/JobCollection.js
doubleface: I wondered too, but chose to let this check to the stack. This avoids some later error if the stack changes.
But since you asked, I tested the stack behavior and it sends a `403 - {"error":"Forbidden"}`

Let's be more explicit and throw with a more explicit message

PR #956, opened 2021-06-07
fix: Import react native in auth only when using it

taratatach:
Some clients like Cozy Desktop don't use React Native and thus don't
include it as part of their dependencies.

This means that Cozy Client can't be imported anymore on those clients
now that React Native is a mandatory peer dependency and trying to do
so results in a import error.

### DIFF
--- a/packages/cozy-client/src/authentication/mobile.js
+++ b/packages/cozy-client/src/authentication/mobile.js
@@ -6,10 +6,6 @@ import {
   isIOSApp
 } from 'cozy-device-helper'
 import { CordovaWindow } from '../types'
-//@ts-ignore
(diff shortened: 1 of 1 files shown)


taratatach:
I'm using [Dynamic Imports](https://v8.dev/features/dynamic-import#dynamic) to solve the import error on Cozy Desktop.

doubleface (approved):
I see we have the same problem for connectors anyway. Thanks @taratatach

Crash--:
@doubleface is it supported by our node version?

doubleface:
I supposed babel and webpack would do the trick. It works for the app, but not for the connectors. Webpack still tries to import react native dependencies. I'm working on it

PR #957, opened 2021-06-08
feat: Use native.js file to make react native really optional

doubleface:
Tested with the react native application
The native.js file will be ignored in other environment and shall cause no regression

This PR is made to replace https://github.com/cozy/cozy-client/pull/956 which did not work for connectors environment.

### DIFF
--- a/packages/cozy-client/src/authentication/mobile.native.js
+++ b/packages/cozy-client/src/authentication/mobile.native.js
@@ -1,4 +1,6 @@
+//@ts-ignore
 import { InAppBrowser } from 'react-native-inappbrowser-reborn'
+//@ts-ignore
 import { Linking } from 'react-native'
 
 /**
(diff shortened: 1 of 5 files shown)


packages/cozy-client/src/authentication/mobile.native.js
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+
+    Linking.addEventListener('url', linkListener)
+  })
+}
+
+/**
+ * Return the method to use for
+ * authentication based on the
```
JF-Cozy, line 60: `authentication based on the` ... suspens ? :)

JF-Cozy: Est-ce qu'il y a déjà des tests associés ?

Crash--: C'est du hot fix et difficilement testable puisque du build grosso-modo

taratatach (approved):
Works on Cozy Desktop :+1:

PR #958, opened 2021-06-09
fix: Moved polyfills in createClientInteractive

JulienMirval:
Running a service using cozy-konnector-dev fails as it tries to authenticate first, calling btoa, which was not available in this context.

Polyfills for fetch and btoa have been moved inside createClientInteractive, as putting them at the top of the file would have caused to many side effects.

### DIFF
--- a/packages/cozy-client/src/cli/index.js
+++ b/packages/cozy-client/src/cli/index.js
@@ -181,6 +181,9 @@ const readJSON = (fs, filename) => {
  * @returns {Promise<CozyClient>} - A client that is ready with a token
  */
 const createClientInteractive = (clientOptions, serverOpts) => {
+  global.fetch = nodeFetch
+  global.btoa = btoa
+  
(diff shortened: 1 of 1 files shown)


Crash-- (approved):

ptbrowne:
Since this is invasive, should it be possible to deactivate polyfilling via options in serverOpts ?

```
if (serverOpts.polyfills != false) {
   // polyfills
}
```

I think that in general, a library should not be responsible for polyfills. Since this is cozy-konnector-dev that needs btoa, why don't we change cozy-konnector-dev to require btoa ?

Crash--:
@ptbrowne You respond before I had the time to write the issue. Let's discuss there #959

ptbrowne:
It seems like in a library it would be more appropriate to not polyfill and only import the functions.

```
+ const fetch = require('isomorphic-fetch') // or cross-fetch that seems more recent https://github.com/lquixada/cross-fetch
fetch()
```

PR #963, opened 2021-06-15
feat: Allow to pass null to instantiate a query

ptbrowne:
During tests, it can be useful to instantiate a query with null to test edge cases, for example testing a component to see that it does not crash while the query has not been loaded yet.

EDIT: better description


flohhhh:
do you have an example in mind?

ptbrowne:
Yep,  for example testing a component to see that it does not crash while the query has not been loaded yet.

flohhhh:
Thx

flohhhh (approved):

PR #964, opened 2021-06-15
fix: improve types for cozy pass web

Ldoppea:
Those modifications are needed to make `cozy-pass-web` compilation work as this project is in Typescript.

Two modifications has to be discussed : 

- `timeseries.js` : TS was unable to type `fetchTimeSeriesByIntervalAndSource` arguments.
  - I removed deconstruction from parameters so it is easier to add type to it. But maybe there is another way do document types for destructured params?
  - I typed it with `TimeSeries` and I added the optional `limit` member to existing `@typedef`. I don't have enough knowledge on this project to tell if it should has its own type instead of reusing `TimeSeries`.
- `Association.js` : this class is inherited by `HasOneInPLace`, `HasOne`, `HasManyInPlace`, and `HasMany`.
  - Each of those children override `query()` but with 3 parameters.
  - Typescript complains about this signature difference
  - I added those 3 parameters to the base class method. But maybe I should have create a new base method with those parameters leaving the original one unchanged?

### DIFF
--- a/packages/cozy-client/src/associations/Association.js
+++ b/packages/cozy-client/src/associations/Association.js
@@ -1,4 +1,5 @@
-import { QueryState } from '../types'
+import CozyClient from '../CozyClient'
+import { CozyClientDocument } from '../types'
 import { QueryDefinition } from '../queries/dsl'
 
 /**
@@ -223,7 +224,11 @@ class Association {
   /**
    * Derived `Association`s need to implement this method.
    *
-   * @returns {QueryDefinition | QueryState}
+   * @param {CozyClientDocument} document - Document to query
+   * @param {CozyClient} client - The CozyClient instance
+   * @param {Association} assoc - Association containing info on how to build the query to fetch related documents
+   *
+   * @returns {CozyClientDocument | QueryDefinition }
    */
   static query(document, client, assoc) {
     throw new Error('A custom relationship must define its query() function')
--- a/packages/cozy-client/src/associations/Association.js
+++ b/packages/cozy-client/src/associations/Association.js
@@ -225,7 +225,7 @@ class Association {
    *
    * @returns {QueryDefinition | QueryState}
    */
-  static query() {
+  static query(document, client, assoc) {
     throw new Error('A custom relationship must define its query() function')
   }
 }
--- a/packages/cozy-client/src/associations/Association.js
+++ b/packages/cozy-client/src/associations/Association.js
@@ -1,4 +1,3 @@
-import CozyClient from '../CozyClient'
 import { CozyClientDocument } from '../types'
 import { QueryDefinition } from '../queries/dsl'
 
@@ -225,7 +224,7 @@ class Association {
    * Derived `Association`s need to implement this method.
    *
    * @param {CozyClientDocument} document - Document to query
-   * @param {CozyClient} client - The CozyClient instance
+   * @param {object} client - The CozyClient instance
    * @param {Association} assoc - Association containing info on how to build the query to fetch related documents
    *
    * @returns {CozyClientDocument | QueryDefinition }
--- a/packages/cozy-client/src/models/timeseries.js
+++ b/packages/cozy-client/src/models/timeseries.js
@@ -64,7 +64,12 @@ export const saveTimeSeries = async (client, timeseriesOption) => {
  * Helper to retrieve time series by their date interval and source.
  *
  * @param {object} client - The CozyClient instance
- * @param {{ startDate, endDate, dataType, source, limit }} The query params.
+ * @param {object} params - The query params
+ * @param {Date} params.startDate - The starting date of the series
+ * @param {Date} params.endDate - The end date of the series
+ * @param {String} params.dataType - The type of time series, e.g. 'electricity'
+ * @param {String} params.source - The data source, e.g. 'enedis.fr'
+ * @param {number} params.limit - Number of serie items to retrieve
  *
  * @typedef TimeSeriesJSONAPI
  * @property data {Array<TimeSeries>} - The JSON-API data response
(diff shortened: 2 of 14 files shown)


packages/cozy-client/src/models/timeseries.js
```diff
@@ -64,16 +65,17 @@ export const saveTimeSeries = async (client, timeseriesOption) => {
  * Helper to retrieve time series by their date interval and source.
  *
  * @param {object} client - The CozyClient instance
- * @param {{ startDate, endDate, dataType, source, limit }} The query params.
+ * @param {TimeSeries} timeSeries - The query params.
```
Crash--, line 68: not sure this is a TimeSeries object. cc @paultranvan we need you here ;)

packages/cozy-client/src/models/timeseries.js
```diff
@@ -64,16 +65,17 @@ export const saveTimeSeries = async (client, timeseriesOption) => {
...
  *
  * @typedef TimeSeriesJSONAPI
  * @property data {Array<TimeSeries>} - The JSON-API data response
  * @returns {Promise<TimeSeriesJSONAPI>} The TimeSeries found by the query in JSON-API format
  */
 export const fetchTimeSeriesByIntervalAndSource = async (
   client,
-  { startDate, endDate, dataType, source, limit }
+  timeSeries
```
Crash--, line 76: To describe destructured object, you can : 
```
/**
@param {string} arg1
@param {string} options.arg2
@param {string} options.arg3

const func(arg1, { arg2, arg3 }) 

**/

```

It should make the trick. See https://jsdoc.app/tags-param.html#parameters-with-properties

packages/cozy-client/src/associations/Association.js
```diff
@@ -225,7 +225,7 @@ class Association {
    *
    * @returns {QueryDefinition | QueryState}
    */
-  static query() {
+  static query(document, client, assoc) {
```
ptbrowne, line 228: Those parameters can be typed:

client is a CozyClient
document is a CozyClientDocument
assoc is an Association I guess

packages/cozy-client/types/associations/Association.d.ts
```diff
@@ -77,7 +77,7 @@ declare class Association {
      *
      * @returns {QueryDefinition | QueryState}
      */
-    static query(): QueryDefinition | QueryState;
+    static query(document: any, client: any, assoc: any): QueryDefinition | QueryState;
```
Crash--, line 80: you can type them. 
document can be CozyClientDocument
client should be CozyClient 
it seems we have an `Association` type also

packages/cozy-client/src/models/timeseries.js
paultranvan: No, those are parameters to find a timeserie: https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.timeseries.md

packages/cozy-client/src/associations/Association.js
Ldoppea: Ok I just made this.
Note that I had to propagate those definitions to those so TS is able to detect the override :

- HasMany.js
- HasManyFiles.js
- HasManyInPlace.js
- HasOne.js
- HadOneInPlace.js

Also TS detected that the return type was not `{QueryDefinition | QueryState}` but `{CozyClientDocument | QueryDefinition}`. I made the change but I didn't check if all callers expect that type (not sure how to easily find them all as `query` is kind of generic term)

packages/cozy-client/types/associations/Association.d.ts
Ldoppea: cf https://github.com/cozy/cozy-client/pull/964#discussion_r652811977

packages/cozy-client/src/models/timeseries.js
Ldoppea: Did the trick, change has been done.

packages/cozy-client/src/models/timeseries.js
Ldoppea: By applying @Crash-- suggested changes from https://github.com/cozy/cozy-client/pull/964#discussion_r651791541 then there is no need to name the object anymore

packages/cozy-client/src/associations/Association.js
Ldoppea: Ok, now there is [a problem with tests](https://travis-ci.com/github/cozy/cozy-client/builds/229340890).

Importing `CozyClient` in those files breaks a lot of tests (`watch` and `build` seem to work correctly though). I'm not sure why, maybe recursive import is not handled correctly?

That being said, i don't know how to type `client` as `CozyClient` without breaking everything. Any idea @ptbrowne @Crash-- ?

ptbrowne: Seems like a circular dependency bug, cozy-client imports association that imports hasmany that imports cozy-client. HasMany here seems to be imported first, which means cozy-client is not ready yet when the subclassing done. 

You can try using `import` inside the type annotation so that the import does not happen in JS context but only in typescript context.

```
  /**
   * @param {typeof import('../CozyClient').default} client - A client
   */
```

See https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#import-types

Ldoppea: Sadly this syntax is not supported by our jsdoc version (see https://github.com/microsoft/TypeScript/issues/14377). `yarn test` will work, but not `yarn docs`.
There is a fix for this but it doesn't seems to be merged in jsdoc yet : https://github.com/ricardohbin/jsdoc/pull/1 

I temporarily typed `CozyClient` as `object` so I can merge this PR and I will open an issue.

ptbrowne: Ah yes, this is unfortunate, good call for the temporary typing. I think jsdoc should be removed and tsdoc should be used. This way there would no incompatibility between different parsers. I have not used it though, and do not have any XP on it.

Crash-- (approved):

Issue #966, opened 2021-06-17
include_docs: true is necessary for fetchChanges


ptbrowne:
The code in fetchChanges expect the caller to call fetchChanges
with { include_docs: true } since in the implementation of
fetchChanges there is changes = changes.filter(change => !!change.doc).

It is not very ergonomic and could be set by default. At the moment,
it makes no sense to have { include_docs: false } otherwise you get
no documents in return.

PR #967, opened 2021-06-17
feat: Include_docs is done by default by fetchChanges

ptbrowne:
Fetch changes needs to have the docs included in the changes response
since it filters null documents, deleted documents, and design docs.
With include_docs: false, fetchChanges was not functional, thus it makes
no sense to support it. Now, the caller of fetchChanges doesn't need to
pass include_docs: true, it is set automatically by
DocumentCollection::fetchChanges.

Fix https://github.com/cozy/cozy-client/issues/966

See https://github.com/cozy/cozy-banks/pull/2124/commits/b358741e8949e9e158194e84ca77e94e02695902 for an example of bug that can be avoided with this change (include_docs: true had been forgotten, thus no doc changes were returned).

### DIFF
--- a/docs/api/cozy-stack-client.md
+++ b/docs/api/cozy-stack-client.md
@@ -104,6 +104,16 @@ contains the indexed attributes which are not in correct order.</p>
 ## Typedefs
 
 <dl>
+<dt><a href="#CouchOptionsRaw">CouchOptionsRaw</a> : <code>object</code></dt>
+<dd><p>Calls _changes route from CouchDB
+No further treatment is done contrary to fetchchanges</p>
+</dd>
+<dt><a href="#FetchChangesReturnValue">FetchChangesReturnValue</a> ⇒ <code><a href="#FetchChangesReturnValue">FetchChangesReturnValue</a></code></dt>
+<dd><p>Use Couch _changes API
+Deleted and design docs are filtered by default, thus documents are retrieved in the response
+(include_docs is set to true in the parameters of _changes).</p>
+<p>You should use fetchChangesRaw to have low level control on _changes parameters.</p>
+</dd>
 <dt><a href="#DirectoryAttributes">DirectoryAttributes</a> : <code>object</code></dt>
 <dd><p>Attributes used for directory creation</p>
 </dd>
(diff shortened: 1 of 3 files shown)


JF-Cozy (approved):

Crash-- (changes requested):
I don't know what to think about this fix yet. 

include_docs false is the default behavior of CouchDB. https://docs.couchdb.org/en/stable/api/database/changes.html 

Also, if we set it to true by default, we should be able to override it by setting it to false. It seems not possible in this PR ATM. 

@paultranvan do you have an opinion about this change?

ptbrowne:
I agree with you that include_docs should be settable to false. The thing is that right now, fetchChanges has a bigger scope that CouchDB fetchChanges since it deals with deleted documents, design docs, etc... If include_docs could be set to false, it would be incompatible with the other options. I could stir the PR in this sense, making it possible to set it as false and removing the filters that need to have docs.

paultranvan:
I agree that include_docs should be settable to false, as I does make sense to query the `_changes` to only retrieve changed doc ids, just like the default behaviour of this API. 

> I could stir the PR in this sense, making it possible to set it as false and removing the filters that need to have docs.

It seems good to me :) 

About the default value of `include_docs`, I generally tend to prefer to respect the CouchDB's defaults, but I think it's also ok to set it to true if it's our default usage of the API, and avoid bugs such as the one referenced by @ptbrowne

ptbrowne:
I tried an implementation in https://github.com/cozy/cozy-client/pull/967/commits/4f314a5eb4d2ef713aec453ceffad2ae5ee7b76a but I do not think it is the way to go. Here fetchChanges is already too high level and I do not think we should try to accomodate the low level include_docs in it. You can see in the implementation that if we do have include_docs set to false, we are forced to send back the complete `results`. Hence the shape of the return value is different if include_docs is set to false or true.

```
true: { docs, newSeq }
false: { results, newSeq }
```

I think to accomodate the need to have low level access to _changes params and keep the existing fetchChanges, instead of including the possibility to  set include_docs to false into fetchChanges, we could have another method `fetchChangesRaw` that would be 1-1 with _changes (thus you could call it with include_docs: false). And fetchChanges would stay at a higher level, with include_docs set to true (and not settable to false), and with the responsibility to filter design docs and deleted documents.

paultranvan:
@ptbrowne Ok, `fetchChangesRaw` looks like a good idea :+1: 
On a side note, I think it's kind of a pity to return docs as `documents`, while we already have `data`, `docs` or `rows` depending on the API (but I assume this is an old design choice)

packages/cozy-client/types/types.d.ts
```diff
@@ -19,7 +19,7 @@ export type ClientResponse = any;
 export type Manifest = any;
 export type OldCozyClient = any;
 export type NodeEnvironment = any;
-export type QueryFetchStatus = "failed" | "loading" | "pending" | "loaded";
+export type QueryFetchStatus = "loading" | "loaded" | "pending" | "failed";
```
Crash--, line 22: the come back!

Crash-- (commented):
I like the idea of the `fetchChangesRaw` method. 

So `fetchChanges` to be "high level" for an application, `fetchChangesRaw` if we want to be as close as possible to the Couch API. 👍

Crash-- (approved):

PR #971, opened 2021-06-17
fix: Updating a file also sanitize its name

Crash--:
Before the sanitize function was only called
during the creation of a file not during the 
update.


packages/cozy-stack-client/src/FileCollection.spec.js
```diff
@@ -399,6 +399,16 @@ describe('FileCollection', () => {
         client.fetchJSON.mock.calls[client.fetchJSON.mock.calls.length - 1]
       ).toMatchSnapshot()
     })
+
+    fit('should sanitize the filename', async () => {
```
JF-Cozy, line 403: `fit` ;)

JF-Cozy (approved):

PR #975, opened 2021-06-24
fix: Remove docs ids that have been removed from docs slice

ptbrowne:
We had a crash when sorted queries that are updated from a response
have ids for documents that have been deleted. In this case,
the properId helper would crash since it expects documents. A
filtering pass after retrieving each document is necessary to
remove deleted documents ids from the query state.

### DIFF
--- a/packages/cozy-client/src/store/queries.js
+++ b/packages/cozy-client/src/store/queries.js
@@ -64,7 +64,9 @@ const updateQueryDataFromResponse = (queryState, response, documents) => {
     const sorter = makeSorterFromDefinition(queryState.definition)
     const doctype = queryState.definition.doctype
     const allDocs = documents[doctype]
-    const docs = updatedIds.map(_id => allDocs[_id])
+    const docs = allDocs
+      ? updatedIds.map(_id => allDocs[_id]).filter(Boolean)
+      : []
     const sortedDocs = sorter(docs)
     updatedIds = sortedDocs.map(properId)
   }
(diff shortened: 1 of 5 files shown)


packages/cozy-client/src/store/queries.js
```diff
@@ -64,7 +64,9 @@ const updateQueryDataFromResponse = (queryState, response, documents) => {
     const sorter = makeSorterFromDefinition(queryState.definition)
     const doctype = queryState.definition.doctype
     const allDocs = documents[doctype]
-    const docs = updatedIds.map(_id => allDocs[_id])
+    const docs = updatedIds
+      .map(_id => (allDocs ? allDocs[_id] : null))
+      .filter(Boolean)
```
Ldoppea, line 69: As `allDocs` impacts all the loop, would it be more readable/optimised to put the condition around the `map` instead of inside? I think it would also allow to remove the boolean filter.

Ldoppea (approved):
Ok for me. I think the code may be written a bit differently (cf my comment) but I don't have any opposition to merge the current version.

packages/cozy-client/src/store/queries.js
ptbrowne: There are two things here, _id maybe not present in allDocs, in this case we'll have an undefined, and there is the allDocs check. I agree that the allDocs check can be put outside, but the boolean filter will stay.

ptbrowne: Done, thanks.

flohhhh (approved):

Issue #969, opened 2021-06-17
`CozyClient` cannot be used as JSdoc type in Association.js and derived classes


Ldoppea:
Affected files : `Association.js`, `HasMany`, `HasManyFiles`, `HasManyInPlace`, `HasOne`, `HasOneInPlace`

In those files `query(document, client, assoc)` should be typed with CozyClient as 2nd parameter.
But if `CozyClient` is imported into those files, then unit tests fail.

Using `@param {import('../CozyClient').default} client` syntax doesn't work either as this syntax is not handled by `jsdoc`.

Jsdoc seems to support this syntax in later versions but then node engine should be updated.

So client is currently typed as `object` until a better solution is found.

More info : https://github.com/cozy/cozy-client/pull/964#discussion_r653360823 and https://github.com/cozy/cozy-client/pull/964

PR #976, opened 2021-06-24
docs: Generate docs for cozy-client via typedoc

ptbrowne:
Result visible here: https://ptbrowne.github.io/cozy-docs-v3/en/cozy-client/api/cozy-client/modules/

jsdoc does not correctly parse some typescript supported syntaxes
(imports for example, or type property extraction). It caused problem
when trying to some types. See https://github.com/cozy/cozy-client/issues/969
By using typedoc, we stay in the typescript ecosystem and use the same parser.

Here, jsdoc is replaced by typedoc to generate the API documentation of
cozy-client in markdown.

remark is used to postprocess the generated markdown documentation to strip
deep headings and replace them with emphasis nodes so that they do not
appear on the right table of contents on https://docs.cozy.io.

An interesting side effect is that each class / or namespace is on one
page, it makes it easier to navigate the doc and to see the API.

Fix https://github.com/cozy/cozy-client/issues/969

Next steps would be (to fully use typescript across the repo, and to remove jsdoc) :

- Move tsconfig.json to the root of the workspace so that it is usable by every package
- Run typecheck on every package
- Have typedoc generate the documentation for all package

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '10'
+  - '12'
 cache:
   yarn: true
   directories:
--- a/docs/api/cozy-client/README.md
+++ b/docs/api/cozy-client/README.md
@@ -1,0 +1,19 @@
+cozy-client / [Exports](modules.md)
+
+# cozy-client
+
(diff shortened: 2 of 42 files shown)


Ldoppea (approved):
Great news :)

Crash-- (approved):

PR #978, opened 2021-06-25
fix: LoadInstanceOptionsFromDOM add root datas and cozy datas

JF-Cozy:
fix https://github.com/cozy/cozy-client/issues/977

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -5,6 +5,7 @@ import uniqBy from 'lodash/uniqBy'
 import zip from 'lodash/zip'
 import forEach from 'lodash/forEach'
 import get from 'lodash/get'
+import omit from 'lodash/omit'
 import MicroEE from 'microee'
 
 import CozyStackClient, { OAuthClient } from 'cozy-stack-client'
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -1492,7 +1493,10 @@ instantiation of the client.`
     }
 
     this.instanceOptions = root.dataset.cozy
-      ? JSON.parse(root.dataset.cozy)
+      ? {
+          ...JSON.parse(root.dataset.cozy),
+          ...omit({ ...root.dataset }, 'cozy')
+        }
```
Ldoppea, line 1499: Would something like this work?
```js
const {cozy, ...dataset} = root.dataset;

this.instanceOptions = {...cozy, ...dataset}
```
if so it would allow to delete the ternary operator and the usage of omit()

Crash--: I think this one should works better: 
```jsx
const {cozy = '{}', ...dataset} = root.dataset;
this.instanceOptions = {...JSON.parse(cozy), ...dataset}
```

Ldoppea (approved):

packages/cozy-client/src/CozyClient.js
```diff
@@ -5,6 +5,7 @@ import uniqBy from 'lodash/uniqBy'
 import zip from 'lodash/zip'
 import forEach from 'lodash/forEach'
 import get from 'lodash/get'
+import omit from 'lodash/omit'
```
Crash--, line 8: We should gave removed this import no? wondering why the CI was not red?

PR #988, opened 2021-07-19
fix: Cached queries optimization

Crash--:
In https://github.com/cozy/cozy-client/pull/925 we added the check on
the fetch status to return nothing when a query was already loading.

In https://github.com/cozy/cozy-client/pull/935 we decided to go deeper
and return the same promise when the query was already loading but
without any early return. The result was a perf degredation in our apps
since CozyClient was dispatching loadQuery / ResultQuery each time. So
if the request was done 10 times, the request was only executed once,
but the dispatch was done 10 times. Resulting in a lot of renders within
the app.

Here, we combine the best of the two PRs. We early return to bypass the
dispatch events, but we also return the stored promise to be able to get
the result even in a world without React.

The test was done within the Banks app.


packages/cozy-client/src/CozyClient.js
```diff
@@ -834,6 +834,14 @@ client.query(Q('io.cozy.bills'))`)
         return
       }
     }
+    // If we already have the same query in loading state, no
+    // need to refetch it. We return the promise we already
+    // have in the promiseCache
+    if (existingQuery && Object.keys(existingQuery).length > 0) {
```
Ldoppea, line 840: is `Object.keys(existingQuery).length > 0` necessary as we check if `existingQuery.fetchStatus` has a value on the next line?

packages/cozy-client/src/promise-cache.js
```diff
@@ -41,6 +41,17 @@ class PromiseCache {
       this.pending[key] = null
     }
   }
+  /**
+   *
+   * @param {function(): string} keyFunc - Returns a key to find in cache to find a pending promise.
```
Ldoppea, line 46: Why do we want a function here?

packages/cozy-client/src/promise-cache.js
```diff
@@ -41,6 +41,17 @@ class PromiseCache {
...
+   * @returns {Promise | null}
+   */
+  get(keyFunc) {
+    const key = keyFunc()
+    const already = this.pending[key]
+    if (already) return already
+    return null
```
Ldoppea, line 53: nit: in my (strictly personal) point of view `return already || null` would be clearer. But it's just a matter of code style.

Ldoppea (approved):
Code logic seems good, but I don't have enough XP on cozy-client to make a "complete" review with side effects in mind.

packages/cozy-client/src/CozyClient.js
Crash--: It's a good question. I think it can't be happen in real live to have `existingQuery` something else than `null` or `{}`. 

Replaced by  
```
if (existingQuery && existingQuery.fetchStatus === 'loading') { 
```

packages/cozy-client/src/promise-cache.js
Crash--: To be coherent with the exec signature.

PR #989, opened 2021-07-21
Add API for Pouch migration

paultranvan:
PouchDB can use different kind of adapter, e.g. idb, sqlite, etc.
We want to be able to migrate the adapter, typically to use the new and
more performant IndexedDB adapter.

To perform such migration, we run a local replication between the old
and new database instance. By doing so, the user can migrate its data
without requiring an internet connection.

Note the migration can take some time, i.e. several minutes depending on
the device, the database size, etc.
It is the responsability of the app to to call the `migrateAdapter` API
and handle this migration on the UI level.

Once the adapter is migrated, the old database is removed, and the page
reloaded so the client will use the new Pouch. The adapter name is saved
in the localstorage to inform the app which adapter should be used.

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -128,13 +128,8 @@ class PouchLink extends CozyLink {
       for (const doctype of doctypes) {
         const prefix = getPrefix(url)
         const dbName = getDatabaseName(prefix, doctype)
-        const oldPouch = new PouchDB(dbName, {
-          adapter: fromAdapter,
-          location: 'default'
-        })
-        await migratePouch(oldPouch, toAdapter)
-        await oldPouch.close()
-        await oldPouch.destroy()
+
+        await migratePouch({ dbName, fromAdapter, toAdapter })
       }
       persistAdapterName('indexeddb')
       // Reload page to benefit from new adapter
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -114,6 +114,11 @@ class PouchLink extends CozyLink {
     this.client = client
   }
 
+  /**
+   * Migrate the current adapter
+   *
+   * @param {{fromAdapter, toAdapter, url, plugins}} - The migration params
+   */
   async migrateAdapter({ fromAdapter, toAdapter, url, plugins }) {
     try {
       for (const plugin of plugins) {
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -117,7 +117,13 @@ class PouchLink extends CozyLink {
   /**
    * Migrate the current adapter
    *
-   * @param {{fromAdapter, toAdapter, url, plugins}} - The migration params
+   * @typedef {object} MigrationParams
+   * @property {string} [fromAdapter] - The current adapter type, e.g. 'idb'
+   * @property {string} [toAdapter] - The new adapter type, e.g. 'indexeddb'
+   * @property {string} [url] - The Cozy URL
+   * @property {Array<object>} [plugins] - The PouchDB plugins
+   *
+   * @param {MigrationParams} params - Migration params
    */
   async migrateAdapter({ fromAdapter, toAdapter, url, plugins }) {
     try {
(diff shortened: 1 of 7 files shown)


packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -97,13 +105,38 @@ class PouchLink extends CozyLink {
     this.client = client
   }
 
+  async migrateAdapter({ fromAdapter, toAdapter, url, plugins }) {
```
Crash--, line 108: can you add the JSDoc please?

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -97,13 +105,38 @@ class PouchLink extends CozyLink {
...
+          adapter: fromAdapter,
+          location: 'default'
+        })
+        await migratePouch(oldPouch, toAdapter)
+        await oldPouch.close()
+        await oldPouch.destroy()
+      }
+      persistAdapterName('indexeddb')
+      // Reload page to benefit from new adapter
+      window.location.reload()
```
Crash--, line 127: Since this is the responsability of the app to do the display during the migration, it should also be the responsability of the app to know what to do after the migration?

packages/cozy-pouch-link/src/migrations/adapter.js
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/**
+ * Migrate a PouchDB database to a new adapter.
+ *
+ * @param {object} oldPouch - The pouch to migrate
+ * @param {string} toAdapter - The target adapter name
+ * @returns {object} - The migrated pouch
+ */
+export const migratePouch = async (oldPouch, toAdapter) => {
```
Crash--, line 14: I'm not a big fan of this API. First arg is a Pouch Object, second one the name of the Adapter.

Can't we pass within the first arg the name of the adapter? And the name of the old DB? Like that, we only instanciate Pouch in this method not before?

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -97,13 +105,38 @@ class PouchLink extends CozyLink {
...
+      for (const doctype of doctypes) {
+        const prefix = getPrefix(url)
+        const dbName = getDatabaseName(prefix, doctype)
+        const oldPouch = new PouchDB(dbName, {
+          adapter: fromAdapter,
+          location: 'default'
+        })
+        await migratePouch(oldPouch, toAdapter)
+        await oldPouch.close()
+        await oldPouch.destroy()
```
Crash--, line 123: In order to be collected by the garbage, we need to "re init" the oldPouch var. It can be done by writing `oldPouch = null" or by using weakref (https://web.dev/detached-window-memory-leaks/)

Crash-- (approved):
few comments but looks good :)

packages/cozy-pouch-link/src/CozyPouchLink.js
paultranvan: Quite interesting, thanks!

packages/cozy-pouch-link/src/CozyPouchLink.js
paultranvan: I think we mentioned together to do it through cozy-client, but I don't quite remember the reasoning. Anyway, it makes sense to me to do it on the app side.

packages/cozy-pouch-link/src/migrations/adapter.js
paultranvan: Is e265413 better?

packages/cozy-pouch-link/src/CozyPouchLink.js
paultranvan: Fixed with 78d3d78

packages/cozy-pouch-link/src/CozyPouchLink.js
paultranvan: Fixed in 24cbbc9

packages/cozy-pouch-link/src/CozyPouchLink.js
paultranvan: Weird, I think I somehow missed a commit :thinking: 
Anyway fixed with ac29aac

packages/cozy-pouch-link/src/migrations/adapter.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param {MigrationParams} params - The migration params
+ * @returns {object} - The migrated pouch
+ */
+export const migratePouch = async ({ dbName, fromAdapter, toAdapter }) => {
+  let oldPouch = new PouchDB(dbName, {
+    adapter: fromAdapter,
+    location: 'default'
+  })
+  const newdbName = getNewIndexedDBDatabaseName(oldPouch.name)
```
Crash--, line 23: Why not using `dbName` instead of `oldPouch.name`?

Crash-- (approved):

PR #990, opened 2021-07-29
feat: Add sharing rules for bitwarden organizations

Ldoppea:
In order to implement folders and ciphers sharing inside of Cozy Pass we need to implement rules that are specific to Bitwarden's organizations

Note: in the Cozy's context, a folder is equivalent to a Bitwarden's organization containing a single collection with the same name. That's why the name "folder" doesn't appears in this new code

This PR is needed for https://github.com/cozy/cozy-pass-web/pull/32

### DIFF
--- a/packages/cozy-stack-client/src/SharingCollection.js
+++ b/packages/cozy-stack-client/src/SharingCollection.js
@@ -2,6 +2,8 @@ import DocumentCollection, { normalizeDoc } from './DocumentCollection'
 import { isFile, isDirectory } from './FileCollection'
 import { uri } from './utils'
 export const SHARING_DOCTYPE = 'io.cozy.sharings'
+export const BITWARDEN_ORGANIZATIONS_DOCTYPE = 'com.bitwarden.organizations'
+export const BITWARDEN_CIPHERS_DOCTYPE = 'com.bitwarden.ciphers'
 
 const normalizeSharing = sharing => normalizeDoc(sharing, SHARING_DOCTYPE)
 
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -231,6 +234,8 @@ export const getSharingRules = (document, sharingType) => {
       `sharingType is deprecated and will be removed. We now set this default rules: ${
         isFile(document)
           ? getSharingRulesForFile(document)
+          : document._type === BITWARDEN_ORGANIZATIONS_DOCTYPE
+          ? getSharingRulesForOrganizations(document)
           : getSharingRulesForPhotosAlbum(document)
```
Ldoppea, line 239: Here the linter doesn't let me to indent the nested ternary operator.

I find this version easier to read :
```js
isFile(document)
  ? getSharingRulesForFile(document)
  : document._type === BITWARDEN_ORGANIZATIONS_DOCTYPE
    ? getSharingRulesForOrganizations(document)
    : getSharingRulesForPhotosAlbum(document)
```

Please tell me what do you prefer :
- to keep current format
- to find a way to indent it correctly -> need to edit linter configuration, maybe to upgrade its version
- rewrite the code with classic if/else code

JF-Cozy: Le 2 idéalement, sans y passer des heures. Si le 1 t'as posé problème à la lecture, go pour le 3 ?

Crash--: Personally, I would go for the 3rd. I don't like nested ternary

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -296,6 +303,40 @@ const getSharingPolicyForFile = (document, sharingType) => {
     : { update: 'push', remove: 'revoke' }
 }
 
+const getSharingRulesForOrganizations = document => {
```
Crash--, line 306: Let's write a JSdoc. Let's prepare the futur ;)

Crash-- (approved):

packages/cozy-stack-client/src/SharingCollection.js
Ldoppea: I finally chose to rewrite it to the old fashion way. Current linter version seems not to handle ternary operator indentation and upgrading needs some edits on the project's configuration.

https://github.com/cozy/cozy-client/pull/990/files#diff-eedaf5c6b0118a4e2a44524e49bf386befb317815ee159d97a2242a578b96724R229-R240

packages/cozy-stack-client/src/SharingCollection.js
Ldoppea: Done, I documented only the added function. Others functions have been documented in a separate PR : https://github.com/cozy/cozy-client/pull/994

PR #995, opened 2021-08-03
Add CozyClient::saveAll

ptbrowne:
Cozy client now has the ability to save multiple documents
in one go.

It uses the _bulk_docs API for the stack link and the
bulkDocs method of PouchDB.


packages/cozy-client/src/CozyClient.js
```diff
@@ -605,6 +606,28 @@ client.query(Q('io.cozy.bills'))`)
...
+   * @param  {Object} mutationOptions
+   * @returns {Promise<void>}
+   */
+  async saveAll(docs, mutationOptions = {}) {
+    const doctypes = Array.from(new Set(docs.map(x => x._type)))
+    if (doctypes.length !== 1) {
+      throw new Error('saveAll can only save documents with the same doctype')
+    }
+    const ret = every(await Promise.all(docs.map(d => this.schema.validate(d))))
+    if (ret !== true) throw new Error('Validation failed for at least one doc')
```
Crash--, line 624: Shouldn't we log something? To help the debug in the app's land? To know the concerned document(s)?

packages/cozy-client/src/CozyClient.js
```diff
@@ -663,6 +686,20 @@ client.query(Q('io.cozy.bills'))`)
     }
   }
 
+  /**
+   * Dehydrates and adds metadata before saving a document
+   *
+   * @param  {CozyClientDocument} doc - Document that will be saved
+   * @returns {CozyClientDocument}
```
Crash--, line 693: We should certainly create a new type here one day.

packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -734,6 +734,55 @@ describe('CozyClient', () => {
...
+      await expect(
+        client.saveAll([
+          { _type: 'io.cozy.todos', label: 'Buy milk', done: false },
+          { _type: 'io.czoy.simpsons', name: 'Homer' }
+        ])
+      ).rejects.toEqual(
+        new Error('saveAll can only save documents with the same doctype')
+      )
+    })
+    it('should call work with the default stack client', async () => {
```
Crash--, line 749: s/call// ?

packages/cozy-client/src/StackLink.js
```diff
@@ -1,8 +1,26 @@
...
+ */
+const transformBulkDocsResponse = (updateAllResponse, originalDocuments) => {
+  return {
+    data: originalDocuments.map((od, i) => ({
+      ...od,
+      _id: updateAllResponse[i].id,
+      _rev: updateAllResponse[i].rev
+    }))
+  }
+}
```
Crash--, line 20: Are we sure that `CouchDB::_bulk_docs` returns the result in the same order as the as the payload we sent?

packages/cozy-client/src/StackLink.js
```diff
@@ -1,8 +1,26 @@
 import { MutationTypes } from './queries/dsl'
 import CozyLink from './CozyLink'
+import { CozyClientDocument } from './types'
 
 /**
+ * Returns full documents after a bulk update
  *
+ * @param  {{ ok: boolean, id: string, rev: string }[]} updateAllResponse - Response from bulk docs
```
Crash--, line 8: We can also receive: 
`error` and `reason` if the update of a particular doc has failed. 

https://docs.couchdb.org/en/stable/api/database/bulk-api.html?highlight=_bulk_docs#bulk-documents-transaction-semantics

Since it'll not throw an error we should store these informations in the response. 

But I don't know the right way to handle the fail? Is it on the app's land to map on the response to check if something has failed and do something? Do we provide feedback from cozyclient?

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -442,19 +446,42 @@ class PouchLink extends CozyLink {
...
   }
 
+  async updateDocuments(mutation) {
+    const docs = mutation.documents
+    const res = await this.dbMethod('bulkDocs', mutation)
+    const badResults = res.filter(x => !x.ok)
+    if (badResults.length > 0) {
+      console.warn('Error while bulk saving, bad results', badResults)
+      throw new Error('Error while bulk saving, see above for errors')
+    }
```
Crash--, line 456: I don't know if we should throw here. Since some documents can have been already updated in the DB we can have difference between the DB and the CozyClient.

packages/cozy-client/src/CozyClient.spec.js
ptbrowne: Done

packages/cozy-client/src/StackLink.js
ptbrowne: Good question, this is not specified in the doc, but it would be a problem if it weren't since otherwise you would not be able to link newly created documents (without id) to their server part. The code seems to indicate that only maps are used so I think it should be OK: https://github.com/apache/couchdb/blob/0059b8f90e58e10b199a4b768a06a762d12a30d3/src/couch/src/couch_httpd_db.erl, I'll recheck tomorrow.

packages/cozy-client/src/StackLink.js
ptbrowne: I have modified the code to be closer to what happens in cozy-pouch-link.


```
    const badResults = updateAllResponse.filter(x => !x.ok)
    if (badResults.length > 0) {
      console.warn('Error while bulk saving, bad results', badResults)
      throw new Error('Error while bulk saving, see above for errors')
    }
```

I think the app should refetch the query in this case.

> Do we provide feedback from cozyclient?

 I attached the bad documents inside the error if the developer needs it.

packages/cozy-pouch-link/src/CozyPouchLink.js
ptbrowne: What do you suggest ? I think in this case, the app should refetch all affected documents.

packages/cozy-client/src/CozyClient.js
ptbrowne: Done in 8c5a895c

packages/cozy-client/src/CozyClient.js
ptbrowne: OK.

packages/cozy-client/src/CozyClient.spec.js
ptbrowne: Solved

packages/cozy-pouch-link/src/CozyPouchLink.js
Crash--: How can the app can know what documents to refetch? We don't return them since we throw, no? 

Let's say the doc1 updates is OK, doc2 is KO.

Why the current implementation I don't see how the app can handle the case. 

```js 
try {
const data1 = {id: 1}
const doc1 = {...data1, foo:'bar'}
const doc2 = {...data2, foo:'bar'}
const updatedDocs = await Cozyclient.saveAll([doc1, doc2])
} catch(e){
const testdoc1 = await CozyClient.query(id1) // => contain foo:bar 
const testdocs2 = await CozyClient.query(id2) // => contain no foo:bar  
}
```
What do you think of returning something like this:

```js
const updatedDocs = await Cozyclient.saveAll([doc1, doc2, doc3])
if(updatedDocs.errored){
   const failedUpdates = e.errored
   failedUpdates.map(doc => {
     console.log(doc.reason)
     console.log(doc.error)
  }
}
if(updatedDocs.data){
  const successfulUpdate = updatedDocs.data....
}
```

So instead of throwing an error and not returning the result, we return the result and the errors. The return is a bit different than our others methods since `errored` attribute doesn't exist in our `normalization` but I think it can be a good move. WDYT?

ptbrowne: I went a different route in https://github.com/cozy/cozy-client/pull/995/commits/dfc32c003ed739b0c81b21e93839280f1b01a7b6 , I attached the bad docs to the error. WDYT ?

ptbrowne: I think that if I went the way of the object with the `errored` object returned as part of th result, I'd need to make more changes to the `mutate` function from cozy-client, which only rely on exceptions to dispatch a receiveMutationError.

For now, inside cozy-client we are not accustomed to checking an `errored` field, it would be a new behavior and I think it is better to stay consistent.

packages/cozy-client/src/CozyClient.js
```diff
@@ -605,6 +606,28 @@ client.query(Q('io.cozy.bills'))`)
     return this.mutate(this.getDocumentSavePlan(doc), mutationOptions)
   }
 
+  /**
+   * Saves multiple documents in one batch
+   * - Can only be called with documents from the same doctype
+   * - Does not support automatic creation of references
```
JF-Cozy, line 611: `Does not support automatic creation of references` Est-ce que ça devrait être noté dans une issue, ou est-ce un comportement normal et on ne fait ici que le préciser ?

packages/cozy-client/src/CozyClient.js
```diff
@@ -605,6 +606,28 @@ client.query(Q('io.cozy.bills'))`)
...
+   *
+   * @param  {CozyClientDocument[]} docs
+   * @param  {Object} mutationOptions
+   * @returns {Promise<void>}
+   */
+  async saveAll(docs, mutationOptions = {}) {
+    const doctypes = Array.from(new Set(docs.map(x => x._type)))
```
JF-Cozy, line 618: Pourquoi un `Set` ici, c'est pour avoir un tableau qu'avec des valeurs uniques ?

packages/cozy-client/src/StackLink.js
ptbrowne: I added the error/reason to the TS definition, I think we can finish this discussion in https://github.com/cozy/cozy-client/pull/995#discussion_r681871162.

packages/cozy-client/src/StackLink.js
ptbrowne: Yep the list is in order https://github.com/apache/couchdb/blob/bf355b96ee3efc24cb6edfb3eeb5b6fe519fdbf8/src/chttpd/src/chttpd_db.erl#L514.

packages/cozy-client/src/CozyClient.js
```diff
@@ -663,6 +686,20 @@ client.query(Q('io.cozy.bills'))`)
...
+   * @returns {CozyClientDocument}
+   */
+  prepareDocumentForSave(doc) {
+    const isNewDoc = !doc._rev
+    const dehydratedDoc = this.ensureCozyMetadata(dehydrate(doc), {
+      event: isNewDoc ? DOC_CREATION : DOC_UPDATE
+    })
+    return dehydratedDoc
+  }
+
```
JF-Cozy, line 710: Le `Dehydrates` permet de supprimer les `Association` ajouté au document et de récupérer la forme originale c'est ça ?

packages/cozy-client/src/CozyClient.js
ptbrowne: C'est un comportement normal à mon avis, les références sont spécifiques aux fichiers et ne sont pas utilisées pour les autres doctypes. A mon avis, il faudra créer l'issue au besoin.

packages/cozy-client/src/CozyClient.js
ptbrowne: Oui.

packages/cozy-client/src/CozyClient.js
ptbrowne: Oui.

packages/cozy-client/src/CozyClient.js
Crash--: L'issue sera plus de supprimer la notion de references et de passer aux relations pour les fichiers.

JF-Cozy: Ok merci pour l’éclaircissement 👍

packages/cozy-pouch-link/src/CozyPouchLink.js
ptbrowne: I've extracted BulkEditError inside cozy-client (remove duplication of the custom error class) and added methods on the error to get the updated documents as well as the errors.

See [`8e5aca5` (#995)](https://github.com/cozy/cozy-client/pull/995/commits/8e5aca5bfde7c0c3675a95ceb638ba37e820aa32)

Crash-- (approved):
It can be good to add an exemple in our `getting-started` on how to make a batch updates with this new method! 

The implem looks good, let's see

ptbrowne:
I added an example in the getting started document, thanks.

PR #997, opened 2021-08-06
fix: Move @types back to dependencies

Ldoppea:
In a previous commit abd847ce4bb29f405fc80109fd78f65fd780d5f2 (https://github.com/cozy/cozy-client/pull/987), I moved `@types/xxx` from `dependencies` to `devDependencies` but I was wrong.

Some of those types are exposed to the package's API and may be manipulated by consuming apps.

For example : 
- `index.js` contains `export { default as compose } from 'lodash/flow'`
- `CozyClient.js` declares `getIncludesRelationships` which contains `return fromPairs(....`. This function is from `lodash` and return a lodash `Dictionary`

Due to this the generated `.d.ts` file will contain references to lodash :
```ts
getIncludesRelationships(queryDefinition: any): import("lodash").Dictionary<any>;
```

Then consuming TS apps (here `cozy-pass-web`) will require to install `@types/lodash` in order to compile.

To prevent this we should put those packages back into the `dependencies` section.

### DIFF
--- a/packages/cozy-client/package.json
+++ b/packages/cozy-client/package.json
@@ -15,6 +15,8 @@
   },
   "dependencies": {
     "@cozy/minilog": "1.0.0",
+    "@types/jest": "^26.0.20",
+    "@types/lodash": "^4.14.170",
     "btoa": "^1.2.1",
     "cozy-device-helper": "^1.12.0",
     "cozy-flags": "2.7.1",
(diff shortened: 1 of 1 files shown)


Crash--:
So how did you resolve the issue on cozy-pass-web? Since this is a revert of https://github.com/cozy/cozy-client/pull/987 ?

Ldoppea:
> So how did you resolve the issue on cozy-pass-web? Since this is a revert of #987 ?

There are 2 options : 
- Revert the old PR -> so every project that references `cozy-client`  will automatically install those `@types/...`
- Manually install those `@types/...` in every project that references `cozy-client`

In `cozy-pass-web` I had to run `npm install --saveDev @types/lodash` in order to compile.
If this PR is merged, I'll be able to revert the edited `package.json`.
image

Crash-- (approved):

PR #1000, opened 2021-08-11
Pouch panel

ptbrowne:
Add devtool panel that displays:

- the last synchronization date of pouches (for all databases)
- the doctypes managed by pouch
- Buttons to start/stop replication
- Button to destroy local pouches

image

### DIFF
--- a/packages/cozy-client/src/devtools/Pouch.jsx
+++ b/packages/cozy-client/src/devtools/Pouch.jsx
@@ -1,0 +1,138 @@
+import React, { useMemo, useState, useCallback, useEffect } from 'react'
+import CozyPouchLink from 'cozy-pouch-link'
+import get from 'lodash/get'
+import Button from 'cozy-ui/transpiled/react/MuiCozyTheme/Buttons'
+import Typography from 'cozy-ui/transpiled/react/Typography'
+import format from 'date-fns/format'
+
+import useClient from '../hooks/useClient'
+import PanelContent from './PanelContent'
+
+const listStyle = { paddingLeft: '1rem' }
+
+const DATE_FORMAT = 'HH:mm:ss DD/MM/YYYY'
+
+const useForceUpdate = () => {
+  const [, setLastRender] = useState(null)
+  const forceUpdate = useCallback(() => {
+    setLastRender(Date.now())
+  }, [])
+  return forceUpdate
+}
+
+const DoctypeSyncInfo = ({ link, doctype }) => {
+  const info = link.getSyncInfo(doctype)
+  return (
+    <div>
+      Status: {info ? get(link, ['replicationStatus', doctype]) : 'Not synced'}
+      <br />
+      Date: {info ? format(info.date, DATE_FORMAT) : 'NA'}
+    </div>
+  )
+}
+
+/**
+ * Allows to view state and manage the PouchLink of the current cozy
+ * client.
+ */
+const PouchDevTool = () => {
+  const client = useClient()
+  const forceUpdate = useForceUpdate()
+  const [status, setStatus] = useState('NA')
+  const pouchLink = useMemo(() => {
+    return client.links.find(link => link instanceof CozyPouchLink)
+  }, [client])
+  const [resetting, setResetting] = useState(false)
+  useEffect(() => {
+    const handleReplicationStart = () => setStatus('replicating')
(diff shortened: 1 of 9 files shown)


Ldoppea (approved):
I just learned that we have custom devtools thanks to that PR 👍  (I think read the `devtools.md` during my first days at Cozy but I forgot that part since).

packages/cozy-client/src/devtools/Pouch.jsx
```diff
@@ -0,0 +1,116 @@
...
+        {pouchLink.doctypes.map(doctype => (
+          <li key={doctype}>{doctype}</li>
+        ))}
+      </ul>
+      <Typography variant="subtitle1">Last replication</Typography>
+      <p>
+        {pouchLink.lastReplicatedAt
+          ? format(pouchLink.lastReplicatedAt, 'HH:mm:ss DD/MM/YYYY')
+          : 'NA'}
+      </p>
```
Crash--, line 82: [troll on: instead of storing this in a new var, why not using the last sync attribute? https://docs.cozy.io/en/cozy-stack/settings/#post-settingssynchronized It'll force us to post the information troll off]. It'll work only for OAuth client... :( 


To be more productive, I don't know if this is the real lastReplicatedAt. If the sync was already done, we refresh the app, the lastReplicatedAt will be NA even though the replication was already done. Can't we get this information from Pouch itself?

Crash-- (approved):

packages/cozy-client/src/devtools/Pouch.jsx
```diff
@@ -0,0 +1,116 @@
...
+        <p>Pouch is not active</p>
+      </PanelContent>
+    )
+  }
+  return (
+    <PanelContent>
+      <Typography variant="subtitle1">Doctypes saved in Pouch</Typography>
+      <ul style={listStyle}>
+        {pouchLink.doctypes.map(doctype => (
+          <li key={doctype}>{doctype}</li>
```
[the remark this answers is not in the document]
ptbrowne: I think I am going to put "managed" in Pouch, it will be clearer that this does not mean that a doctype as been replicated yet.

> I'm wondering if we shouldn't add this "persisted information" here.

What kind ?

packages/cozy-client/src/devtools/Pouch.jsx
ptbrowne: This panel is meant to only be a view on the pouch link. I think the synchronisation status to the cozy is bit different and I am not quite aware of the semantic differences :/

As for the global replication status on the pouch link, in fact I doubt that this is useful since the link executes the query based on the per-doctype synchronisation status.

I've modified the PR to have synchronisation statuses per doctype. It allowed me also to piggyback on the syncedDoctypes attributes of the pouch manager that is automatically saved in localStorage so that it works across reloads.

image

packages/cozy-client/src/devtools/Pouch.jsx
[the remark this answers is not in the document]
ptbrowne: I think this is fixed by the latest changes, information is now displayed per doctype, and this information is persisted to localStorage.

packages/cozy-client/src/devtools/Pouch.jsx
```diff
@@ -0,0 +1,138 @@
...
+
+const listStyle = { paddingLeft: '1rem' }
+
+const DATE_FORMAT = 'HH:mm:ss DD/MM/YYYY'
+
+const useForceUpdate = () => {
+  const [, setLastRender] = useState(null)
+  const forceUpdate = useCallback(() => {
+    setLastRender(Date.now())
+  }, [setLastRender])
```
paultranvan, line 19: Out of curiosity, what is the purpose of using `useCallback` here? Won't `setLastRender` remain stable between renders?

paultranvan (approved):
LGTM, thanks for this and everything :)

packages/cozy-client/src/devtools/Pouch.jsx
ptbrowne: I've removed the callback ;)

paultranvan: :relieved:

Crash--:
thanks you for the update @ptbrowne :)

PR #1002, opened 2021-08-12
feat: Replace fetchAppLatestVersion by fetchAppVersion

doubleface:
The stack's registry API also allow to get a specific version of an app
and this is needed by the amiral app.

This is a breaking change since the fetchAppLatestVersion won't exist
anymore, but it is not used yet. I think it is safe to replace it now.- feat: Replace fetchAppLatestVersion by fetchAppVersion

### DIFF
--- a/docs/api/cozy-client/classes/registry.md
+++ b/docs/api/cozy-client/classes/registry.md
@@ -16,7 +16,7 @@
 
 *Defined in*
 
-[packages/cozy-client/src/registry.js:39](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/registry.js#L39)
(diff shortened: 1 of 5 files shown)


Crash-- (changes requested):
You should add a BREAKING CHANGE: message in the commit in order to create a new major version and to explain how to migrate. 

We're not alone using CozyClient :)

doubleface:
@Crash-- Done

Crash-- (approved):

JF-Cozy (approved):

Crash--:
nit: one way to not create a BC can be : 

fetchAppLatestVersion(app){
return fetchAppVersion(app, 'latest')
}

PR #1004, opened 2021-08-13
feat: Convert cozy-stack-client source to TypeScript

acezard:


packages/cozy-stack-client/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
Crash--, line 3: can we use our default https://github.com/cozy/cozy-libs/tree/master/packages/babel-preset-cozy-app ? We define  the browser we want to support here https://github.com/cozy/cozy-libs/tree/master/packages/browserslist-config-cozy 

This should be based on that no?

acezard: Well here I'm using directly TSC to compile instead of babel-loader (which is often done on libs https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html). But I could use Babel too, it's just harder to fine tune the compilation options. It's definitely an important decision in any case

acezard: By the way the compilation method can be changed at any point, starting with TSC doesn't block anything . It's just easier to get started with the development than with babel, but nothing will disallow us to switch to Babel when the project is error-free. The main concern I have is to make the source works in a very permissive TS config without breaking any runtime functionality. Luckily the tests will be helpful here

Crash--: My point is that if we compile TS to JS that is not supported by our target browsers it'll crash since we do not transpile the node_modules in the app's land. 

So we really need to be sure that the generated JS code works for the old browsers (think about an Android 5 Webview that is not updated ;))

acezard: yea es5 is generally very compatible https://caniuse.com/es5, problems will only arise on incredibly outdated runtimes like ie8.
TSC doesn't have browserlist support, but es5 should be pretty close to what is needed. That doesn't solve polyfills issues though if there are any, I don't know about that yet

Crash--: I'll repeat myself but we should have the same compilation / transpilation option for our all JS / TS codebase.

acezard: I understand, then the way to go is to use babel-loader (compilation) in conjunction with TSC (typechecking). We can tackle this issue at the end of the "typification" of this lib.

PR #1012, opened 2021-08-25
docs: Improve relationships documentation

paultranvan:
The relationships concept are a bit hard to grasp at first. Here, we
try to make it more understandable, and fix the section about files
reference creation.

### DIFF
--- a/docs/relationships.md
+++ b/docs/relationships.md
@@ -4,8 +4,15 @@ CouchDB is a document store. It does not materialize relations between documents
 
 ### In CouchDB
 
-To materialize relations, we use a special key inside the document named `relationships`. Please see the cozy-doctypes [documentation](https://github.com/cozy/cozy-doctypes/#relationships) for the exact syntax.
+Relations between documents are materialized under a `relationships` object at the root of the document. Relations are referenced by their names, e.g. `authors`.
 
+Each relation is an object with a `data` property, containing either **one reference**, or an **array of references**.
+
+A reference is an object containing at least:
+* `_type`:  the name of the referenced doctype, e.g. `io.cozy.contacts`
+* `_id`:  the id of the referenced document, e.g. ``.
+
+For instance, a book -> authors relationship might be represented like this:
 ```javascript
 {
   "_id": "mobydick",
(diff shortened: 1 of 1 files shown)


docs/relationships.md
```diff
@@ -42,59 +52,50 @@ const client = new CozyClient({
 })
 ```
 
-With the help of the schema, Cozy-Client will be able to manage the `relationships` attribute of your documents for you.
+We explain below what are the different types of relations.
```
acezard, line 55: Sounds a bit redundant, isn't the `### Relation Types` title enough here?

acezard (approved):

docs/relationships.md
```diff
@@ -4,8 +4,15 @@ CouchDB is a document store. It does not materialize relations between documents
...
 ### In CouchDB
 
-To materialize relations, we use a special key inside the document named `relationships`. Please see the cozy-doctypes [documentation](https://github.com/cozy/cozy-doctypes/#relationships) for the exact syntax.
+Relations between documents are materialized under a `relationships` object at the root of the document. Relations are referenced by their names, e.g. `authors`.
 
+Each relation is an object with a `data` property, containing either **one reference**, or an **array of references**.
+
+A reference is an object containing at least:
+* `_type`:  the name of the referenced doctype, e.g. `io.cozy.contacts`
+* `_id`:  the id of the referenced document, e.g. ``.
```
Ldoppea, line 13: `e.g.` is missing

docs/relationships.md
```diff
@@ -42,59 +52,50 @@ const client = new CozyClient({
...
-// there should be an io.cozy.contact document 
-// for which the _id is "hermanmelville"
-```
 
-`'has-one-in-place'` or `'has-many-in-place'` are here for backward compatibility. Instead of using the `relationships` attribute, you reference directly the ids of linked documents at the root of your data:
+#### Customize relations
+
+You are free to create your own relation type if you need a special behaviour.
+
+For instance, [Banks doctypes](https://github.com/cozy/cozy-banks/blob/master/src/doctypes.js) implements `HasManyBills` or `HasManyReimbursements` that inherits from `HasManyInPlace`.
```
Ldoppea, line 78: Here we refer to `HasManyInPlace` for the first time without defining what it is, maybe we should swap this section with the `Old relation type` so it will be defined first?

docs/relationships.md
```diff
@@ -191,13 +190,13 @@ const firstDoc = docs[0]
 firstDoc.printingCompany.unset()
 ```
 
-### Create a new document with existing relations
+### Create a new file with existing relations
 
-Simply pass your relations as a third parameter to `create()`:
+Simply pass your reference as a third parameter to `create()`:
```
Ldoppea, line 195: This phrase would be clearer by adding a subject before `reference` are we talking about a file reference? a relationship reference? etc

docs/relationships.md
paultranvan: Personally, when I encounter a new concept, such as the relation type here, I like to be informed right ahead that this concept will be explained later, without having to skip to the next section :)

PR #1013, opened 2021-08-25
feat: Allow to query a HasManyFiles relation in both ways 

paultranvan:
When using a HasManyFiles relationship, we can manipulate documents from
the referenced doc, or from the referencing doc.

Let's assume the following photo albums:

```
{
  _id: '1234'
  _type: 'io.cozy.photos.albums'
}
```

And the following file, with a reference to the album:

```
{
  _id: '5678'
  _type: 'io.cozy.files',
  referenced_by: [
    {
      id: '1234'
      type: 'io.cozy.photos.albums'
    }
  ]
}
```

Previously, it was only possible to query the files from the album.
But, we might want to query the album directly from the file. This is
what we add here, by querying documents specified in the `referenced_by`
relationship specified in the file, through their id.

### DIFF
--- a/packages/cozy-client/src/associations/HasManyFiles.js
+++ b/packages/cozy-client/src/associations/HasManyFiles.js
@@ -3,6 +3,7 @@ import omit from 'lodash/omit'
 import { Mutations, Q, QueryDefinition } from '../queries/dsl'
 import { getDocumentFromState } from '../store'
 import { CozyClientDocument } from '../types'
+import { DOCTYPE_FILES } from '../const'
 import Association from './Association'
 import HasMany from './HasMany'
 
@@ -13,7 +14,7 @@ import HasMany from './HasMany'
  */
 export default class HasManyFiles extends HasMany {
   async fetchMore() {
-    const queryDef = new QueryDefinition({ doctype: 'io.cozy.files' })
+    const queryDef = new QueryDefinition({ doctype: DOCTYPE_FILES })
     const relationships = this.getRelationship().data
     // Get last datetime for cursor
     const lastRelationship = relationships[relationships.length - 1]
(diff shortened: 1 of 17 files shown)


packages/cozy-client/src/associations/HasManyFiles.spec.js
```diff
@@ -122,6 +123,7 @@ describe('HasManyFiles', () => {
     const queryDef = HasManyFiles.query(file, null, {
       doctype: 'io.cozy.todos'
     })
+
     expect(queryDef.doctype).toEqual('io.cozy.todos')
     expect(queryDef.ids).toEqual(['1234'])
   })
```
JF-Cozy, line 130: nit: commit qui peut être squashé je pense

packages/cozy-client/types/const.d.ts
```diff
@@ -1 +1,2 @@
 export const REGISTRATION_ABORT: "REGISTRATION_ABORT";
+export const DOCTYPE_FILES: "io.cozy.files";
```
JF-Cozy, line 2: commit description à nettoyer peut-être ?

JF-Cozy (approved):

packages/cozy-client/src/associations/HasManyFiles.js
```diff
@@ -106,10 +107,9 @@ export default class HasManyFiles extends HasMany {
   static query(document, client, assoc) {
     if (document._type === DOCTYPE_FILES) {
       const refs = get(document, `relationships.referenced_by.data`, [])
-      const ids = refs
-        .filter(ref => ref.type === assoc.doctype)
-        .map(ref => ref.id)
-
+      const ids = uniq(
```
JF-Cozy, line 110: ça n'a cassé aucun test, est-ce que ça vaudrait pas le coup d'en rajouter un ?

JF-Cozy (approved):

packages/cozy-client/src/associations/HasManyFiles.js
paultranvan: Yes, j'ai ajouté une condition au test dans 5e75c47

PR #1020, opened 2021-09-02
fix: Correcty iterate over object keys

paultranvan:
This was causing a migration failure.

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -139,7 +139,7 @@ class PouchLink extends CozyLink {
         PouchDB.plugin(plugin)
       }
       const doctypes = getPersistedSyncedDoctypes()
-      for (const doctype of doctypes) {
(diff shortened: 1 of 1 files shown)


packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -139,7 +139,7 @@ class PouchLink extends CozyLink {
         PouchDB.plugin(plugin)
       }
       const doctypes = getPersistedSyncedDoctypes()
-      for (const doctype of doctypes) {
+      for (const doctype of Object.keys(doctypes)) {
```
Ldoppea, line 142: Antoher (purely cosmetic) way to write this may be `for (const doctype in doctypes)`

`of` iterates over elements while `in` iterates over indexes

Ldoppea:
It would be great to amend the commit message in order to explain that here doctypes is not an array but a dictionary. We can "guess" that by reading the code but it takes more time.

Ldoppea (approved):

Crash-- (approved):

PR #1023, opened 2021-09-08
feat: Pass name as param in updateFile

paultranvan:
We used to expect the updated file to be a File object with its name set
directly inside the File.
However, we could also manipulate files as ArrayBuffer. Hence, it is
necessary to be able to pass the name as a param.

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -24,6 +24,7 @@ import { dontThrowNotFoundError } from './Collection'
  * @property {string} dirId - Id of the parent directory.
  * @property {string} name - Name of the created file.
  * @property {Date} lastModifiedDate - Can be used to set the last modified date of a file.
+ * @property {boolean} executable - Whether or not the file is executable
  * @property {object} metadata io.cozy.files.metadata to attach to the file
  */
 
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -431,29 +432,25 @@ class FileCollection extends DocumentCollection {
...
+   * @param {File|Blob|Stream|string|ArrayBuffer} data file to be uploaded
+   * @param {FileAttributes} params       Additional parameters
    * @param  {object}  params.options     Options to pass to doUpload method (additional headers)
-   * @returns {object}                     Updated document
+   * @returns {object}                    Updated document
    */
   async updateFile(
     data,
-    { executable = false, fileId, metadata, ...options } = {}
+    { executable = false, fileId, name = '', metadata, ...options } = {}
```
JF-Cozy, line 442: est-ce qu'on ne rajouterait pas le `name` dans les `FileAttributes` ?

paultranvan: Déjà fait ;) 
https://github.com/cozy/cozy-client/blob/47ec6ceac4f918df97ddb005739acdfe39846583/packages/cozy-stack-client/src/FileCollection.js#L25

Crash-- (approved):

PR #1025, opened 2021-09-08
feat: Handle Onboarding V2

Crash--:
Now, in a specific workflow we can receive an registerToken and
we need to pass it to the authorize endpoint

So let's use the oauthOptions to write it

documentation is currently tested with a partner. I'll write it down to cozy-client after the first iteration

### DIFF
--- a/packages/cozy-stack-client/src/OAuthClient.js
+++ b/packages/cozy-stack-client/src/OAuthClient.js
@@ -287,14 +287,19 @@ class OAuthClient extends CozyStackClient {
   getAuthCodeURL(stateCode, scopes = this.scope) {
     if (!this.isRegistered()) throw new NotRegisteredException()
 
-    const query = {
+    let query = {
       client_id: this.oauthOptions.clientID,
       redirect_uri: this.oauthOptions.redirectURI,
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -287,14 +287,19 @@ class OAuthClient extends CozyStackClient {
...
-    const query = {
+    let query = {
       client_id: this.oauthOptions.clientID,
       redirect_uri: this.oauthOptions.redirectURI,
       state: stateCode,
       response_type: 'code',
       scope: scopes.join(' ')
     }
-
+    if (this.oauthOptions.registerToken) {
```
JF-Cozy, line 297: nit: pas de garde-fou style `this.oauthOptions?.registerToken` ? `this.oauthOptions` est forcément defined ?

JF-Cozy (approved):

packages/cozy-stack-client/src/OAuthClient.js
Crash--: oui

PR #1027, opened 2021-09-09
feat: Infer content-type from file extension

paultranvan:
We used to infer the file content-type only for File objects. However,
it can be useful to extend this for binary structs such as ArrayBuffer,
typically when dealing with encryption. Say we have decrypted a file,
represented as an ArrayBuffer: we have no way to know its content-type
without inferring it from the name, as it was previously encrypted with
a specific content-type.

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -870,23 +870,22 @@ class FileCollection extends DocumentCollection {
 
     let { contentType, contentLength, checksum, lastModifiedDate, ifMatch } =
       options || {}
+
     if (!contentType) {
-      if (isBuffer) {
-        contentType = CONTENT_TYPE_OCTET_STREAM
-      } else if (isFile) {
-        contentType =
-          data.type ||
-          getFileTypeFromName(data.name.toLowerCase()) ||
-          CONTENT_TYPE_OCTET_STREAM
-        if (!lastModifiedDate) {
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -870,23 +870,22 @@ class FileCollection extends DocumentCollection {
...
+        } else {
+          // Extract the name from the path and infer the type
+          const sPath = path.split('?')
+          const params = sPath.length > 1 ? sPath[1] : ''
+          const name = new URLSearchParams(params).get('Name')
+          contentType =
+            getFileTypeFromName(name.toLowerCase()) || CONTENT_TYPE_OCTET_STREAM
+        }
       }
     }
```
Crash--, line 890: it miss the `lastModifiedDated` code in your new version, no? https://github.com/cozy/cozy-client/pull/1027/files#diff-76c9574e3b7f8c2ec150af6ff717b3d38ff269adb500ea34697860b335e5ef79L881-L883

Crash--: And since there is no test for it, we should add one ^^'

paultranvan: My bad, I missed that. Fair enough for the test

paultranvan: There it is: 5e3158e

Crash-- (approved):

PR #1029, opened 2021-09-14
feat: Use add and remove API for both has-one and has-many

paultranvan:
We used to have `addById/removeById` for has-many relationships and
`set/unset` for has-one relationships, with different behaviour:
has-many saves the relationship in database, but not has-one.
To be consistent, we add `save/remove` API for both relationships.

### DIFF
--- a/docs/relationships.md
+++ b/docs/relationships.md
@@ -134,23 +134,23 @@ const firstAuthors = firstDoc.authors.data
 
 ### Add a relation to an existing document
 
-When the relationship is a `has-many` type, you can call the `addById(documentId)` to create the relationship:
+When the relationship is a `has-many` type, you can call the `add(docs)` to create the relationship:
 
 ```javascript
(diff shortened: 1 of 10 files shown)


docs/relationships.md
```diff
@@ -134,23 +134,23 @@ const firstAuthors = firstDoc.authors.data
...
-`addById` also accepts an array of document ids:
+`add` also accepts single document:
 
 ```javascript
-firstDoc.authors.addById([mainAuthorId, coAuthorId])
+firstDoc.authors.add({_id: 'Adleman'})
 ```
 
-When the relationship is a `has-one` type, use `set(document)` instead:
+Likewise, When the relationship is a `has-one`, use `add(doc)`:
```
Crash--, line 153:
```suggestion
Likewise, when the relationship is a `has-one`, use `add(doc)`:
```

Crash-- (approved):

PR #1041, opened 2021-10-01
feat: Add OAuthClientsCollection

taratatach:
The `io.cozy.oauth.clients` doctype is protected so we need to
manually build a collection to query it.
We only add the `all`, `get` and `destroy` methods as they can be used
right away in the Settings app.

### DIFF
--- a/packages/cozy-stack-client/src/OAuthClientsCollection.js
+++ b/packages/cozy-stack-client/src/OAuthClientsCollection.js
@@ -1,0 +1,125 @@
+import get from 'lodash/get'
+
+import DocumentCollection from './DocumentCollection'
+import { uri } from './utils'
+import * as querystring from './querystring'
+import { dontThrowNotFoundError } from './Collection'
+import { FetchError } from './errors'
+
+export const OAUTH_CLIENTS_DOCTYPE = 'io.cozy.oauth.clients'
+
+const normalizeDoc = DocumentCollection.normalizeDoctypeJsonApi(
+  OAUTH_CLIENTS_DOCTYPE
+)
+const normalizeOAuthClient = client => ({
+  ...normalizeDoc(client, OAUTH_CLIENTS_DOCTYPE),
+  ...client.attributes
+})
+
+/**
+ * Implements `DocumentCollection` API to interact with the /settings/clients endpoint of the stack
+ */
+class OAuthClientsCollection extends DocumentCollection {
+  constructor(stackClient) {
+    super(OAUTH_CLIENTS_DOCTYPE, stackClient)
+  }
+
+  /**
+   * Fetches all OAuth clients
+   *
+   * @param  {object} options           Query options
+   * @param  {number} options.limit     For pagination, the number of results to return.
+   * @param  {object} options.bookmark  For cursor-based pagination, the index cursor.
+   * @param  {array} options.keys       Ids of specific clients to return (within the current page),
+   * @returns {object}                  The JSON API conformant response.
+   */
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/OAuthClientsCollection.js
```diff
@@ -0,0 +1,122 @@
...
+  constructor(stackClient) {
+    super(OAUTH_CLIENTS_DOCTYPE, stackClient)
+  }
+
+  /**
+   * Fetches all OAuth clients
+   *
+   * @param  {object} options           Query options
+   * @param  {number} options.limit     For pagination, the number of results to return.
+   * @param  {object} options.bookmark  For cursor-based pagination, the index cursor.
```
nono, line 32: No documentation for `options.keys`. Is it on purpose?

packages/cozy-stack-client/src/OAuthClientsCollection.js
```diff
@@ -0,0 +1,122 @@
...
+   *
+   * @param  {string} id The client id.
+   * @returns {object}  JsonAPI response containing normalized client as data attribute
+   */
+  async get(id) {
+    let resp = await this.all({ keys: [id] })
+
+    while (resp && resp.next) {
+      resp = await this.all({ keys: [id], bookmark: resp.bookmark })
+    }
```
nono, line 87: Why those 3 lines?

packages/cozy-stack-client/src/OAuthClientsCollection.js
taratatach: Hum no, I forgot to add it after I copy/pasted it.

packages/cozy-stack-client/src/OAuthClientsCollection.js
taratatach: Since I don't have any way to fetch a single OAuth client from the Cozy, I need to read through all the pages until I find the one I'm looking for or I've read all of them and none of the clients had the requested id.

packages/cozy-stack-client/src/OAuthClientsCollection.js
```diff
@@ -0,0 +1,122 @@
...
+        meta,
+        next: keys.length > data.length && nextBookmark != null,
+        bookmark: nextBookmark || undefined
+      }
+    } else {
+      return {
+        data: resp.data.map(c => normalizeOAuthClient(c)),
+        meta: resp.meta,
+        next: nextBookmark != null,
+        bookmark: nextBookmark || undefined
```
Ldoppea, line 71: nit: as this code is repeated I would have introduced a `hasBookmark` var just after `nextBookmark` declaration. And also i would have converted `nextBookmark` to undefined directly on its initialization. This would avoid usage of `null` and `undefined` that differs depending on the context.

Ldoppea (approved):
Looks good to me. But I did not write any client collection before so my review is not really valuable.

packages/cozy-stack-client/src/OAuthClientsCollection.js
Ldoppea: Those lines also puzzle me. The `while` condition implies that resp may be undefined. But on line 89 you check `resp.data.length` without checking its nullity first.

Is that scenario possible?

taratatach: `resp` is not supposed to be `undefined` indeed (`all()` only returns objects).
This could be simplified into `while(resp.next)`.

PR #1047, opened 2021-10-08
Fix OAuth clients' realtime updates

taratatach:
Using the `realtime` plugin and `RealTimeQueries` component with `io.cozy-oauth.clients` documents would not work for 2 reasons:
1. the realtime subscriptions would always set documents' doctype to `io.cozy.files`, no matter what data they would receive
2. the `OAuthClientsCollection.destroy()` method would expect cozy-stack to return data while the `DELETE /settings/clients/:id` route only returns a response header when successful

### DIFF
--- a/packages/cozy-client/types/models/permission.d.ts
+++ b/packages/cozy-client/types/models/permission.d.ts
@@ -52,7 +52,7 @@ export type Document = {
     _type: string;
     type: string;
 };
-export type PermissionVerb = "DELETE" | "ALL" | "GET" | "PATCH" | "POST" | "PUT";
+export type PermissionVerb = "DELETE" | "PUT" | "POST" | "GET" | "ALL" | "PATCH";
 export type PermissionItem = {
     /**
      * - ALL, GET, PUT, PATCH, DELETE, POST…
--- a/packages/cozy-client/types/types.d.ts
+++ b/packages/cozy-client/types/types.d.ts
@@ -6,7 +6,9 @@ export type KonnectorsDoctype = "io.cozy.konnectors";
 export type NotesDoctype = "io.cozy.notes";
 export type AppsDoctype = "io.cozy.apps";
 export type SettingsDoctype = "io.cozy.settings";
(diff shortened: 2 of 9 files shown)


packages/cozy-stack-client/src/OAuthClientsCollection.js
```diff
@@ -104,17 +104,13 @@ class OAuthClientsCollection extends DocumentCollection {
...
-   * @param {string}                client._id The client's id
-   *
    * @returns {{ data }} The deleted client
    */
-  async destroy({ _id }) {
-    const resp = await this.stackClient.fetchJSON(
-      'DELETE',
-      uri`/settings/clients/${_id}`
-    )
+  async destroy(client) {
```
paultranvan, line 109: nit: `client` is generally referring to a cozy-client instance, I think it would avoid confusion to name the parameter `oauthClient` for instance ;)

packages/cozy-client/types/models/permission.d.ts
```diff
@@ -52,7 +52,7 @@ export type Document = {
     _type: string;
     type: string;
 };
-export type PermissionVerb = "DELETE" | "ALL" | "GET" | "PATCH" | "POST" | "PUT";
+export type PermissionVerb = "DELETE" | "PUT" | "POST" | "GET" | "ALL" | "PATCH";
```
nono, line 55: Why this change?

packages/cozy-client/types/types.d.ts
```diff
@@ -193,6 +193,27 @@ export type CordovaWindow = {
...
+     */
+    _id?: string;
+    /**
+     * - Current revision of the document
+     */
+    _rev?: string;
+    /**
+     * - When the document has been deleted
+     */
+    _deleted?: string;
```
nono, line 211: Is it a string or a bool?

taratatach: Good question. I copy/pasted the `CozyClientDocument` type but it's strange to see a `string` type for this attribute.
I'll check it out.

taratatach: It's definitely a boolean :)

PR #1049, opened 2021-10-13
refactor: Move getFileTypeFromName in FileCollection

paultranvan:
We move this method here because it can be useful for apps such as Drive
in order to infer a mime-type from its name. Typically, this is
necessary when a file is encrypted and the apps needs to know its
decrypted type so it can display it in the appropriate viewer.


Merkur39 (approved):
Looks seems good

PR #571, opened 2019-11-22
[WIP] feat: FilesModel

Crash--:
- Not working yet, but will finish it next week 

- [ ] refactor uploadFileWithConflictStrategy to pass object instead of args

PR #1051, opened 2021-10-15
Relocate file model methods

paultranvan:
We move file model methods from https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/File.js and
https://github.com/cozy/cozy-libs/blob/master/packages/cozy-scanner/src/ScannerUpload.js

Note this is mainly based on https://github.com/cozy/cozy-client/pull/571/ with tests fixed and some refactoring.

### DIFF
--- a/packages/cozy-client/src/models/file.js
+++ b/packages/cozy-client/src/models/file.js
@@ -415,15 +415,15 @@ export const overrideFileForPath = async (client, dirPath, file, metadata) => {
  */
 export const generateNewFileNameOnConflict = filenameWithoutExtension => {
   //Check if the string ends by _1
-  const regex = new RegExp('(_)([0-9]+)$')
+  const regex = new RegExp('(?<=_)([0-9]+)$') // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Assertions#other_assertions
   const matches = filenameWithoutExtension.match(regex)
   if (matches) {
-    let versionNumber = parseInt(matches[2])
+    let versionNumber = parseInt(matches[1])
     //increment versionNumber
     versionNumber++
     const newFilenameWithoutExtension = filenameWithoutExtension.replace(
-      new RegExp('(_)([0-9]+)$'),
-      `_${versionNumber}`
+      regex,
+      `${versionNumber}`
     )
     return newFilenameWithoutExtension
   } else {


packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
...
+}
+/**
+ * Method to generate a new filename if there is a conflict
+ *
+ * @param {string} filenameWithoutExtension A filename without the extension
+ * @returns {string} A filename with the right suffix
+ */
+export const generateNewFileNameOnConflict = filenameWithoutExtension => {
+  //Check if the string ends by _1
+  const regex = new RegExp('(_)([0-9]+)$')
```
nono, line 417: You may want a look-behind assertion for matching the `_`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Assertions#other_assertions

packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
...
+    return `${filenameWithoutExtension}_1`
+  }
+}
+
+/**
+ * Generate a file name for a revision
+ *
+ * @param {IOCozyFile} file  - io.cozy.files document
+ * @param {object} revision - The revision containing the updated_at
+ * @param {function} f - A function to apply to generate the name
```
nono, line 438: The `f` function should be explained a bit more: what are its arguments?

packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
...
+/**
+ *
+ * Method to upload a file even if a file with the same name already exists.
+ *
+ * @param {CozyClient} client - The CozyClient instance
+ * @param {string} pathArg Fullpath for the file ex: path/to/
+ * @param {object} file HTML Object file
+ * @param {Object} metadata An object containing the wanted metadata to attach
+ */
+export const overrideFileForPath = async (client, pathArg, file, metadata) => {
```
nono, line 382: May I suggest `overwrite` instead of `override`?

packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
...
+}
+
+/**
+ *
+ * Method to upload a file even if a file with the same name already exists.
+ *
+ * @param {CozyClient} client - The CozyClient instance
+ * @param {string} pathArg Fullpath for the file ex: path/to/
+ * @param {object} file HTML Object file
+ * @param {Object} metadata An object containing the wanted metadata to attach
```
nono, line 380: No `@returns`?

packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
...
+  }
+}
+
+/**
+ *
+ * Method to upload a file even if a file with the same name already exists.
+ *
+ * @param {CozyClient} client - The CozyClient instance
+ * @param {string} pathArg Fullpath for the file ex: path/to/
+ * @param {object} file HTML Object file
```
nono, line 379: What is an `HTML Object file`?

packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
...
+    }
+  }
+}
+
+/**
+ *
+ * Method to upload a file even if a file with the same name already exists.
+ *
+ * @param {CozyClient} client - The CozyClient instance
+ * @param {string} pathArg Fullpath for the file ex: path/to/
```
nono, line 378: For me, the fullpath of a file is with the name (like `/path/to/file.txt`), not something that ends with `/`

packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
...
+      })
+      return createdFile
+    }
+    throw error
+  }
+}
+/**
+ * Method to generate a new filename if there is a conflict
+ *
+ * @param {string} filenameWithoutExtension A filename without the extension
```
nono, line 412: I really don't understand when the `@param` line should have a `-` and when it shouldn't

packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
...
 export const hasMetadataAttribute = ({ file, attribute }) =>
   has(file, `metadata.${attribute}`)
+
+/**
+ * async getFullpath - Gets a file's path
+ *
+ * @param {CozyClient} client - The CozyClient instance
+ * @param {string} dirId  - The id of the parent directory
+ * @param {string} name  - The file's name
+ * @returns {Promise<string>}       The full path of the file in the cozy
```
nono, line 302: Why so many spaces between `}` and `The`?

packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
  */
 export const hasMetadataAttribute = ({ file, attribute }) =>
   has(file, `metadata.${attribute}`)
+
+/**
+ * async getFullpath - Gets a file's path
+ *
+ * @param {CozyClient} client - The CozyClient instance
+ * @param {string} dirId  - The id of the parent directory
+ * @param {string} name  - The file's name
```
nono, line 301: [nit] I would have made this function take two arguments the client, and the file: `getFullPath(client, { dirId, name })`

packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
...
+
+/**
+ * Move file to destination.
+ *
+ * @param {CozyClient} client - The CozyClient instance
+ * @param   {string} fileId               - The file's id (required)
+ * @param   {object} destination
+ * @param   {string} destination.folderId - The destination folder's id (required)
+ * @param   {string} destination.path     - The file's path after the move (optional, used to optimize performance in case of conflict)
+ * @param   {boolean} force                - Whether we should overwrite the destination in case of conflict (defaults to false)
```
nono, line 324: [nit] this `-` is not aligned with the others `-`

packages/cozy-client/src/models/file.js
```diff
@@ -291,3 +292,268 @@ export const isReferencedByAlbum = file => {
...
+ *
+ * @param {IOCozyFile} file  - io.cozy.files document
+ * @param {object} revision - The revision containing the updated_at
+ * @param {function} f - A function to apply to generate the name
+ */
+export const generateFileNameForRevision = (file, revision, f) => {
+  const { filename, extension } = splitFilename(file)
+  return `${filename}_${f(
+    revision.updated_at,
+    'DD MMMM - HH[h]mm'
```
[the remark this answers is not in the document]
paultranvan: No idea, this is taken from https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/File.js#L159

packages/cozy-client/src/models/file.js
paultranvan: As this method already exist in https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/File.js#L101 , I feel like this will easier to stick with the same name for future migrations?

packages/cozy-client/src/models/file.js
paultranvan: This is taken from here: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/File.js#L98
I assume this a file uploaded in the browser

packages/cozy-client/src/models/file.js
paultranvan: I tend to agree, this looks more like `dirPath`

packages/cozy-client/src/models/file.js
paultranvan: This is relevant, but I think it's better to keep the current function to ease migration

Crash--: nit: this method should also be renamed to fetchFullPatch. But the work done in this PR is already 👍 so let's keep that in mind for the future

packages/cozy-client/src/models/file.js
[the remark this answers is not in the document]
Crash--: This is the desired format (by @poupotte 🤐 ) for displaying revisions in Driv
Capture d’écran 2021-10-21 à 09 25 05
e

PR #1062, opened 2021-11-01
fix: Use regex supported by Safari

Crash--:
This king of regex was not supported by Safari.
Let's revert it

This reverts commit 3d7cce2a8f7e185edfd86b137e13401dd561abcc.


Merkur39 (approved):
😍

PR #1063, opened 2021-11-02
feat: Add FileCollection.fetchChanges() method

taratatach:
We introduce a new method to fetch `io.cozy.files` changes from the
Cozy which does not go directly to CouchDB but uses the new
`cozy-stack` `/files/_changes` route.

This new route allows us to filter out trashed and deleted files from
the returned changes without using the CouchDB `selector` parameter
which can make queries really slow if the Cozy has lots of deleted
documents and thus time out.
This will be especially useful for the Desktop client.

This route has other benefits such as requesting specific fields or
for the path of files (i.e. as opposed to directories).
It also comes with limitations such as not being able to request
changes on specific documents via their ids.

This change is based on https://github.com/cozy/cozy-stack/pull/3200
and as such should probably not be merged until this PR makes its
way into production.

### DIFF
--- a/docs/api/cozy-stack-client.md
+++ b/docs/api/cozy-stack-client.md
@@ -157,6 +157,19 @@ Deleted and design docs are filtered by default, thus documents are retrieved in
 <dt><a href="#OAuthClient">OAuthClient</a> : <code>object</code></dt>
 <dd><p>Document representing a io.cozy.oauth.clients</p>
 </dd>
+<dt><a href="#FetchChangesReturnValue">FetchChangesReturnValue</a> ⇒ <code><a href="#FetchChangesReturnValue">FetchChangesReturnValue</a></code></dt>
+<dd><p>Use cozy-stack&#39;s _changes API for io.cozy.files
+Design docs are filtered by default, thus documents are retrieved in the
+response (includeDocs is set to true in the parameters of _changes).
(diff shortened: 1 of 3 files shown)


docs/api/cozy-stack-client.md
```diff
@@ -2113,6 +2126,45 @@ Updates the OAuth informations
...
+| since | <code>string</code> | Bookmark telling CouchDB from which point in time should changes be returned |
+| limit | <code>number</code> | The maximum number of returned documents for one call |
+| includeDocs | <code>boolean</code> | Whether or not complete documents should be returned |
+| fields | <code>Array.&lt;string&gt;</code> | The list of fields that should be returned for each document |
+| includeFilePath | <code>boolean</code> | Whether to include the path of file changes (needs includeDocs to be true) |
+| skipDeleted | <code>boolean</code> | Whether to skip changes for deleted documents |
+| skipTrashed | <code>boolean</code> | Whether to skip changes for trashed documents (needs includeDocs to be true) |
+| newLastSeq | <code>string</code> |  |
+| pending | <code>boolean</code> |  |
+| documents | <code>Array.&lt;object&gt;</code> |  |
```
nono, line 2166: It's a bit weird to mix the properties for the request and for the response

nono (approved):

PR #1064, opened 2021-11-02
fix: Return meta after file's referenced_by update

taratatach:
After updating a file's `referenced_by` attribute, `cozy-stack` will
not only return the updated file but also a `meta` object containing
the file's new revision and the file's new references count.

Fixes a regression introduced by https://github.com/cozy/cozy-client/pull/1048


paultranvan (approved):
Thanks!

Crash-- (approved):

Crash-- (approved):

PR #1065, opened 2021-11-03
fix: Build new OAuth client from old OAuth client

taratatach:
We offer 2 methods to build new `CozyClient` instances from an old
client built by `cozy-client-js`:
- `CozyClient.fromOldClient` to build WebApp clients
- `CozyClient.fromOldOAuthClient` to build OAuth clients

However, the options we passed to the `CozyClient` constructor when
building an OAuth client were not adequate and we ended up with a
WebApp client.
This would work most of the time though because of the way we make
requests to the Cozy and the fact that the OAuth token was still
passed but it would fail once we tried to refresh the token.

We're now extracting the OAuth credentials and scope from the old
client to instantiate the new one properly.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -293,6 +293,7 @@ class CozyClient {
    * a client with a cookie-based instance of cozy-client-js.
    *
    * @param {OldCozyClient} oldClient - An instance of the deprecated cozy-client
+   * @param {object} options - CozyStackClient options
    * @returns {CozyClient}
    */
   static fromOldClient(oldClient, options) {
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -309,21 +310,26 @@ class CozyClient {
...
-    const hasOauthCreds = oldClient._oauth && oldClient._authcreds != null
-    if (hasOauthCreds) {
-      const token = (await oldClient._authcreds).token
+    if (oldClient._oauth) {
+      const credentials = await oldClient.authorize()
+      const oauthOptions = {
+        oauth: credentials.client,
+        token: credentials.token,
+        scope: credentials.token.scope,
+        ...options
```
Ldoppea, line 324: nit:
Why did you move `...option` into `oauthOptions`?

Result is the same as when put in `CozyClient` constructor, but here it seems like `options` should only contain oauth related properties. Is that wanted?

packages/cozy-client/src/CozyClient.js
```diff
@@ -309,21 +310,26 @@ class CozyClient {
    *
    * Warning: unlike other instantiators, this one needs to be awaited.
    *
-   * @param {OldCozyClient} oldClient - An instance of the deprecated cozy-client
+   * @param {OldCozyClient} oldClient - An OAuth instance of the deprecated cozy-client
+   * @param {object} options - CozyStackClient options
    * @returns {Promise<CozyClient>} An instance of a client, configured from the old client
    */
   static async fromOldOAuthClient(oldClient, options) {
-    const hasOauthCreds = oldClient._oauth && oldClient._authcreds != null
```
Ldoppea, line 316: There is still a reference to `_authcreds` in `CozyClient.spec.js` in `can be instantiated from an oauth-based old client` test.

As I don't see any other ref of it, maybe we should remove this prop from the test?

Ldoppea (approved):

packages/cozy-client/src/CozyClient.js
taratatach: It's already been removed as part of a force push ;-)

PR #1073, opened 2021-11-24
Feat/throw error

flohhhh:
We can now pass an onError callback to CozyClient.
This callback can be defined globally when creating
the client

new CozyClient({..., onError: (e) => alert(e)})
or with client.setOnError(cb)

Or it can be defined locally:
client.query(QDef, {..., onError: (e) => alert(e)})
useQuery(QDef, {..., onError: (e) => alert(e)})

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -102,6 +102,7 @@ const DOC_UPDATE = 'update'
  * @property {boolean} [autoHydrate]
  * @property {object} [oauth]
  * @property {Function} [onTokenRefresh]
+ * @property {Function} [onError] - Default callback if a query is errored
  * @property  {Link}         [link]   - Backward compatibility
  * @property  {Array<Link>}  [links]  - List of links
  * @property  {object}       [schema] - Schema description for each doctypes
@@ -919,7 +920,15 @@ client.query(Q('io.cozy.bills'))`)
       return response
     } catch (error) {
       this.dispatch(receiveQueryError(queryId, error))
-      throw error
+      // specific onError
+      if (options.onError) {
+        options.onError(error)
+        // defaulted onError
+      } else if (this.options.onError) {
+        this.options.onError(error)
+      } else {
+        throw error
+      }
     }
   }
 
@@ -1589,6 +1598,18 @@ instantiation of the client.`
     Object.entries(data).forEach(([doctype, data]) => {
       this.dispatch(receiveQueryResult(null, { data }))
     })
+  }
+
+  /**
+   * At any time put an error function
+   *
+   * @param {Function} [onError] - Set a callback for queries which are errored
+   */
+  setOnError(onError) {
+    if (this.options && this.options.onError) {
+      throw new Error('On Error is already defined')
+    }
+    this.options.onError = onError
   }
 
   toJSON() {
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -102,7 +102,6 @@ const DOC_UPDATE = 'update'
  * @property {boolean} [autoHydrate]
  * @property {object} [oauth]
  * @property {Function} [onTokenRefresh]
- * @property {Function} [onTokenRefresh]
  * @property  {Link}         [link]   - Backward compatibility
  * @property  {Array<Link>}  [links]  - List of links
  * @property  {object}       [schema] - Schema description for each doctypes
--- a/packages/cozy-client/src/CozyClient.spec.js
+++ b/packages/cozy-client/src/CozyClient.spec.js
@@ -162,7 +162,6 @@ describe('CozyClient initialization', () => {
       })
     }
     const client = await CozyClient.fromOldOAuthClient(oldClient)
-    console.log({ stackClient: client.stackClient })
     expect(client.stackClient.uri).toBe(url)
     expect(client.stackClient.oauthOptions.clientID).toBe(oauth.clientID)
     expect(client.stackClient.token.accessToken).toBe(token.accessToken)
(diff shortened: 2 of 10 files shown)


packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -1402,6 +1401,42 @@ describe('CozyClient', () => {
...
     })
+
+    const setupOnError = () => {
+      const error = new TypeError('Failed to fetch')
+      requestHandler.mockReturnValueOnce(Promise.reject(error))
+      const onError = jest.fn().mockReturnValueOnce(() => true)
+      return onError
+    }
+
+    it('should call onError callback when an error is catched. It is passed as argument', async () => {
```
Crash--, line 1412: should call onError callback when an error is thrown. Callback passed as argument

packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -1402,6 +1401,42 @@ describe('CozyClient', () => {
...
+      const onError = setupOnError()
+
+      try {
+        await client.query(query, { as: 'allTodos', onError })
+      } catch (e) {} // eslint-disable-line no-empty
```
Crash--, line 1417: Do we really need this try / catch block? Since we should not throw the error because we have have onError callback, I guess this block is not mandatory

packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -1402,6 +1401,42 @@ describe('CozyClient', () => {
...
+
+    it('should call onError callback when an error is catched. It is passed with setOnError', async () => {
+      const onError = setupOnError()
+      client.setOnError(onError)
+
+      try {
+        await client.query(query, { as: 'allTodos' })
+      } catch (e) {} // eslint-disable-line no-empty
+      expect(onError).toBeCalled()
+    })
```
Crash--, line 1439: We should remove the try / catch block when there is an `onError` callback. 

But we should add a test to test that if we don't have any onError callback the error is thrown and catchable.

packages/cozy-client/src/hooks/useQuery.js
```diff
@@ -21,15 +22,16 @@ const generateFetchMoreQueryDefinition = queryResult => {
...
+ * @property  {object} as - Name for the query [required]
+ * @property  {boolean} enabled - If set to false, the query won't be executed
+ * @property  {object} fetchPolicy - Fetch policy
+ * @property  {object} singleDocData - If true, the "data" returned will be
  * a single doc instead of an array for single doc queries. Defaults to false for backward
  * compatibility but will be set to true in the future.
+ * @property {Function} onError - Callback if the query is errored
  *
  * @returns {UseQueryReturnValue}
  */
```
Crash--, line 37: @flohhhh can you add a test for useQuery plz? 

Since we can do: 
```
useQuery(..., onError)
```

we should at least test this behavior. But we should also test the same behaviors than query itself. WDYT?

packages/cozy-client/src/CozyClient.spec.js
flohhhh: You are right, it's not mandatory and it's logical

packages/cozy-client/src/CozyClient.js
```diff
@@ -1592,6 +1600,15 @@ instantiation of the client.`
     })
   }
 
+  /**
+   * At any time put an error function
+   *
+   * @param {Function} [onError] - Set a callback for queries which are errored
+   */
+  setOnError(onError) {
+    this.options.onError = onError
```
Crash--, line 1609: I'm wondering if you should not test if onError is already defined and thrown an error or log.warn / error? In order to avoid to set several times?

packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -1402,6 +1401,43 @@ describe('CozyClient', () => {
...
+
+      await client.query(query, { as: 'allTodos' })
+      expect(onError).toBeCalled()
+    })
+
+    it('should thrown an error when there is no onError callback', async () => {
+      try {
+        await client.query(query, { as: 'allTodos' })
+      } catch (e) {} // eslint-disable-line no-empty
+      await client.query(query, { as: 'allTodos' })
```
Crash--, line 1439: I don't understand this test? 

await expect(client.query(query, { as: 'allTodos' })).rejects.toThrow();

packages/cozy-client/src/hooks/useQuery.spec.jsx
```diff
@@ -159,6 +166,29 @@ describe('use query', () => {
...
+      queryOptions: {
+        as: 'fake-query',
+        onError
+      },
+      queryDefinition: () => Q('io.cozy.todos')
+    })
+
+    setTimeout(() => {
+      expect(onError).toBeCalled()
+    }, 100)
```
Crash--, line 190: This is alway a bad idea to rely on timeout. 

```
it('should query through the client in context', () => {
    const { client } = setupQuery({
      queryOptions: {
        as: 'simpsons',
        onError
      },
      queryDefinition: () => Q('io.cozy.simpsons')
    })
    expect(onError).toHaveBeenCalled()
  })
```

this doesn't work?

flohhhh: I know, but it doesn't work any other way

packages/cozy-client/src/CozyClient.spec.js
flohhhh: Oops, an oversight

packages/cozy-client/src/CozyClient.js
flohhhh: Yes it would make sense to throw an error, we don't really need to want to set it several times

Crash-- (approved):
If you can add a todo in the test to say we should find something better one day...

PR #1085, opened 2021-12-08
fix(filename): Get illegal characters from file name and returns them

trollepierre:
To alert in Drive when filename contains illegal characters instead of 'the name already exists'

[waiting for equivalent MR in cozy-drive](https://github.com/cozy/cozy-drive/pull/2442)

### DIFF
--- a/packages/cozy-client/src/models/file.js
+++ b/packages/cozy-client/src/models/file.js
@@ -22,7 +22,7 @@ const FILENAME_WITH_EXTENSION_REGEX = /(.+)(\..*)$/
  * @returns {object}  {filename, extension}
  */
 export const splitFilename = file => {
-  if (!isString(file.name)) throw new Error('file should have a name property ')
+  if (!isString(file.name)) throw new Error('file should have a name property')
 
   if (file.type === 'file') {
     const match = file.name.match(FILENAME_WITH_EXTENSION_REGEX)
@@ -318,7 +318,7 @@ export const getFullpath = async (client, dirId, name) => {
  *
  * @param {CozyClient} client             - The CozyClient instance
  * @param   {string} fileId               - The file's id (required)
- * @param   {object} destination
+ * @param   {object} destination          - The destination object containing:
  * @param   {string} destination.folderId - The destination folder's id (required)
  * @param   {string} destination.path     - The file's path after the move (optional, used to optimize performance in case of conflict)
  * @param   {boolean} force               - Whether we should overwrite, i.e. put to trash, the destination in case of conflict (defaults to false).
--- a/packages/cozy-client/src/models/file.spec.js
+++ b/packages/cozy-client/src/models/file.spec.js
@@ -212,7 +212,9 @@ describe('File Model', () => {
     }
 
     it('should throw an error if the file is not correct', () => {
-      expect(() => fileModel.splitFilename({})).toThrow()
+      expect(() => fileModel.splitFilename({})).toThrow(
+        'file should have a name property'
+      )
       expect(() => fileModel.splitFilename({ name: null })).toThrow()
       expect(() => fileModel.splitFilename({ name: '' })).not.toThrow()
     })
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -9,6 +9,7 @@ import DocumentCollection, { normalizeDoc } from './DocumentCollection'
 import { uri, slugify, formatBytes } from './utils'
 import { FetchError } from './errors'
 import { dontThrowNotFoundError } from './Collection'
+import { getIllegalCharacters } from './getIllegalCharacter'
 
 /**
  * Attributes used for directory creation
@@ -67,18 +68,26 @@ const normalizeReferences = references => {
 
 const sanitizeFileName = name => name && name.trim()
 
-/**
- * Sanitize Attribute for an io.cozy.files
- *
- * Currently juste the name
- *
- * @param {FileAttributes|DirectoryAttributes} attributes - Attributes of the created file/directory
- * @returns {FileAttributes|DirectoryAttributes}
- */
-const sanitizeAttributes = attributes => {
-  if (attributes.name) attributes.name = sanitizeFileName(attributes.name)
-  return attributes
+const sanitizeAndValidateFileName = name => {
+  let safeName = sanitizeFileName(name)
+  if (typeof safeName !== 'string' || safeName === '') {
+    throw new Error('Missing name argument')
+  }
+
+  if (name === '.' || name === '..') {
+    throw new Error(`Invalid filename: ${name}`)
+  }
+
+  const illegalCharacters = getIllegalCharacters(safeName)
+  if (illegalCharacters.length) {
+    throw new Error(
+      `Invalid filename containing illegal character(s): ${illegalCharacters}`
+    )
+  }
+
+  return safeName
 }
+
 export const isFile = ({ _type, type }) =>
   _type === 'io.cozy.files' || type === 'directory' || type === 'file'
 
@@ -384,6 +393,7 @@ class FileCollection extends DocumentCollection {
    *
    * @param {FileAttributes|DirectoryAttributes} attributes - Attributes of the created file/directory
    * @param {File|Blob|string|ArrayBuffer} attributes.data Will be used as content of the created file
+   * @throws {Error} - explaining reason why creation failed
    */
   async create(attributes) {
     if (attributes.type === 'directory') {
@@ -397,8 +407,9 @@ class FileCollection extends DocumentCollection {
   /***
    * Update the io.cozy.files
    * Used by StackLink to support CozyClient.save({file})
-   * @param {FileAttributes}  The file with its new content
+   * @param {FileAttributes} file - The file with its new content
    * @returns {FileAttributes} Updated document
+   * @throws {Error} - explaining reason why update failed
    */
 
   async update(file) {
@@ -412,6 +423,7 @@ class FileCollection extends DocumentCollection {
    * @param {File|Blob|Stream|string|ArrayBuffer} data file to be uploaded
    * @param {FileAttributes} params Additional parameters
    * @param  {object}  params.options     Options to pass to doUpload method (additional headers)
+   * @throws {Error} - explaining reason why creation failed
    */
   async createFile(
     data,
@@ -429,10 +441,9 @@ class FileCollection extends DocumentCollection {
     if (!name && typeof data.name === 'string') {
       name = data.name
     }
-    name = sanitizeFileName(name)
-    if (typeof name !== 'string' || name === '') {
-      throw new Error('missing name argument')
-    }
+
+    name = sanitizeAndValidateFileName(name)
+
     if (executable === undefined) {
       executable = false
     }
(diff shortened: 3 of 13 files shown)


packages/cozy-client/src/models/file.js
```diff
@@ -115,6 +115,29 @@ export function normalize(file) {
   return { _id: id, id, _type: doctype, ...file }
 }
 
+/**
+ * Get the list of illegal characters in the file name
+ *
+ * Ensures existence of `_id` and `_type`
+ *
+ * @public
+ * @param {object} file - object representing the file
```
Crash--, line 124: import { IOCozyFile } from '../../types' (something like that)

and then use the `IOCozyFile` as type's param ;)

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -386,6 +388,14 @@ class FileCollection extends DocumentCollection {
    * @param {File|Blob|string|ArrayBuffer} attributes.data Will be used as content of the created file
    */
   async create(attributes) {
+    if (getIllegalCharacters(attributes).length) {
+      throw new Error(
+        `Invalid filename containing illegal character(s): ${getIllegalCharacters(
+          attributes
+        )}`
+      )
+    }
```
Crash--, line 397: Since we can throw, Is there any point in adding a @throw property to the JSDoc?

packages/cozy-client/src/models/file.js
trollepierre: we won't implement anything in the cozy-client package. This thread is outdated

packages/cozy-stack-client/src/FileCollection.js
trollepierre: I updated the whole file, thx

packages/cozy-stack-client/src/getIllegalCharacter.spec.js
```diff
@@ -0,0 +1,24 @@
...
+
+  it('should return empty string when no illegal characters found', () => {
+    const name =
+      'Valid file name with #@&"\'()§è!çà)-°_^¨*$€ù%`£=+:;.,?∞…÷≠≤<>•ë“‘{¶«Çø}æê®†Úºîœπ‡Ò∂ƒﬁÌÏÈ¬µ‹≈©◊ß~ '
+    expect(getIllegalCharacters({ name })).toEqual('')
+  })
+
+  it('should return illegal characters split with spaces when illegal characters found', () => {
+    const name = 'Invalid file name with / and / and \x00 and \n and \r'
+    expect(getIllegalCharacters({ name })).toEqual('/ / \x00 \n \r')
```
Ldoppea, line 22: `\x00`, `\n` and `\r` won't be visible in the user interface, or will they?

Maybe we should escape them first?

packages/cozy-stack-client/src/getIllegalCharacter.spec.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+  })
+
+  it('should return empty string when no illegal characters found', () => {
+    const name =
+      'Valid file name with #@&"\'()§è!çà)-°_^¨*$€ù%`£=+:;.,?∞…÷≠≤<>•ë“‘{¶«Çø}æê®†Úºîœπ‡Ò∂ƒﬁÌÏÈ¬µ‹≈©◊ß~ '
+    expect(getIllegalCharacters({ name })).toEqual('')
+  })
+
+  it('should return illegal characters split with spaces when illegal characters found', () => {
```
Ldoppea, line 20: nit: can we have a test method for each individual invalid char (in addition to this one)?

This would ensure that the regex does not ignore one of them (if we change it in the future).

Current test would pass even if the regex tests only for `\n`. But it tests results' concatenation, which is great, so do not remove it.

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -263,7 +264,7 @@ class FileCollection extends DocumentCollection {
    *
    * @param  {object} document        A JSON representing a document, with at least a `_type` and `_id` field.
    * @param  {Array}  documents       An array of JSON files having an `_id` field.
-   * @returns 204 No Content
+   * @returns {number} 204 No Content
```
Ldoppea, line 267: Does this mean that doing a "no content" http request would return `204` as a number? Shouldn't it return a http result or maybe void? (I did not use cozy-client enough yet, so this is a real question, I don't know what is the actual behavior)

packages/cozy-stack-client/src/getIllegalCharacter.js
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Get the list of illegal characters in the file name
+ *
+ * @public
+ * @param {object} file - object representing the file
+ * @returns {string} illegal characters separated by spaces
+ */
+export const getIllegalCharacters = file => {
+  if (!isString(file.name)) throw new Error('file should have a name property ')
```
Ldoppea, line 11: Is that trailing space wanted?
```suggestion
  if (!isString(file.name)) throw new Error('file should have a name property')
```

Ldoppea: I would choose "consistency by fixing existing code".

This seems to be a typo, I don't find anything in the code that would require that space. @Crash-- are you aware of any Cozy practice that would imply adding a trailing space on error messages?

Crash--: I think that the trailing space was not wanted at the first place. 

Feel free to update it ;)

packages/cozy-stack-client/src/FileCollection.js
trollepierre: `If a function that is not in externs has no return value, you can omit the @return tag, and the compiler will assume that the function returns undefined.`

according to https://github-wiki-see.page/m/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#return-type-description

trollepierre: I will update the MR and remove the @returns tag

Crash--: We added this return because it's a bit special, we had several regression on it, so we added that this method returns a 204 without any content https://github.com/cozy/cozy-client/commit/917ebb576f5d199ba596039285c9f074da79e13b

packages/cozy-stack-client/src/getIllegalCharacter.spec.js
trollepierre: do you want me to add a backslash \ in front of them?

trollepierre: let's resolve this

packages/cozy-stack-client/src/FileCollection.js
trollepierre: I will not touch to these returns.

packages/cozy-stack-client/src/getIllegalCharacter.spec.js
Ldoppea: Choice made after a call : `\x00`, `\n` and `\r` seem to be impossible to insert from the UI using classic way (i.e copy/paste correctly sanitizes input), so we think it would be useless to spend time to find a way to display them in the UI.

However the question may come back later if we add additional invisible forbidden characters.

packages/cozy-client/src/models/file.spec.js
```diff
@@ -215,6 +215,11 @@ describe('File Model', () => {
       expect(() => fileModel.splitFilename({})).toThrow()
       expect(() => fileModel.splitFilename({ name: null })).toThrow()
       expect(() => fileModel.splitFilename({ name: '' })).not.toThrow()
+      try {
+        fileModel.splitFilename({})
+      } catch (error) {
+        expect(error.message).toEqual('file should have a name property')
```
Ldoppea, line 221: Care, here if `splitFileName` does not throw, then the test will still pass because the expect won't be called. Prefere to use `.toThrow('message')` instead.

packages/cozy-stack-client/src/FileCollection.spec.js
```diff
@@ -454,9 +519,58 @@ describe('FileCollection', () => {
...
+
+    it('should throw error when file name is forbidden', async () => {
+      try {
+        await collection.updateAttributes('42', {
+          dir_id: '123',
+          name: '.'
+        })
+      } catch (error) {
+        expect(error.message).toEqual(
+          'Invalid filename containing illegal character(s): .'
```
Ldoppea, line 554: Would it be possible to have a different error message here? `.` is a legal character but is an illegal name.

packages/cozy-stack-client/src/ShortcutsCollection.js
```diff
@@ -14,16 +15,23 @@ class ShortcutsCollection extends DocumentCollection {
...
    * @param {string} attributes.url Shortcut's URL
    * @param {string} attributes.dir_id dir_id where to create the shortcut
+   * @throws {Error} - explaining reason why creation failed
    */
   async create(attributes) {
     if (!attributes.type) {
       attributes.type = SHORTCUTS_DOCTYPE
     }
-    if (!attributes.name || !attributes.url || !attributes.dir_id) {
+    if (!attributes.name.trim() || !attributes.url || !attributes.dir_id) {
```
Ldoppea, line 24: `attributes.name` may be undefined so you should check for nullity first before calling `.trim()` on it.

Ldoppea (changes requested):
There are still a few change requests, but you are close to the end 💪

packages/cozy-stack-client/src/getIllegalCharacter.js
```diff
@@ -0,0 +1,23 @@
...
+
+  let illegalCharacters = ''
+
+  const FILENAME_ILLEGAL_CHARACTERS = /\/|\x00|\n|\r/g // eslint-disable-line no-control-regex
+  const match = file.name.match(FILENAME_ILLEGAL_CHARACTERS)
+  if (match) {
+    match.forEach(illegalCharacter => {
+      illegalCharacters += illegalCharacter + ' '
+    })
```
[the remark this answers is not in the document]
trollepierre: done

packages/cozy-stack-client/src/ShortcutsCollection.js
```diff
@@ -14,16 +15,32 @@ class ShortcutsCollection extends DocumentCollection {
...
+   * @throws {Error} - explaining reason why creation failed
    */
   async create(attributes) {
     if (!attributes.type) {
       attributes.type = SHORTCUTS_DOCTYPE
     }
-    if (!attributes.name || !attributes.url || !attributes.dir_id) {
+    if (
+      !attributes.name ||
+      !attributes.name.trim() ||
```
Crash--, line 26: why this specific line?

Crash-- (approved):

packages/cozy-stack-client/src/ShortcutsCollection.js
trollepierre: When name was "    " , we used to send no error.

trollepierre: https://github.com/cozy/cozy-client/pull/1085/files#diff-4d747c875bb3490e6493ca37186847024c987742d081ad7be03b86676097241cR82

Ldoppea (approved):

PR #1090, opened 2021-12-17
fix: Use fetchOwnPermissions instead of deprecated getOwnPermissions

trollepierre:
- prevent warnings in app consuming permission

### DIFF
--- a/packages/cozy-client/src/models/permission.js
+++ b/packages/cozy-client/src/models/permission.js
@@ -51,7 +51,7 @@ export function isReadOnly(perm, options = {}) {
  */
 export async function fetchOwn(client) {
   const collection = client.collection('io.cozy.permissions')
-  const data = await collection.getOwnPermissions()
+  const data = await collection.fetchOwnPermissions()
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/models/permission.js
```diff
@@ -51,7 +51,7 @@ export function isReadOnly(perm, options = {}) {
  */
 export async function fetchOwn(client) {
   const collection = client.collection('io.cozy.permissions')
-  const data = await collection.getOwnPermissions()
+  const data = await collection.fetchOwnPermissions()
```
Crash--, line 54: This commit can not be tagged as chore. Chore is only about build stuff so it’ll not publish a new version . Here we: 
- edited a file published and used in our apps 
- so we want a new version 
 Let’s tag it as a fix or refactor should work too but not sure (and I can’t check today)

trollepierre: Only fix will trigger a patch => https://www.conventionalcommits.org/en/v1.0.0/#summary

Crash-- (approved):

PR #1092, opened 2021-12-20
Add of missing papers targeted for v1

Crash--:


Merkur39:
You should name your commit like this: `feat: Add of missing papers targeted for v1`

packages/cozy-client/src/models/document/locales/en.json
```diff
@@ -19,7 +19,7 @@
     "items": {
       "identity": "Identity",
       "family": "Family",
-      "work_study": "Work",
+      "work_study": "Work / Study",
```
Merkur39, line 22: I'm not a big fan of using this kind of character ('/') here.
These names can be used in file names, and this kind of character will cause some conflicts.
Maybe the word "or" or an underscore ?

Merkur39:
See https://github.com/cozy/cozy-client/pull/1093

PR #1094, opened 2021-12-28
Feat/flagship certification

Ldoppea:
Add ability to register to `cozy-stack` as a Cozy flagship app

Consumer app can now init `cozy-client` with
`oauth.shouldRequireFlagshipPermissions` set to `true` in order to
register itself as a flagship app

When doing this, the OAuth process can now query respective platform's
stores in order to get an attestation of genuineness that can be
analysed by `cozy-stack`

This mechanism requires to be in a ReactNative environment

____

Related PR : https://github.com/cozy/cozy-react-native/pull/56
____

Todo:

- [x] Implement Android SafetyNet
- [x] Implement iOS DeviceCheck
- [x] Document the process
- [x] Drop commit `tests: TEMPORARY LOGS`

### DIFF
--- a/packages/cozy-client/package.json
+++ b/packages/cozy-client/package.json
@@ -67,7 +67,8 @@
     "react": "^16.7.0",
     "react-native": "^0.65.0",
     "react-native-google-safetynet": "0.0.4",
-    "react-native-inappbrowser-reborn": "^3.5.1"
+    "react-native-inappbrowser-reborn": "^3.5.1",
+    "react-native-ios11-devicecheck": "https://github.com/cozy/react-native-devicecheck#app-attest-v0.1"
   },
   "sideEffects": false
 }
--- a/packages/cozy-client/package.json
+++ b/packages/cozy-client/package.json
@@ -66,6 +66,7 @@
     "cozy-ui": ">34.0.0",
     "react": "^16.7.0",
     "react-native": "^0.65.0",
+    "react-native-google-safetynet": "0.0.4",
     "react-native-inappbrowser-reborn": "^3.5.1"
   },
   "sideEffects": false
--- a/packages/cozy-client/src/flagship-certification/README.md
+++ b/packages/cozy-client/src/flagship-certification/README.md
@@ -1,0 +1,86 @@
+# Flagship certification
+
+Flagship certification is the process of verifying that the current running app is a genuine Cozy application.
+
+This verification is done by querying an app certificate from the app store (Apple AppStore or Google Play) and sending  the resulting certificate to `cozy-stack`. Then `cozy-stack` will be able to an analyze the certificate and conclude if the app is genuine or not.
+
+## Nomenclature
+
+- `flagship`: refers to the Cozy main application
+- `store certification`: process of verifying app's genuineness through the platform's store 
+- `attestation`: result from the `store certification`
+- `challenge`: unique token given to the app by `cozy-stack` that may be encrypted in the `attestation` as a proof of authenticity
+- `nonce`: data type used to store the `challenge` token
+- `SafetyNet`: Google's implementation of the `store certification`
+- `AppAttest`: Apple's implementation of the `store certification`
+
+## Android certification
+
+Android certification is based on [SafetyNet](https://developer.android.com/training/safetynet/index.html).
+
+This process requires to query a `challenge` from `cozy-stack` and to use it to init the `store certification` process through `SafetyNet`. Then the received `attestation` is send to `cozy-stack` for verification.
+
+The resulting `attestation` is in the form of a `JSON Web Signature` that embbed the following `JSON`:
+```json
+{
+    "apkCertificateDigestSha256": [
+        "base64 encoded, SHA-256 hash of the certificate used to sign requesting app="
+    ],
+    "apkDigestSha256": "kNv83tJLFqwliYQ/6HUPCeGkBzLLCX/nvT+EF3OEB2I=",
+    "apkPackageName": "com.package.name.of.requesting.app",
+    "basicIntegrity": true,
+    "ctsProfileMatch": true, 
+    "evaluationType": "BASIC",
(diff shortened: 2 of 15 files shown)


packages/cozy-client/src/flagship-certification/README.md
```diff
@@ -0,0 +1,49 @@
...
+Flagship certification is the process of verifying that the current running app is a genuine Cozy application.
+
+This verification is done by querying an app certificate from the app store (Apple AppStore or Google Play) and sending  the resulting certificate to `cozy-stack`. Then `cozy-stack` will be able to an analyze the certificate and conclude if the app is genuine or not.
+
+## Nomenclature
+
+- `flagship`: refers to the Cozy main application
+- `store certification`: process of verifying app's genuineness through the platform's store 
+- `attestation`: result from the `store certification`
+- `challenge`: unique token given to the app by `cozy-stack` (or by the store in iOS) that may be encrypted in the `attestation` as a proof of authenticity
```
nono, line 12: s/(or by the store in iOS)//

packages/cozy-client/src/flagship-certification/README.md
```diff
@@ -0,0 +1,49 @@
...
+    "nonce": "R2Rra24fVm5xa2Mg",
+    "timestampMs": 9860437986543
+}
+```
+
+The `attestation`'s content is described in the SafetyNet's documentation: https://developer.android.com/training/safetynet/attestation#use-response-server
+
+## iOS certification
+
+iOS certification is based on [DeviceCheck](https://developer.apple.com/documentation/devicecheck).
```
nono, line 43: I'm not sure if we should speak of DeviceCheck or of AppAttest.

packages/cozy-client/src/flagship-certification/flagship-certification.js
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A JSON Web Signature
+ * @typedef {string} jws
+ */
+
+/**
+ * Request a challenge from the Stack that can be used to request the app attestation from the app store
+ * @param {any} stackClient - the CozyStackClient instance
+ * @returns {Promise<string>} - the Nonce string retruned by the stack
```
nono, line 11: s/retruned/returned/

nono (commented):
Are we sure we want to put this code in cozy-client? It will only be used for the flagship app, and it adds some dependencies.

Ldoppea:
I force-pushed the following:
- Refactored `getAppAttestationFromStore` to return the current platform, the attestation and an option KeyID (from iOS)
  - This allow `giveAppAttestationToStack` to be independent from the platform, it just send to the stack what was returned by `getAppAttestationFromStore`
- Refactored `certifyFlagship` to take a `certificationConfig` argument
  - For now it is used only to pass the `androidSafetyNetApiKey` from the flagship app (it should not be set directly on cozy-client's code)
- Added `typedefs.js` file for JSDoc types declaration
- Added `Client's configuration` and `Stacks' configuration` sections on the `README.md` documentation

packages/cozy-client/src/flagship-certification/flagship-certification.js
Ldoppea: Fixed 👍

packages/cozy-client/src/flagship-certification/README.md
Ldoppea: I kept the link to https://developer.apple.com/documentation/devicecheck but I now only refer to `AppAttest` (`DeviceCheck` is not mentioned anymore)

packages/cozy-client/src/flagship-certification/README.md
Ldoppea: Fixed 👍

Ldoppea:
> Are we sure we want to put this code in cozy-client? It will only be used for the flagship app, and it adds some dependencies.

I'm not sure how to extract this code from cozy-client. My initial understanding was that I had to put the code after `stackClient.register()` but before the `authorize()` step. But since our last conversation I'm not sure anymore that doing it before `authorize()` is a requirement.

A solution would be to inject a function pointer that would to the work on the flagship app.

However here is my opinion:
- This is fully related to cozy-client's OAuth registration, so it seems natural for me to have all related method in cozy-client's
  - Putting the code on the flagship side would segment the OAuth process too much (but this is a subjective point of view)
- The stack implementation uses arrays for all package names, certificates and app ids
  - This makes it easy to configure additional `cozy flagship` apps
  - So it seems more homogeneous to do a "generic" implementation on `cozy-client` too
  - Even if there is a few chances that we will do another flagship later, this would ease the work for auto-heberged cozy that would want to benefits from this feature on their own mobile app.
- This is a peerDependency, so until we are in a web context there is no need to install it. It will be required only in ReactNative contexts

Ldoppea:
I completed the doc to describe how to fill `apk_certificate_digests` using Android's local dev certificate: https://github.com/cozy/cozy-client/pull/1094/commits/b235e3f257cac4c2eabbfe010037d231ad3dceee

packages/cozy-client/package.json
```diff
@@ -66,7 +66,9 @@
     "cozy-ui": ">34.0.0",
     "react": "^16.7.0",
     "react-native": "^0.65.0",
-    "react-native-inappbrowser-reborn": "^3.5.1"
+    "react-native-google-safetynet": "0.0.4",
+    "react-native-inappbrowser-reborn": "^3.5.1",
+    "react-native-ios11-devicecheck": "0.0.3"
```
Crash--, line 71: should be our fork, right?

Crash-- (approved):

packages/cozy-client/package.json
Ldoppea: [Fixed](https://github.com/cozy/cozy-client/compare/1100dc38c7a24690f485b99a53b421a7d3e53b09..876e2ca52277c285a8e2bed5437fa6fa6ee58cc7)

PR #1097, opened 2022-01-10
feat: Add a method to get session_code

acezard:
Simple facade to get flagship app session_code without directly using the fetchJSON method

### DIFF
--- a/packages/cozy-stack-client/src/OAuthClient.js
+++ b/packages/cozy-stack-client/src/OAuthClient.js
@@ -365,6 +365,21 @@ class OAuthClient extends CozyStackClient {
   }
 
   /**
+   * @typedef SessionCodeRes
+   * @property {string} session_code The value of the session code
+   */
+
+  /**
+   * Fetches a new session code. Only usable by the Flagship application
+   *
+   * @throws {NotRegisteredException} When the client isn't certified to be the Flagship application
+   * @returns {Promise<SessionCodeRes>} A promise that resolves with a new session_code
+   */
+  async fetchSessionCode() {
+    return this.fetchJSON('POST', '/auth/session_code')
+  }
+
+  /**
    * Retrieves a new access token by refreshing the currently used token.
    *
    * @throws {NotRegisteredException} When the client doesn't have it's registration information
--- a/docs/api/cozy-stack-client.md
+++ b/docs/api/cozy-stack-client.md
@@ -181,6 +181,8 @@ not.</p>
 <dt><a href="#DesignDoc">DesignDoc</a> : <code>object</code></dt>
 <dd><p>Attributes representing a design doc</p>
 </dd>
+<dt><a href="#SessionCodeRes">SessionCodeRes</a></dt>
+<dd></dd>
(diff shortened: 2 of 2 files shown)


packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -364,6 +364,19 @@ class OAuthClient extends CozyStackClient {
...
   }
 
+  /**
+   * Fetches a new session code. Only usable by the Flagship application
+   *
+   * @throws {NotRegisteredException} When the client isn't certified to be the Flagship application
+   * @returns {Promise} A promise that resolves with a new session_code
+   */
+  async fetchSessionCode() {
+     return this.fetchJSON(
```
Crash--, line 374: not sure you have a direct access to fetchJSON.

Instead
```
this.stackClient.fetchJSON
```

should work

packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -364,6 +364,19 @@ class OAuthClient extends CozyStackClient {
     return new AccessToken(result)
   }
 
+  /**
+   * Fetches a new session code. Only usable by the Flagship application
+   *
+   * @throws {NotRegisteredException} When the client isn't certified to be the Flagship application
+   * @returns {Promise} A promise that resolves with a new session_code
```
Crash--, line 371: can you document the return please. I don't care that it's a promise, I care about the structure of the return ;)

Crash-- (changes requested):
see comment

packages/cozy-stack-client/src/OAuthClient.js
acezard: according to the implementation `class OAuthClient extends CozyStackClient `, `fetchJSON` should  be an inherited public method. Other methods in OAuthClient use this.fetchJSON() directly

packages/cozy-stack-client/src/OAuthClient.js
acezard: I'm not very used to JSdoc syntax, what do you think of the update?

packages/cozy-stack-client/src/OAuthClient.js
Crash--: Oh. We're in OAuthClient.js ! It's worse. 

OAuth client is there to handle... OAuthClient "connexion". Here this is not what you're doing. 

You are manipulating the OAuthClient doctype. You should definitively use `OAuthClientsCollection` ;)

Crash--: In fact I don't know exactly... 

You're not manipulating a doctype since you're calling the /auth/ route. And you're just creating a new code, not something we want to store in the redux store... 

OauthClient.js could be the right place, let me think about it

Crash--: Let's keep it there. 

Can you fix the CI please

acezard: @Crash--  it is done

Crash-- (approved):

PR #1098, opened 2022-01-10
fix: UseQuery returns fetch method

Crash--:
fetch was missing from the API. So making
a request from <Query or with queryConnect
didn't share the same API than useQuery.

So let's fix it.

https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/ObservableQuery.js#L49-L51

### DIFF
--- a/packages/cozy-client/src/hooks/useQuery.js
+++ b/packages/cozy-client/src/hooks/useQuery.js
@@ -84,7 +84,15 @@ const useQuery = (queryDefinition, options) => {
     return client.query(generateFetchMoreQueryDefinition(queryState), { as })
   }, [as, client])
 
-  return { ...queryState, fetchMore: fetchMore }
+  const fetch = useCallback(() => {
+    return client.query(definition, options)
+  }, [client, definition, options])
+
+  return {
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/hooks/useQuery.js
```diff
@@ -84,7 +84,15 @@ const useQuery = (queryDefinition, options) => {
...
 
-  return { ...queryState, fetchMore: fetchMore }
+  const fetch = useCallback(() => {
+    return client.query(definition, options)
+  }, [client, definition, options])
+
+  return {
+    ...queryState,
+    fetchMore: fetchMore,
+    fetch: fetch
```
trollepierre, line 94: Good change! Let's add a test to document this feature and ensure no regression 🙏

trollepierre (changes requested):

packages/cozy-client/src/hooks/useQuery.js
Crash--: added, thanks :)

trollepierre (approved):

PR #1093, opened 2021-12-22
feat: Add of missing papers targeted for v1

Merkur39:
Adding multiples missing Qualifications.
Adding information to multiples existing Qualifications.
Required for the "Mes papiers" application for the moment, in order to qualify more documents.

PR #1099, opened 2022-01-11
feat: Add activity theme

Merkur39:
Added `activity` theme & `other_activity_document` label.
Moved `personal_sporting_licence` label from `identity` theme to `activity` theme


Merkur39:
[Ajout d'un thème & label dans Cozy Client](https://trello.com/c/o3cqfjZ4/833-ajout-dun-th%C3%A8me-label-dans-cozy-client)

JF-Cozy (commented):
le "Moved `personal_sporting_licence` label" n'est pas un BC ? 🤔

Merkur39:
> le "Moved `personal_sporting_licence` label" n'est pas un BC ? 🤔

Je ne pense pas, @paultranvan saurais-tu confirmer ? 😇

paultranvan:
If an app uses it, it would be a breaking change, but I guess this is not the case yet, right? If yes, I assume there is no need for a BC

Merkur39:
> If an app uses it, it would be a breaking change, but I guess this is not the case yet, right? If yes, I assume there is no need for a BC

The change is very recent (3 weeks) here https://github.com/cozy/cozy-client/pull/1093/files
And I confirm that no app (Cozy) uses it

Crash-- (approved):

PR #1100, opened 2022-01-13
feat: Move other_revenue in financeLabels theme

Merkur39:
Moved `other_revenue` label from `work_study` theme to `finance` theme  for more pertinence

Changed `other_identity_document` translations


paultranvan (approved):

packages/cozy-client/src/models/document/documentTypeData.js
```diff
@@ -149,7 +148,8 @@ const financeLabels = [
   'bank_statement',
   'loan_agreement',
   'payment_proof_family_allowance',
-  'other_bank_document'
+  'other_bank_document',
+  'other_revenue'
 ]
 
 /**
```
JF-Cozy, line 155: Concernant la description du commit, est-ce qu'on peut ajouter un peu de contexte sur le BC stp ? On en sait pas trop de quoi il s'agit (le message va apparaître dans le changelog)

Merkur39:
@paultranvan
After researching, I don't feel that the Breaking Change is necessary.
No applications and konnectors should be strongly impacted by this change.
What do you think?

paultranvan:
> No applications and konnectors should be strongly impacted by this change.

For me, from the moment something change the behaviour of an existing API, it should be considered as a breaking change. Typically, if I have an app that expect a qualification to be in a certain theme, but is not anymore, that could break some stuff. 

But I guess that if we are confident in the actual change impact we could skip the BC to avoid having one each time we change a qualification theme? Not sure is there is a policy for such situation?

Crash--:
> For me, from the moment something change the behaviour of an existing API, it should be considered as a breaking change. Typically, if I have an app that expect a qualification to be in a certain theme, but is not anymore, that could break some stuff.

Theme and other stuff should be considered as private. Only the label is public. Third party should only use `getByLabel()` to get the right informations. 

If third party needs to rely on our theming, then they shouldn't expect that a specific label is in a specific theme. 

Maybe we should enhance that stuff to reflect that for the "Qualification" API.

Merkur39:
I removed the BC, I rather agree with @Crash-- 👍

Merkur39:
[Attributs dynamiques dans le nom des papiers](https://trello.com/c/iHF5o3bI/782-attributs-dynamiques-dans-le-nom-des-papiers)

PR #1103, opened 2022-01-14
feat: Use lib alias for safetynet

acezard:
This modifies the compileSdkVersion from 27 to 28 to avoid a compile bug when looking for sdk28 components


Crash-- (approved):
It's more a fix than a feat IMHO.

PR #1105, opened 2022-01-17
fix: Delete File

Crash--:
- We should not remove the reference before
deleting the file. cozy-stack needs to know
if the document was shared before moving it
to the Trash in order to revoke the sharing

- Use the route to remove all the referenced_by
by doing only one request instead of one by
reference.


nono:
Ping @Crash-- ;-)

cballevre (approved):

zatteo (approved):

PR #1106, opened 2022-01-20
feat: Add rootCozyUrl helper

taratatach:
To help our users connect our applications to their Cozy, we want them
to be able to simply use any URL of their Cozy copied from their
browser as their Cozy URL (e.g.
`https://claude-drive.mycozy.cloud/#/folder/io.cozy.files.root-dir`).

Without this help, we would not recognize the above URL as a valid
Cozy URL and would refuse to attempt the connection.

The new `rootCozyUrl` helper will reduce it down to the user's Cozy
URL (e.g. `https://claude.mycozy.cloud`) and return it or reject if:
- the domain does not exist
- the provided protocol is not HTTP or HTTPS
- no instance could be found with the resulting URL

Since our applications can be connected to a self-hosted Cozy, we also
handle URL ports.

### DIFF
--- a/packages/cozy-client/src/helpers.js
+++ b/packages/cozy-client/src/helpers.js
@@ -75,3 +75,163 @@ export const generateWebLink = ({
 
   return url.toString()
 }
+
+export class InvalidProtocolError extends Error {
+  constructor(url) {
+    super(`Invalid URL protocol ${url.protocol}`)
+
+    this.url = url
+  }
+}
+
+export class InvalidCozyUrlError extends Error {
+  constructor(url) {
+    super(`URL ${url.toString()} does not seem to be a valid Cozy URL`)
+
+    this.url = url
+  }
+}
+
+/* uri - Returns a well formed URL origin from a protocol, a hostname and a port
+ *
+ * If the protocol and/or port are omitted from the argument, the function will
+ * default to HTTPS and omit the port in the returned origin.
+ *
+ * @param {object} url          Object of URL elements
+ * @param {string} url.protocol Protocol to use in the origin (e.g. http)
+ * @param {string} url.hostname Hostname to use in the origin (e.g. claude.mycozy.cloud)
+ * @param {string} url.port     Port to use in the origin (e.g. 8080)
+ *
+ * @returns {string} Generated URL origin
+ */
+const uri = ({ protocol, hostname, port }) => {
+  return (
+    (protocol !== '' ? `${protocol}//` : 'https://') +
+    hostname +
+    (port !== '' ? `:${port}` : '')
+  )
+}
+
+/* wellKnownUrl - Returns a valid URL string to a Well Known password change page
+ *
+ * The built URL will point to the origin generated from the given protocol,
+ * hostname and port.
+ *
+ * @param {object} url          Object of URL elements
+ * @param {string} url.protocol Protocol to use in the origin (e.g. http)
+ * @param {string} url.hostname Hostname to use in the origin (e.g. claude.mycozy.cloud)
+ * @param {string} url.port     Port to use in the origin (e.g. 8080)
+ *
+ * @returns {string} Generated Well Known password change URL string
+ */
+const wellKnownUrl = url => uri(url) + '/.well-known/change-password'
+
+/* isValidOrigin - Checks whether a given URL is a valid Cozy origin
+ *
+ * This method tries to fetch the Well Known change password page of the Cozy
+ * supposedly at the given origin. This allows us to determine whether the given
+ * origin is the root URL of a Cozy or not via the status of the response:
+ * - a 200 response status means there's an actual Well Known password change
+ *   page accessible from the given origin so we suppose it's a valid Cozy
+ *   origin (i.e. it could be another site altogether though)
+ * - a 401 response status means the pointed page requires authentication so the
+ *   origin is probably pointing to a Cozy app
+ * - another status means there aren't any Cozy behind to the given origin
+ *
+ * @param {object} url          Object of URL elements
+ * @param {string} url.protocol Protocol to use in the origin (e.g. http)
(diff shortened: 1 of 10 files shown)


packages/cozy-client/src/helpers.spec.js
```diff
@@ -141,3 +149,128 @@ describe('generateWebLink', () => {
...
+  })
+
+  afterAll(() => {
+    disableFetchMocks()
+  })
+
+  it('should handle cozy-hosted https', async () => {
+    await expect(
+      rootCozyUrl(new URL('https://camillenimbus.mycozy.cloud'))
+    ).resolves.toEqual(new URL('https://camillenimbus.mycozy.cloud'))
```
trollepierre, line 186: let's add a const variable = new URL('https://camillenimbus.mycozy.cloud')

```suggestion
    ).resolves.toEqual(expectedRootCozyUrl)
```

taratatach: The value will be different in a number of tests so I don't see a major improvement in settings this constant.

packages/cozy-client/src/helpers.js
```diff
@@ -75,3 +75,108 @@ export const generateWebLink = ({
...
   return url.toString()
 }
+
+const uri = (
+  {
+    protocol,
+    hostname,
+    port
+  } /*: { protocol: string, hostname: string, port: string } */
+) /*: string */ => {
```
Crash--, line 85: I think this is a flow stuff, here we don't use it. Can you convert it to jsdoc please? ;)

Crash-- (commented):
nit: I don't know if this kind of helper should be in `cozy-client`. Since it makes a request to the stack it can be in `cozy-stack-client`. I'm not thinking clearly about this, so let's move on. 

I'll approve once `flow` will be converted to `jsdoc`

packages/cozy-client/src/helpers.js
```diff
@@ -75,3 +75,108 @@ export const generateWebLink = ({
...
+  url /*: { protocol: string, hostname: string, port: string } */
+) /*: Promise<boolean> */ => {
+  const { status } = await fetch(wellKnownUrl(url))
+
+  if (status === 200) {
+    return true
+  } else if (status === 404) {
+    throw new InvalidCozyUrlError(url)
+  } else {
+    return false
```
trollepierre, line 123: refactor:
```
  if (status === 404) {
    throw new InvalidCozyUrlError(url)
  } 
  return status === 200
```

packages/cozy-client/src/helpers.js
```diff
@@ -75,3 +75,108 @@ export const generateWebLink = ({
...
+
+  // If the entered URL is good, use it
+  if (await isValidOrigin(url)) {
+    return url
+  }
+
+  // If the entered URL's lowest sub-domain contains a dash, remove it and
+  // what follows and try the new resulting url.
+  if (/^[^.]+-[^.]+\./.test(url.hostname)) {
+    // XXX: url.hostname always contains a `.` as we append `.mycozy.cloud`
```
trollepierre, line 155: i see a polynomial regex error above this line. Why?

packages/cozy-client/src/helpers.js
taratatach: Ah right, sorry, old habits :sweat_smile:

packages/cozy-client/src/helpers.spec.js
```diff
@@ -141,3 +149,234 @@ describe('generateWebLink', () => {
     ).toEqual(`https://drive.alice.cozy.tools/public/#/files/432432`)
   })
 })
+
+describe.only('rootCozyUrl', () => {
```
trollepierre, line 153:
```suggestion
describe('rootCozyUrl', () => {
```

packages/cozy-client/src/helpers.js
taratatach: Ah yes. I had a test on status 401 for a little while and forgot to bring back the simple comparison.

trollepierre (approved):

packages/cozy-client/src/helpers.js
taratatach: I'm don't know what this issue is but I need to make a little tweak to the regexp. Maybe the error will go away with it.

packages/cozy-client/src/helpers.js
```diff
@@ -75,3 +75,139 @@ export const generateWebLink = ({
...
+  }
+
+  // If the entered URL is good, use it
+  if (await isValidOrigin(url)) {
+    return url
+  }
+
+  // If the entered URL's lowest sub-domain contains a dash, remove it and
+  // what follows and try the new resulting url.
+  if (/^[^.-][^.]+-[^.-]+\./.test(url.hostname)) {
```
nono, line 185: Why the regexp is not simpler? Like `/^[^.-]+-[^.-]/

packages/cozy-client/src/helpers.js
```diff
@@ -75,3 +75,139 @@ export const generateWebLink = ({
...
+    // XXX: url.hostname always contains a `.` as we append `.mycozy.cloud`
```
nono, line 186: I don't see where we append .mycozy.cloud

packages/cozy-client/src/helpers.js
taratatach: The first version looked like this but it would not match urls like `https://jean-michel-drive.mycozy.cloud` (i.e. it only allows one dash in the sub-domain).
There might be a simpler version though but as always with Regexp, it's easy to miss it. I'll take a second look and see if I can come up with a simpler one.

taratatach: I couldn't figure out a simpler version. I suppose we can always change it later.

nono (approved):

Ldoppea:
I just saw this PR and I just want to say: impressive documentation 👍

PR #1114, opened 2022-01-26
feat(file): Fill last modified date in UpdatedAt path query string

### LINKED ISSUE #690: FileUpload Date header

trollepierre:
close #690 #699 

We can now use xhr to upload file in order to get process callback. But it comes with a drawback... For security reason, Chromium based browser, don't let us set the Date header for a file. (https://bugs.chromium.org/p/chromium/issues/detail?id=38382 & https://stackoverflow.com/questions/6973403/chrome-extension-refuse-header-date-in-ajax-request/6974039 )

We use now a QueryString updated_at & created_at in order to avoid using this Date header.

It needs a modification of cozy-stack to accept this new updated_at QS.
https://github.com/cozy/cozy-stack/blob/master/web/files/files.go#L153-L163

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -950,6 +950,7 @@ class FileCollection extends DocumentCollection {
    * @param {string} method POST / PUT / PATCH
    */
   async doUpload(dataArg, path, options, method = 'POST') {
+    let correctPath = path
     let data = dataArg
     if (!data) {
       throw new Error('missing data argument')
@@ -982,8 +983,8 @@ class FileCollection extends DocumentCollection {
           // The type is specified in the file object
           contentType = data.type
         } else {
-          // Extract the name from the path and infer the type
-          const sPath = path.split('?')
+          // Extract the name from the correctPath and infer the type
+          const sPath = correctPath.split('?')
           const params = sPath.length > 1 ? sPath[1] : ''
           const name = new URLSearchParams(params).get('Name')
           contentType = this.getFileTypeFromName(name.toLowerCase())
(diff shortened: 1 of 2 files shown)


trollepierre:
[📁 DRIVE - Upload fichier en cours - bar de progression (1-2j)](https://trello.com/c/mQkBdlQ5/547-%F0%9F%93%81-drive-upload-fichier-en-cours-bar-de-progression-1-2j)

Crash--:
When I read the code's stack here https://github.com/cozy/cozy-stack/blob/master/web/files/files.go#L166-L182 I see that when we set the `date header` then the stack assign the `updated_at` and `created_at` (doc.CreatedAt = t doc.UpdatedAt = t) 

So I think than we should do the same in Cozy-Client. If we previously set the `date header` then we should set `updated_at` and `created_at` to get the same behavior.

Crash-- (approved):

Issue #1075, opened 2021-11-29
CozyClient should expose a `destroyAll`


JulienMirval:
There are `destroy`, `queryAll` and `saveAll` methods, a `destroyAll` would be nice to have

Merkur39:
For `CozyClient.destroyAll`, the `destroyAll` method of `DocumentCollection` uses its `updateAll` method, but the `io.cozy.files` are write locked.
https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/DocumentCollection.js#L391

The simplest approach would be to loop over `CozyClient.destroy` and not create a new mutation, or add a `destroyAll` method to `FileCollection` that loops over its `destroy` method (?)

PR #1115, opened 2022-01-27
feat: Add deleteFileId* functions

Merkur39:
Avoid the app to have to fetch the document if it has only its ID, to be able to delete it

### DIFF
--- a/docs/api/cozy-client/modules/models.file.md
+++ b/docs/api/cozy-client/modules/models.file.md
@@ -20,6 +20,48 @@
 
 ## Functions
 
+### deleteFileById
+
+▸ `Const` **deleteFileById**(`client`, `fileId`): `Promise`<`void`>
+
+*Parameters*
+
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `client` | [`CozyClient`](../classes/cozyclient.md) | Instance of CozyClient |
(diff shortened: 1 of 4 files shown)


Crash-- (changes requested):
I'm not very happy to have it inside the model. We'll have to add these methods in every doctypes we use. This is not futur proof. 

Why not updating at least the `CozyClient.destroy()` method to also accept only `_type` and `_id`? And make the request to get the document before calling the rest of the method? 

Also, it seems that there is a `destroyAll()` in Document Collection. Maybe we should expose it from CozyClient

Merkur39:
Indeed, I went a bit fast on my development
I had another logic in mind but it doesn't exist... :man-shrugging:

Merkur39:
For `CozyClient.destroy`, I don't think it is really relevant.
For `CozyClient.destroyAll`, the `destroyAll` method of `DocumentCollection` uses its `updateAll` method, but the `io.cozy.files` are write locked.
https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/DocumentCollection.js#L391

The simplest approach would be to loop over `CozyClient.destroy` and not create a new mutation (?)

Merkur39:
I propose to continue the discussion on this issue https://github.com/cozy/cozy-client/issues/1075

PR #1122, opened 2022-02-09
fix(documents): Reducer should assign new object for concerned doctype

trollepierre:
In Banks, we noticed the reference of the state.cozy.documents[doctype]
does not change when we refetch more transaction.

This was dued to extractAndMergeDocument that created a new
mergedData object but providing again the same properties.

Our property that got updated, from 1000 documents to 1500, was still
inside the same reference, so reselect did not recompute after store
got updated, and users faced this bug while moving from home page with
reimbursements, users see old reimbursements from home/transactions
instead of seeing the full reimbursements fetched on Reimbursements page.

when removing memoize options and adding `console.log('documents', documents ? Object.keys(documents).length : 0)` inside documentSelector line 72 => https://github.com/cozy/cozy-banks/blob/master/src/selectors/index.js#L72

Capture d’écran 2022-02-09 à 16 17 30

### DIFF
--- a/packages/cozy-client/src/store/documents.js
+++ b/packages/cozy-client/src/store/documents.js
@@ -151,16 +151,16 @@ export const getCollectionFromSlice = (state = {}, doctype) => {
 }
 
 /*
-  This method has been created in order to get a returned object 
-  in `data` with the full set on information coming potentially from 
-  ìncluded`
+  This method has been created in order to get a returned object
+  in `data` with the full set on information coming potentially from
+  `included`
 
-  This method should be somewhere else. The `document` shall not be 
-  deal with included / data and so on. 
+  This method should be somewhere else. The `document` shall not be
+  dealt with included / data and so on.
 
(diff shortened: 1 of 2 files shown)


packages/cozy-client/src/store/documents.js
```diff
@@ -172,6 +172,8 @@ export const extractAndMergeDocument = (data, updatedStateWithIncluded) => {
   const sortedData = keyBy(data, properId)
 
   let mergedData = Object.assign({}, updatedStateWithIncluded)
+  mergedData[doctype] = Object.assign({}, updatedStateWithIncluded[doctype])
+
   Object.values(sortedData).map(data => {
     if (!mergedData[doctype]) mergedData[doctype] = {}
```
nono, line 178: I think we can remove this line now

trollepierre:
ping @Crash--

Crash-- (approved):

PR #1124, opened 2022-02-14
fix(client): Explicit type of file class and type that should be string

trollepierre:
Apparently:

Capture d’écran 2022-02-14 à 16 52 21


Crash--:
I don't know how we ended up with the wrong type, but master is right ATM: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/types/types.d.ts#L190-L194

Also, `types.d.ts` is generated, you should not edit it manually.

PR #1125, opened 2022-02-16
feat: Ignore flagship certification failure to allow 2FA certification

Ldoppea:
Flagship certification heavily depends on the Google and Apple official
stores API

Not every devices are compatible with those API like for iOS old
devices or F-Droid based devices

We don't want to block those devices' users so we allow them to certify
the app using alternative 2FA verification

The 2FA verification is handled by the cozy-stack so we only need to
continue the OAuth flow instead of throwing an error

This requires cozy/cozy-stack#3287

### DIFF
--- a/packages/cozy-client/src/flagship-certification/flagship-certification.js
+++ b/packages/cozy-client/src/flagship-certification/flagship-certification.js
@@ -1,4 +1,5 @@
 import CozyClient from '../CozyClient'
+import logger from '../logger'
 
 import { getAppAttestationFromStore } from './store-attestation'
 
@@ -82,12 +83,19 @@ export const certifyFlagship = async (certificationConfig, client) => {
     )
   }
 
-  const stackChallengeNonce = await getStackChallenge(client)
+  try {
+    const stackChallengeNonce = await getStackChallenge(client)
 
-  const appAttestation = await getAppAttestationFromStore(
-    stackChallengeNonce,
-    certificationConfig
-  )
+    const appAttestation = await getAppAttestationFromStore(
+      stackChallengeNonce,
+      certificationConfig
+    )
 
-  await giveAppAttestationToStack(appAttestation, stackChallengeNonce, client)
+    await giveAppAttestationToStack(appAttestation, stackChallengeNonce, client)
+  } catch (e) {
+    logger.warn(
+      '[FLAGSHIP_CERTIFICATION] Certification failed but the cozy-stack will continue with 2FA certification'
+    )
+    logger.warn(e.message)
+  }
 }
--- a/packages/cozy-client/src/flagship-certification/flagship-certification.spec.js
+++ b/packages/cozy-client/src/flagship-certification/flagship-certification.spec.js
@@ -1,0 +1,115 @@
+import { certifyFlagship } from './flagship-certification'
+import { createMockClient } from '../mock'
+
+import { getAppAttestationFromStore } from './store-attestation'
+
+jest.mock('./store-attestation', () => ({
+  getAppAttestationFromStore: jest.fn()
+}))
+
+const getClientMock = () => {
+  const client = createMockClient({
+    clientOptions: {
+      oauth: {}
+    }
+  })
+
+  client.stackClient.oauthOptions = {
+    clientID: 'SOME_CLIENT_ID',
+    registrationAccessToken: 'SOME_REGISTRATION_ACCESS_TOKEN'
+  }
+
+  return client
+}
+
+const mockCorrectChallengeRequest = client => {
+  client.stackClient.fetchJSON.mockResolvedValueOnce({
+    nonce: 'SOME_NONCE'
+  })
+}
+
+const mockCorrectStoreApiRequest = () => {
+  getAppAttestationFromStore.mockResolvedValueOnce({
+    platform: 'android',
+    attestation: 'SOME_STORE_ATTESTATION',
+    keyId: 'SOME_KEY_ID'
+  })
+}
+
+const mockCorrectCertificationConfig = () => {
+  return {
+    androidSafetyNetApiKey: 'SOME_ANDROID_SAFETY_NET_API_KEY'
+  }
+}
+
+const getFetchJsonPostParams = (url, body = null) => {
+  return [
+    'POST',
+    url,
+    body,
+    {
+      headers: {
+        Authorization: 'Bearer SOME_REGISTRATION_ACCESS_TOKEN'
+      }
+    }
+  ]
+}
+
+describe('certifyFlagship', () => {
+  afterEach(() => {
+    jest.restoreAllMocks()
+    getAppAttestationFromStore.mockReset()
+  })
+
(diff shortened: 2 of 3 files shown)


packages/cozy-client/src/flagship-certification/flagship-certification.js
```diff
@@ -82,12 +82,19 @@ export const certifyFlagship = async (certificationConfig, client) => {
...
+      certificationConfig
+    )
 
-  await giveAppAttestationToStack(appAttestation, stackChallengeNonce, client)
+    await giveAppAttestationToStack(appAttestation, stackChallengeNonce, client)
+  } catch (e) {
+    console.warn(
+      '[FLAGSHIP_CERTIFICATION] Certification failed but the cozy-stack should continue with 2FA certification'
+    )
+    console.warn(e.message)
```
trollepierre, line 98: should we use cozy-logger here 🙏 ?

packages/cozy-client/src/flagship-certification/flagship-certification.js
```diff
@@ -82,12 +82,19 @@ export const certifyFlagship = async (certificationConfig, client) => {
...
+    const appAttestation = await getAppAttestationFromStore(
+      stackChallengeNonce,
+      certificationConfig
+    )
 
-  await giveAppAttestationToStack(appAttestation, stackChallengeNonce, client)
+    await giveAppAttestationToStack(appAttestation, stackChallengeNonce, client)
+  } catch (e) {
+    console.warn(
+      '[FLAGSHIP_CERTIFICATION] Certification failed but the cozy-stack should continue with 2FA certification'
```
trollepierre, line 96: this sentence is not easy to understand. 

Negative "failed"  "but" "should"...

packages/cozy-client/src/flagship-certification/flagship-certification.spec.js
```diff
@@ -0,0 +1,235 @@
...
+
+    // Should ask the stack for a challenge
+    expect(client.stackClient.fetchJSON).toHaveBeenCalledWith(
+      'POST',
+      `/auth/clients/SOME_CLIENT_ID/challenge`,
+      null,
+      {
+        headers: {
+          Authorization: 'Bearer SOME_REGISTRATION_ACCESS_TOKEN'
+        }
```
trollepierre, line 74: the code scanning error can be dismissed (used in tests)

packages/cozy-client/src/flagship-certification/flagship-certification.spec.js
```diff
@@ -0,0 +1,235 @@
...
+
+    mockCorrectChallengeRequest(client)
+
+    mockCorrectStoreApiRequest()
+
+    await certifyFlagship(certificationConfig, client)
+
+    // Should ask the stack for a challenge
+    expect(client.stackClient.fetchJSON).toHaveBeenCalledWith(
+      'POST',
```
trollepierre, line 68: let's  refactor

expect(client.stackClient.fetchJSON).toHaveBeenCalledWith(fetchJSONExpectedParams(url))

in all the following tests 🙏

packages/cozy-client/src/flagship-certification/flagship-certification.spec.js
```diff
@@ -0,0 +1,235 @@
...
+
+    const certificationConfig = mockCorrectCertificationConfig()
+
+    mockCorrectChallengeRequest(client)
+
+    mockCorrectStoreApiRequest()
+
+    await certifyFlagship(certificationConfig, client)
+
+    // Should ask the stack for a challenge
```
trollepierre, line 66: instead of using comment in  the test, let's split the test =>
```
it('should ask the stack for a challenge', () => {
// this expect below only
```

trollepierre: same for all other comments

trollepierre (commented):
if you want, we can pair prog to take into consideration all my discussions in a row faster 🙏

packages/cozy-client/src/flagship-certification/flagship-certification.spec.js
Ldoppea: I'm not sure, I want to test this specific call combination as the certification process require to ask a challenge to the stack, then query the Store api, then give the result to the stack. Multiple test cases would not allow to test the sequential order, would it?

packages/cozy-client/src/flagship-certification/flagship-certification.js
Ldoppea: I don't remember, did we talked with @Crash-- about cozy-logger usage outside of konnectors?

It seems not to be used in cozy-client (expect for the CLI part) and using console.warn seems to be the common practice. Do we wan't to start using cozy-logger instead?

packages/cozy-client/src/flagship-certification/flagship-certification.spec.js
Ldoppea: What about this? https://github.com/cozy/cozy-client/compare/f99251f97be750c0363bcd659d1628c5937602de..d9ea74471e1e23ed356a9e710722a2939f4b6972

packages/cozy-client/src/flagship-certification/flagship-certification.js
trollepierre: I have no history on this.

Fow now, let's keep console warn and let's discuss about this with @Crash-- !

packages/cozy-client/src/flagship-certification/flagship-certification.spec.js
trollepierre: ok, then, instead of using comment I suggest to rename the test:

'it should ask the stack for a challenge, before doing B action, then doing C, and finally D'

trollepierre (approved):
🎉 Nice

packages/cozy-client/src/flagship-certification/flagship-certification.js
Crash--: https://github.com/cozy/cozy-client/blob/master/docs/logging.md

Crash--: So...  :  
```
import logger from './logger'
 logger.warn(``)
```

packages/cozy-client/src/flagship-certification/flagship-certification.js
Ldoppea: I made the edit here: https://github.com/cozy/cozy-client/compare/d9ea74471e1e23ed356a9e710722a2939f4b6972..559190a18cb89ec0392b90d02f58b85f577e03e7

packages/cozy-client/src/flagship-certification/flagship-certification.js
Ldoppea: I made the edit here: https://github.com/cozy/cozy-client/compare/d9ea74471e1e23ed356a9e710722a2939f4b6972..559190a18cb89ec0392b90d02f58b85f577e03e7

PR #1130, opened 2022-02-21
feat: Add CoachCO2 to MAGIC_FOLDERS

Merkur39:
To export the data always in the same folder, we need to add Coach_CO2 to Magic_Folders


Crash-- (approved):
fix the doc 🙏 . After ok to merge directly.

PR #1131, opened 2022-02-21
Feat/handle www authenticate header error

Ldoppea:
In cozy/cozy-stack@2d1e27652c76fb9cfb66751019f40ee9a1ae64d1 we
introduced `WWW-Authenticate` header usage to notify that the token
expired

CozyStackClient is now able to refresh its token when
`WWW-Authenticate` is set with expired token error

### DIFF
--- a/packages/cozy-stack-client/src/errors.js
+++ b/packages/cozy-stack-client/src/errors.js
@@ -8,6 +8,27 @@ export default {
   INVALID_TOKEN
 }
 
+const getWwwAuthenticateErrorMessage = response => {
+  const invalidTokenRegex = /invalid_token/
+  const expiredTokenRegex = /access token expired/
+  const wwwAuthenticateHeader =
+    response.headers && response.headers.get('www-authenticate')
+
+  if (!wwwAuthenticateHeader) {
+    return undefined
+  }
+
+  if (expiredTokenRegex.test(wwwAuthenticateHeader)) {
+    return 'Expired token'
+  }
+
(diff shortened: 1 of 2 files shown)


Ldoppea:
Note that https://github.com/cozy/cozy-stack/commit/2d1e27652c76fb9cfb66751019f40ee9a1ae64d1 added multiple case where the `Www-authenticate` header may contain only `invalid_token` error with no `expired` description.

My code only handle errors with `expired` description. Should I react to all `invalid_token` errors instead?

packages/cozy-stack-client/src/errors.js
```diff
@@ -8,6 +8,19 @@ export default {
   INVALID_TOKEN
 }
 
+const hasWwwAuthenticateExpiredTokenError = response => {
+  const invalidTokenRegex = /invalid_token/
```
Crash--, line 12: We should not mask this error and should not transform it in `expired token`. 

If you don't have the time to handle it correctly, I prefer that we do not transform it.

Ldoppea: I force-pushed a new version that returns `Expired token` only when the description contains the expired token description, and `Invalid token` when this description is not about expiration or if unset : https://github.com/cozy/cozy-client/compare/220a641ca974f0eef33a6f22826251071bdf9c63..3c7d3aea066bdfd2c93b1a731385f8fc52fed615

Crash-- (approved):

PR #1133, opened 2022-02-22
feat: Add Manual to Job creation

Crash--:
We had the capacity to launch trigger `manually`. But there was no way to create a `manual` job. 

Since https://github.com/cozy/cozy-stack/pull/3308 we can create the job with the `manual` argument. 

What is this `manual` stuff? By default, jobs are queued and the gozy-worker enqueue the jobs progressively. But sometimes, we need to run a job as quick as possible since the user is currently waiting for the response of the job. To achieve that, we have this concept of `manual` job. The job is still async, but not in the `global` queue. 

This is needed when we want to launch the banking connector to get a temporary token in order to create the account on the BI's side.

### DIFF
--- a/packages/cozy-stack-client/src/JobCollection.js
+++ b/packages/cozy-stack-client/src/JobCollection.js
@@ -39,17 +39,19 @@ class JobCollection {
    * Creates a job
    *
    * @param  {string} workerType - Ex: "konnector"
-   * @param  {object} args - Ex: {"slug": "my-konnector", "trigger": "trigger-id"}
-   * @param  {object} options - creation options
+   * @param  {object} [args={}] - Ex: {"slug": "my-konnector", "trigger": "trigger-id"}
+   * @param  {object} [options={}] - creation options
+   * @param  {boolean} [manual=false] -  Manual execution
    * @returns {object} createdJob
    */
-  create(workerType, args, options) {
+  create(workerType, args = {}, options = {}, manual = false) {
     return this.stackClient.fetchJSON('POST', `/jobs/queue/${workerType}`, {
       data: {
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/JobCollection.js
```diff
@@ -41,15 +41,17 @@ class JobCollection {
    * @param  {string} workerType - Ex: "konnector"
    * @param  {object} args - Ex: {"slug": "my-konnector", "trigger": "trigger-id"}
    * @param  {object} options - creation options
+   * @param  {boolean} manual -  Manual execution
```
trollepierre, line 44:
```suggestion
   * @param  {boolean} [manual=false] -  Manual execution
```
According to https://jsdoc.app/tags-param.html

trollepierre: actually, I think we should uniformize. Either use default param  like you did (I prefer that), or use like line 53, the `or` operator.

packages/cozy-stack-client/src/JobCollection.js
```diff
@@ -41,15 +41,17 @@ class JobCollection {
...
    */
-  create(workerType, args, options) {
+  create(workerType, args, options, manual = false) {
     return this.stackClient.fetchJSON('POST', `/jobs/queue/${workerType}`, {
       data: {
         type: JOBS_DOCTYPE,
         attributes: {
           arguments: args || {},
-          options: options || {}
+          options: options || {},
```
trollepierre, line 53: i think options and args can also have a optional value passed in the jsdocs

trollepierre (approved):

PR #1139, opened 2022-03-01
fix(deps): Upgrade cozy-flags + consider cozy lib as peerDependencies

trollepierre:
after this commit: https://github.com/cozy/cozy-libs/commit/96ea387d4db7360f283e8af7c6d5a4a3c94640a7

BREAKING CHANGE: Each consuming app should install the following
dependencies:
    "@cozy/minilog": "1.0.0",
    "cozy-device-helper": ">1.12.0",
    "cozy-flags": ">2.8.6",
    "cozy-logger": ">1.7.0",


trollepierre:
cc @Crash-- 🙏

Crash--:
cozy-flag should not be a dep of cozy-client. It should be a peerDep / devDep. 

Our rule is: no dep between cozy librairies.

trollepierre:
@Crash-- : so should I remove also

    "cozy-device-helper": "^1.12.0",
    "cozy-flags": "2.8.6",
    "cozy-logger": "^1.6.0",
    "cozy-stack-client": "^27.19.0",
    
    from dependencies, and consider them as peerDependencies?

Crash--:
For stack-client I don't have a clear mind on it. Currently we assume that installing cozy-client installs cozy-stack-client. We don't say anywhere we should install stack-client too. So I'm not in favor of removing it for now. 

But for others yes. Maybe the commit should be tagged as breaking change.

Crash-- (approved):

trollepierre:
@Crash-- : if this is a breaking change, should I wait for merging this?

Crash--:
You can merge it. This kind of BC where we just have to install other libs are not blocking

PR #1150, opened 2022-03-16
fix(cozy-client): Remove unit tests, mocks, snapshots, testing from dist

trollepierre:
- `--copy-files` was not ignoring specified ignored files
- add more files to remove from final dist

After:
Capture d’écran 2022-03-16 à 18 19 38

Before:
Capture d’écran 2022-03-16 à 18 19 38


Crash--:
Waiting for green CI before reviewing :)

PR #1157, opened 2022-03-24
chore(dependabot): Force dependabot PR to run only once a week

trollepierre:


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
```
Crash--, line 3: since we use yarn, why not using yarn here?

trollepierre: No. https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem=

Crash--: This is exclatly my point, yarn exsists Capture d’écran 2022-03-28 à 11 42 43 https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

trollepierre: yaml value is "npm" 😅

Crash-- (approved):

PR #1162, opened 2022-04-20
feat: Improve createMockClient

Merkur39:
Added a property so that this function can be used in mocks other than for testing (cozy-ui documentation)

Example of a need: https://github.com/cozy/cozy-ui/blob/master/react/FilePicker/docs/DemoProvider.jsx#L41


packages/cozy-client/src/mock.js
```diff
@@ -27,32 +36,44 @@ const mockedQueryFromMockedRemoteData = remoteData => qdef => {
...
+ * @param {boolean} [options.isDocumentation] Define if used for documentation or test
  * @returns {CozyClient}
  */
-const createMockClient = ({ queries, remote, clientOptions }) => {
+const createMockClient = ({
+  queries,
+  remote,
+  clientOptions,
+  isDocumentation = false
+}) => {
```
Crash--, line 60: I dislike this kind of API. We're creating a client, with all arguments related to the client. I don't think that adding a `isDocumentation` there is not a good idea. 

What do you think of instead of adding a new boolean attribute, we add a new object containing the method we want to mock? 

Something like: createMockClient({queries, remote, clientOptions, mockedMethod: ... ?

Also, why in UI you need to fillQueryInsideClient but you don't want to mock client.query() to read from there?

Crash--: ping @Merkur39

cballevre:
I recently add `createFakeClient` which is same as `createMockClient` but without jest to help fake queries inside Storybook maybe it solved your issue ([PR](https://github.com/cozy/cozy-client/pull/1431)) 🤔

PR #1165, opened 2022-04-22
feat: Add getSharingLink function

Merkur39:
Added `getSharingLink` function to generate a sharing link for one or more files.
This function already exists in different libraries like `cozy-sharing` or `cozy-mespapiers-lib`, and soon in `cozy-ui`.
So I think it's good to centralize it in `cozy-client`.

### DIFF
--- a/docs/api/cozy-client/modules/models.md
+++ b/docs/api/cozy-client/modules/models.md
@@ -13,6 +13,7 @@
 *   [instance](models.instance.md)
 *   [note](models.note.md)
 *   [permission](models.permission.md)
+*   [sharing](models.sharing.md)
 *   [timeseries](models.timeseries.md)
 *   [trigger](models.trigger.md)
 *   [utils](models.utils.md)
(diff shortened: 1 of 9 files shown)


packages/cozy-client/src/helpers.js
```diff
@@ -76,6 +78,34 @@ export const generateWebLink = ({
...
+    cozyUrl: client.getStackClient().uri,
+    searchParams: [['sharecode', sharedLink?.attributes?.shortcodes?.code]],
+    pathname: '/public',
+    slug: 'drive',
+    subDomainType: isFlatDomain ? 'flat' : 'nested'
+  })
+
+  return webLink
+}
+
```
Crash--, line 108: can you move it to a new file called sharing.js(https://github.com/cozy/cozy-client/tree/master/packages/cozy-client/src/models) ?

Merkur39: Ok, that's better indeed!
It's done

Crash--: can you remove this commit ? :p

Merkur39: :face_palm:

Crash-- (approved):

PR #1166, opened 2022-04-22
Fix/ci typecheck

Merkur39:
For the good functioning, we need to have "declaration" set to "true" if "emitDeclarationOnly" is "true"
Since https://github.com/cozy/cozy-client/commit/09889e5d60b0da5ab1a4c44db5ac5d4c02d007de


trollepierre (approved):
👌

PR #1172, opened 2022-05-04
feat: Add secure protocol to URI when window.cozy.isSecureProtocol=true

Ldoppea:
In our ReactNative app we will need to display cozy-apps from a webview
served from a local HTTP Server

By doing so, the webview's base URL is using an unsecure `http`
protocol

As cozy-client's URI is often created using `window.location.protocol`
then it would register itself to the cozy-stack using the same unsecure
`http` protocol from the webview's base URL

With current code, cozy-client has no clue if it should trust
`window.location.protocol` or not

As this info is known by the ReactNative app, we should to inject the
information into the webview by using the
`window.cozy.isSecureProtocol` global object

So when `new CozyClient()` is called, if `window.cozy.isSecureProtocol`
is set to true, then cozy-client now transforms the URI's protocol into
a secure `https` one

### DIFF
--- a/docs/api/cozy-client/classes/CozyClient.md
+++ b/docs/api/cozy-client/classes/CozyClient.md
@@ -43,7 +43,7 @@ Cozy-Client will automatically call `this.login()` if provided with a token and 
 
 *Defined in*
 
-[packages/cozy-client/src/CozyClient.js:150](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L150)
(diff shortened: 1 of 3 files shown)


Ldoppea:
⚠️ Please note that the conversion to `https` adds a trailing `/` to the url string. After doing some test in the ReactNative app this seems to have no side effect. But if you think this is problematic I can try to remove it.

trollepierre (approved):

Crash-- (approved):

PR #1174, opened 2022-05-10
fix: Cozy-Client runs on node

Crash--:
BREAKING CHANGE: We need to upgrade cozy-device-helper
in order to be able to run cozy-client within a node
context. So let's upgrade it.

You need to install cozy-device-helper >= 2.1.0 in
your app.


trollepierre (approved):
all good

PR #1176, opened 2022-05-19
fix: CozyClient can be used in a node env

Crash--:
We should not rely on `window` since CozyClient can be used in a Node env. 

This commit https://github.com/cozy/cozy-client/commit/3a9d557bbf70f6c7a30acbdad225000b40a23215 introduced a regression for our app's services.


Ldoppea (approved):

paultranvan (approved):
Thanks! :zap:

Issue #1179, opened 2022-05-25
Current tests are not covering Node environment


Ldoppea:
Current Jest configuration is using `testEnvironment: 'jest-environment-jsdom-sixteen'` which means that all tests are run in a valid DOM environment.

This is problematic as CozyClient can be used in a Node.js environment (i.e. connectors' jobs)

A recent consequence from this configuration is that we are not able to correctly test the `securiseUri` method that is based on `window` global object. This method should not throw in Node.js environment which doesn't declare the `window` global object.
Because tests are run using `jest-environment-jsdom-sixteen`, all tests were success in #1176
But the method was still failing in Node.js environment. So we fixed that in #1180 but without ability to add a unit test to validate the change.

A good way to resolve this would be to run tests 2 times. One time with the `jest-environment-jsdom-sixteen` environment, and another time with the `node` environment.
But here a lot of existing tests would fail as some of them are not supposed to be run in `node` environment.
We should find a way to discriminate `node` only tests and `DOM` only tests.

paultranvan:
The node [examples](https://github.com/cozy/cozy-client/tree/master/packages/cozy-client/examples) might be enough to guarantee the node execution is not broken. It won't cover everything but at least check the init/query methods

PR #1180, opened 2022-05-25
fix: CozyClient can be used in a node env

Ldoppea:
Previous fix tried to handle undefined `window` in node env

But in a node env window is not `undefined` as it is not even an
existing variable

So instead of checking its value, a better way is to check its type as
`typeof` will return `'undefined'` when called with a variable that
does not exist

Related commit: 509a0ca8daa630e300daecae23a14ebd83b745df

> **Warning**
> This fix cannot be unit tested until [#1179 - Current tests are not covering Node environment](https://github.com/cozy/cozy-client/issues/1179) is resolved

Following test is success with the new fix if run from a `node` env
```js
    it('should not secure the URL if window is not existing (node env)', () => {
      const client = new CozyClient({
        uri: 'http://cozy.tools',
        schema: '',
        token: 'SOME_TOKEN'
      })

      expect(client.options.uri).toBe('http://cozy.tools')
    })
```

Following test is success with the new fix if run from a `jsdom` env
```js
    it('should not secure the URL if window is undefined (node env)', () => {
      let windowSpy = jest.spyOn(window, 'window', 'get')
      windowSpy.mockImplementation(() => undefined)
      const client = new CozyClient({
        uri: 'http://cozy.tools',
        schema: '',
        token: 'SOME_TOKEN'
      })

      expect(client.options.uri).toBe('http://cozy.tools')
      windowSpy.mockRestore()
    })
```

### DIFF
--- a/docs/api/cozy-client/classes/CozyClient.md
+++ b/docs/api/cozy-client/classes/CozyClient.md
@@ -43,7 +43,7 @@ Cozy-Client will automatically call `this.login()` if provided with a token and 
 
 *Defined in*
 
-[packages/cozy-client/src/CozyClient.js:161](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L161)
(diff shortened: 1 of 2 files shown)


paultranvan (approved):
Thanks :tada:

PR #1188, opened 2022-06-09
fix: Enforce limit on documents coming from store

paultranvan:
BREAKING CHANGE: we used to return all data from the store
positively evaluated by sift without enforcing any limit, even when a
`.limitBy` was specified.

To demonstrate the problematic, consider this example:

```
Q1 = Q('io.cozy.todos').select(['name']).limitBy(10)
Q2 = Q('io.cozy.todos').limitBy(1000)
```

When Q2 is run, Q1 is updated and receives all data coming from Q2.
Thus, Q1 returns 1000 docs, while expecting only 10 max.

Another problem was related to the use of `.select`: Q1 only needs
the `name` attribute, and displays todos accordingly to this expected
format. But since it actually also receives the results from Q2, some
todos will have more data than expected.
Likewise, Q2 will expect all the attributes from Q2. But, if the todos
retreived by Q1 and Q2 are not the same, Q2 will have todos without
all the expecting fields, which could leads to issues.


packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -1338,6 +1338,44 @@ describe('CozyClient', () => {
...
+        fetchStatus: 'loaded',
+        definition: query.toDefinition(),
+        totalQueryFetchCount: 1,
+        data: allDocs
+      })
+      await client.query(query, { as: 'one-todo' })
+      const dispatchCalls = client.store.dispatch.mock.calls
+      const lastDispatchCall = dispatchCalls[dispatchCalls.length - 1]
+      expect(lastDispatchCall[0]).toEqual(
+        receiveQueryResult('one-todo', allDocs)
```
trollepierre, line 1375: where do we test the limit / no-limitation in these 2 tests?

packages/cozy-client/src/store/queries.js
```diff
@@ -54,23 +54,56 @@ const queryInitialState = {
...
 }
 
-const updateQueryDataFromResponse = (queryState, response, documents) => {
-  let updatedIds = uniq([...queryState.data, ...response.data.map(properId)])
-  if (queryState.definition.sort) {
+/**
+ * Return the docs ids accordingly to the given sort and fetched docs
+ *
+ * @param {QueryState} queryState - Current state
+ * @param {DocumentsStateSlice} documents - Referencec to the documents slice
```
trollepierre, line 66:
```suggestion
 * @param {DocumentsStateSlice} documents - Reference to the documents slice
```

packages/cozy-client/src/store.spec.js
```diff
@@ -230,6 +230,54 @@ describe('Store', () => {
...
+      )
+      store.dispatch(
+        receiveQueryResult('Q2', {
+          data: [TODO_4, TODO_2]
+        })
+      )
+      queryResult = getQueryFromStore(store, 'Q1')
+      expect(queryResult.data.map(x => x._id)).toEqual(['todo_3', 'todo_1'])
+      queryResult = getQueryFromStore(store, 'Q2')
+      expect(queryResult.data.map(x => x._id)).toEqual(['todo_4', 'todo_2'])
```
trollepierre, line 278: I struggle to understand those 2 tests above. Why changing 
```
store.dispatch(
        receiveQueryResult('Q1', {
          data: [TODO_1, TODO_2]
        })
      )
      store.dispatch(
        receiveQueryResult('Q2', {
          data: [TODO_3, TODO_4]
        })
      )
```
      
      by 
      
```
      store.dispatch(
        receiveQueryResult('Q1', {
          data: [TODO_3, TODO_1]
        })
      )
      store.dispatch(
        receiveQueryResult('Q2', {
          data: [TODO_4, TODO_2]
        })
      )
      ```
      would test the sorting method in the expectation?

File: packages/cozy-client/src/store.spec.js [hunk 4]
```diff
@@ -230,6 +230,54 @@ describe('Store', () => {
...
+      )
+      store.dispatch(
+        receiveQueryResult('Q2', {
+          data: [TODO_3, TODO_4]
+        })
+      )
+      queryResult = getQueryFromStore(store, 'Q1')
+      expect(queryResult.data.map(x => x._id)).toEqual(['todo_1', 'todo_2'])
+      queryResult = getQueryFromStore(store, 'Q2')
+      expect(queryResult.data.map(x => x._id)).toEqual(['todo_3', 'todo_4'])
```
Crash--, line 254: Q1 and Q2 are the same query `query` right? So I don't understand why the result from the store is not the same.

Maybe we should just have a qdef for Q1 (limit: 2) and an other QDef for Q2 (skip: 2, limit: 2) to be more understandable? 

By the way, why did you use two queries? Why not just one: Q1 limit 2, check the queryResult.length then dispatching new todos (via dispatchMutationResult) and check the result?

packages/cozy-client/src/store.spec.js
```diff
@@ -230,6 +230,54 @@ describe('Store', () => {
...
+          data: [TODO_3, TODO_1]
+        })
+      )
+      store.dispatch(
+        receiveQueryResult('Q2', {
+          data: [TODO_4, TODO_2]
+        })
+      )
+      queryResult = getQueryFromStore(store, 'Q1')
+      expect(queryResult.data.map(x => x._id)).toEqual(['todo_3', 'todo_1'])
```
Crash--, line 276:
```suggestion
      expect(queryResult.data.map(x => {_id: x._id, label: x.label)).toEqual(['{_id: todo_3', 'label' : a}, ...)
```

packages/cozy-client/src/store/queries.js
```diff
@@ -54,23 +54,56 @@ const queryInitialState = {
...
+  if (queryState.definition.limit) {
+    evaluatedIds = queryState.definition.limit
+      ? evaluatedIds.slice(0, count * fetchedPagesCount)
+      : evaluatedIds
   }
-  return updatedIds
+  return evaluatedIds
+}
+
+const updateQueryDataFromResponse = (
```
Crash--, line 96: come on! You have the same types just above :p

Crash-- (approved):
Nice work!

packages/cozy-client/src/store.spec.js
paultranvan: Hopefully, [2a31fff](https://github.com/cozy/cozy-client/pull/1188/commits/2a31fff42ada54d4d713f53c328095d862118c22) will make it clearer

packages/cozy-client/src/store.spec.js
paultranvan: Right, [2a31fff](https://github.com/cozy/cozy-client/pull/1188/commits/2a31fff42ada54d4d713f53c328095d862118c22) should improve this

Crash-- (approved):

Issue #1054, opened 2021-10-25
IndexName should use partialIndex/Filter 


Crash--:
Currently the indexName is based on the sort() fields or indexedFileds. 

But if we add a partial selector in the request, we'll use the same index. We need to include the `partialFilter` in the index name. 

(see https://mattermost.cozycloud.cc/cozycloud/pl/6jk58iifziftmfq166shbogprw for more informations)

Issue #1091, opened 2021-12-17
_design doc ends up with 2 index definitions


Merkur39:
I have a file "myFile.md" in my Drive (anywhere outside the trash) and the same in the trash.

This request returns both files:
```js
const queryDef = Q(FILES_DOCTYPE)
      .where({ name: "myFile.md" })
      .partialIndex({
        type: 'file',
        trashed: false
      })
      .indexFields(['name'])
```
This request generates this payload:
```json
{
  "selector": {
    "name": "myFile.md"
  },
  "use_index": "_design/by_name",
  "skip": 0
}
```
The problem seems to come from the `use_index`, in db:
```json
{
  "_id": "_design/by_name",
  "_rev": "2-c2c21817c4435fe6316c1038976fb5ce",
  "language": "query",
  "views": {
    "358cf1970125c4d7130143c33fb50472e058acd0": {
      "map": {
        "fields": {
          "name": "asc"
        },
        "partial_filter_selector": {}
      },
      "reduce": "_count",
      "options": {
        "def": {
          "fields": [
            "name"
          ]
        }
      }
    },
    "b046488cf91b15ec368567adcce573798d530f81": {
      "map": {
        "fields": {
          "name": "asc"
        },
        "partial_filter_selector": {
          "trashed": {
            "$eq": false
          }
        }
      },
      "reduce": "_count",
      "options": {
        "def": {
          "fields": [
            "name"
          ],
          "partial_filter_selector": {
            "trashed": false
          }
        }
      }
    }
  }
}
```
There are 2 `views` objects, one with `"partial_filter_selector": {}`, which would explain this bug 🤔

paultranvan:
It seems the index used for your query is the first one defined in the _design doc, thus the one without partial index, which explain why you retrieve the trashed file.
I have to investigate how we can end up with such design index. My guess is that a `_design/by_name` was already existing and it was merged by couchdb when trying to create the one with the partial index. 
So it could be related to https://github.com/cozy/cozy-client/issues/1054

Crash--:
It's clearly related to #1054 👍

PR #1191, opened 2022-06-23
feat: Add the partial filter fields in the index name

paultranvan:
When using a partial index, its fields weren't part of the index name.
Therefore, we might have a conflict situation where there is already an
existing index without the partial index.
Fix https://github.com/cozy/cozy-client/issues/1054 and https://github.com/cozy/cozy-client/issues/1091

### DIFF
--- a/packages/cozy-stack-client/src/mangoIndex.js
+++ b/packages/cozy-stack-client/src/mangoIndex.js
@@ -12,6 +12,7 @@ import isEqual from 'lodash/isEqual'
  *
  * @property {Array<object>} [sort] The sorting parameters
  * @property {Array<string>} [fields] The fields to return
+ * @property {Array<string>} [partialFilterFields] The partial filter fields
  * @property {number|null} [limit] For pagination, the number of results to return
  * @property {number|null} [skip] For skip-based pagination, the number of referenced files to skip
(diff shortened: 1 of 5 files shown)


Crash-- (approved):
Looks good 👍 

what about pouch? I know that we convert partial filter to selector but what about the index name?

packages/cozy-stack-client/src/mangoIndex.js
```diff
@@ -61,11 +78,12 @@ export const transformSort = sort => {
  *
  * @returns {Array} - Fields to index
  */
-export const getIndexFields = ({ selector, sort = [] }) => {
+export const getIndexFields = ({ selector, partialFilter, sort = [] }) => {
```
trollepierre, line 81: could we document the options.partialFilter in the JSDocs, please?

trollepierre (approved):

paultranvan:
> what about pouch? I know that we convert partial filter to selector but what about the index name?

The method to build the index name is duplicated in pouch-link, yet without the partial filter: https://github.com/cozy/cozy-client/blob/d289d3676de1de89d0ff38d659c188319e73c16f/packages/cozy-pouch-link/src/mango.js#L27
However, the partial filter is indeed merged with the selector, so I'm not sure it's relevant to include it

PR #1192, opened 2022-06-24
fix: Make a correct deep merge on objects.

### LINKED ISSUE #1175: Conflicted mutation in store with queries including ".select()"
In the following example :
```js
const defaultFetchPolicy = fetchPolicies.olderThan(30 * 1000)

const A = () => {
  return {
    definition: () =>
      Q(FILES_DOCTYPE)
        .partialIndex({
          type: 'file',
          trashed: false
        })
        .select(['name', 'metadata.qualification.label'])
        .limitBy(1000),
    options: {
      as: `${FILES_DOCTYPE}}/labels`,
      fetchPolicy: defaultFetchPolicy
    }
  }
}

const B = label => ({
  definition: Q(FILES_DOCTYPE)
    .where({
      'metadata.qualification': {
        label: label
      }
    })
    .partialIndex({
      type: 'file',
      trashed: false
    })
    .indexFields(['created_at', 'metadata.qualification'])
    .sortBy([{ created_at: 'desc' }]),
  options: {
    as: `${FILES_DOCTYPE}/${label}`,
    fetchPolicy: defaultFetchPolicy
  }
})
```

Let's assume the following queries execution: 

1. B -> query database, store data in store
2. A -> query database, store data in store
3. B -> query store thanks to fetchPolicy

After step 1, the query `B` puts in the store all the retrieved `io.cozy.files` with with all their fields.
But after 2., the documents retrieved through `B` are updated in the store to match the query `A` that restricts the fields thanks to the `select`. In the store, the documents now look like this:

```js
// io.cozy.files
{ 
  ...
  metadata { qualification: { label: "invoices" } }
  ...
}
```

Therefore, when step 3 occurs, with a new `B` executed under the fetchPolicy, the store is queried and the files retrieved. But only the fields from `A` are there, so the files are not complete...

Crash--:
When we merge object with the spread operator, it
doesn't make a deep merge, it replace the second
level.

Exemple:
```js
d1 = {a: 1, b: {b1: 1}}
d2 =  {c: 3, b: {b2: 2}}
merge = {...d1, ...d2}
// merge = { a: 1, b: { b2: 2 }, c: 3 }
```

This is not what we want in this case. We want a
fully merged object, not a replace. We expect the
result to be:
```
// mege = { a : 1, b { b1: 1, b2: 2}, c: 3}
```

To achieve that, we use lodash merge!

Fix #1175

(credits to @paultranvan for the finding!)

### DIFF
--- a/packages/cozy-client/src/store/documents.spec.js
+++ b/packages/cozy-client/src/store/documents.spec.js
@@ -10,7 +10,10 @@ describe('extractAndMerge', () => {
       _id: 'b6ff135b34e041ffb2d4a4865f3e0a53',
       type: 'io.cozy.files',
       _type: 'io.cozy.files',
-      blabla: 'new field'
+      blabla: 'new field',
+      cozyMedata: {
+        created_at: '123456'
+      }
     },
     1: {
       id: 'b6ff135b34e041ffb2d4a4865f3e235f',
@@ -38,7 +41,10 @@ describe('extractAndMerge', () => {
           }
         },
         type: 'io.cozy.files',
-        _type: 'io.cozy.files'
+        _type: 'io.cozy.files',
+        cozyMedata: {
+          updated_at: '987654'
+        }
       },
       b6ff135b34e041ffb2d4a4865f3e235f: {
         attributes: {
@@ -78,6 +84,10 @@ describe('extractAndMerge', () => {
             type: 'file'
           },
           blabla: 'new field',
+          cozyMedata: {
+            created_at: '123456',
+            updated_at: '987654'
+          },
           id: 'b6ff135b34e041ffb2d4a4865f3e0a53',
           links: { self: '/files/b6ff135b34e041ffb2d4a4865f3e0a53' },
           meta: { rev: '5-87840eceaab358e38aa8b6bb0d4577b1' },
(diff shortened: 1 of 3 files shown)


Merkur39 (approved):
Tested and approved 🎉

packages/cozy-client/src/store/documents.spec.js
```diff
@@ -78,6 +84,10 @@ describe('extractAndMerge', () => {
             type: 'file'
           },
           blabla: 'new field',
+          cozyMedata: {
+            created_at: '123456',
+            updated_at: '987654'
+          },
```
paultranvan, line 90: Would you mind adding another depth level e.g. `cozyMetadata.something.x` just to be safe?

packages/cozy-client/src/store/documents.spec.js
```diff
@@ -10,7 +10,10 @@ describe('extractAndMerge', () => {
       _id: 'b6ff135b34e041ffb2d4a4865f3e0a53',
       type: 'io.cozy.files',
       _type: 'io.cozy.files',
-      blabla: 'new field'
+      blabla: 'new field',
+      cozyMedata: {
```
paultranvan, line 14: s/cozyMedata/cozyMetadata ;)

packages/cozy-client/src/store/documents.spec.js
paultranvan: I minded ;)

paultranvan (approved):

PR #1193, opened 2022-06-27
Fix partial filter

paultranvan:
When a partial filter is defined and the index is not created yet,
CouchDB might fallback on another existing index and return a warning
saying the specified index was not used.
This is not what we want, because the existing index might not actually
use the partial filter and make a full-scan on the index to evaluate
it.

### DIFF
--- a/packages/cozy-stack-client/src/Collection.js
+++ b/packages/cozy-stack-client/src/Collection.js
@@ -41,6 +41,18 @@ export const isIndexConflictError = error => {
 }
 
 /**
+ * Helper to identify a not used index
+ *
+ * @param {string} warning - The warning returned by CouchDB
+ * @returns {boolean} Whether or not this is a not used index warning
+ */
+export const isIndexNotUsedWarning = warning => {
+  return warning.match(
+    /was not used because it does not contain a valid index for this query/
+  )
+}
+
+/**
  * Helper to identify a no usable index error
  *
  * @param {Error} error - An error
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -438,7 +438,7 @@ class DocumentCollection {
   }
 
   toMangoOptions(selector, options = {}) {
-    let { sort, indexedFields } = options
+    let { sort, indexedFields, partialFilter } = options
     const { fields, skip = 0, limit, bookmark } = options
 
     sort = transformSort(sort)
@@ -446,8 +446,15 @@ class DocumentCollection {
     indexedFields = indexedFields
       ? indexedFields
       : getIndexFields({ sort, selector })
+
+    const partialFilterFields = partialFilter
+      ? getIndexFields({ partialFilter })
+      : null
     const indexName =
-      options.indexId || `_design/${getIndexNameFromFields(indexedFields)}`
+      options.indexId ||
+      `_design/${getIndexNameFromFields(indexedFields, {
+        partialFilterFields
+      })}`
     if (sort) {
       const sortOrders = uniq(
         sort.map(sortOption => head(Object.values(sortOption)))
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -3,6 +3,7 @@ import { uri, attempt, sleep } from './utils'
 import uniq from 'lodash/uniq'
 import omit from 'lodash/omit'
 import head from 'lodash/head'
+import merge from 'lodash/merge'
 import startsWith from 'lodash/startsWith'
 import qs from 'qs'
 import { MangoQueryOptions } from './mangoIndex'
@@ -12,7 +13,8 @@ import Collection, {
   isIndexNotFoundError,
   isIndexConflictError,
   isNoUsableIndexError,
-  isDocumentUpdateConflict
+  isDocumentUpdateConflict,
+  isIndexNotUsedWarning
 } from './Collection'
 import {
   getIndexNameFromFields,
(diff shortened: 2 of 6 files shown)


packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -468,9 +479,14 @@ class DocumentCollection {
...
     }
+    // We need to pass the partialFilter in the selector, otherwise CouchDB might
+    // fallback on another index if it does not exist yet
+    const mergedSelector = partialFilter
+      ? merge(selector, partialFilter)
+      : selector
 
     const opts = {
-      selector,
+      selector: mergedSelector,
```
Crash--, line 489: Can you explain why this is needed? Since now we return an error if the index is not correct, Cozy Client will stop its flow, create the correct index and retry the request. Why do we need to pass to partial filter to the selector?

paultranvan: This is not required per se, but the CouchDB doc suggests to add the content of the partial index in the selector, so it seems worth adding it: https://docs.couchdb.org/en/stable/api/database/find.html#find-partial-indexes

packages/cozy-stack-client/src/DocumentCollection.spec.js
```diff
@@ -573,6 +573,31 @@ describe('DocumentCollection', () => {
       ).resolves.toBe(FIND_RESPONSE_FIXTURE)
     })
 
+    it('should not throw an error when a warning is returned and a partial filter is defined', async () => {
```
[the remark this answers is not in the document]
paultranvan: We throw inside the method, but we catch it to create the index. So, for this test, it's like it never throwed

packages/cozy-stack-client/src/DocumentCollection.js
Crash--: 🤷‍♂️

packages/cozy-stack-client/src/Collection.js
```diff
@@ -40,6 +40,18 @@ export const isIndexConflictError = error => {
...
+/**
+ * Helper to identify a not-used index
+ *
+ * @param {string} warning - The warning returned by CouchDB
+ * @returns {boolean} Whether or not this is a not-used index warning
+ */
+export const isIndexNotUsedWarning = warning => {
+  return warning.match(
+    /was not used because it does not contain a valid index for this query/
+  )
```
Crash--, line 52: Are we sure that we have the same message between Couch 2 and Couch 3 ? (sorry for this kind of question)

Crash-- (approved):

packages/cozy-stack-client/src/Collection.js
```diff
@@ -40,6 +40,18 @@ export const isIndexConflictError = error => {
   return error.message.match(/error_saving_ddoc/)
 }
 
+/**
+ * Helper to identify a not-used index
```
trollepierre, line 44:
```suggestion
 * Helper to identify a not used index
```

packages/cozy-stack-client/src/Collection.js
```diff
@@ -40,6 +40,18 @@ export const isIndexConflictError = error => {
...
+ *
+ * @param {string} warning - The warning returned by CouchDB
+ * @returns {boolean} Whether or not this is a not-used index warning
```
trollepierre, line 47:
```suggestion
 * @returns {boolean} Whether or not this is a not used index warning
```

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -260,6 +262,15 @@ class DocumentCollection {
...
     try {
       resp = await this.fetchDocumentsWithMango(path, selector, options)
+      if (
+        resp.warning &&
+        options.partialFilter &&
+        isIndexNotUsedWarning(resp.warning)
+      ) {
+        // This warning might happen when an index including a partial filter
+        // is not created yet.
+        throw new Error('no_index')
```
trollepierre, line 272: maybe we should use a more precise error message:
```suggestion
        throw new Error('no index including a partial filter created')
```

trollepierre (approved):

packages/cozy-stack-client/src/Collection.js
paultranvan: No need to be sorry, those are valid questions ;)
And yes, I checked on Couchdb 2.3 and 3.2

packages/cozy-stack-client/src/DocumentCollection.js
paultranvan: No, this is the error produced by the stack when there is [no usable index](https://github.com/cozy/cozy-stack/blob/ac45c1e8f685cceb9fd0aa9dd44937f95368bb69/pkg/couchdb/errors.go#L231), here we generate the same error to be consistent and trap it thanks to this helper:  https://github.com/cozy/cozy-client/blob/6ef72b9296f796afac002d25cd73079a85fa9fa2/packages/cozy-stack-client/src/DocumentCollection.js#L275

Crash--: We can't use your suggestion. The goal is to run this method https://github.com/cozy/cozy-client/blob/6ef72b9296f796afac002d25cd73079a85fa9fa2/packages/cozy-stack-client/src/DocumentCollection.js#L278-L279 

So we need to throw something passing this test: https://github.com/cozy/cozy-client/blob/6ef72b9296f796afac002d25cd73079a85fa9fa2/packages/cozy-stack-client/src/Collection.js#L29-L31 . 

This `no_index` is raised by couch / stack

paultranvan: I win by speed, @Crash-- wins by completeness ;)

PR #1196, opened 2022-07-08
fix: Handle query limit for first and last paginated query

paultranvan:
When enforcing the query limit in the store, we relied on the query
count to slice the docs. However, the count can be inferior to the
limit for the fisrt and last paginated query, leading to issues with
the slice.


Merkur39 (approved):
LGTM

Crash--:
Thanks for the fix @paultranvan !

PR #1197, opened 2022-07-18
fix: Check the correct path to get the warning

Crash--:
We were checking only in resp.warning, but we can also have this warning in resp.meta.warning. 

We now handle the both case

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -262,11 +262,8 @@ class DocumentCollection {
     let resp
     try {
       resp = await this.fetchDocumentsWithMango(path, selector, options)
-      if (
-        resp.warning &&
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -92,7 +92,7 @@ class DocumentCollection {
    *                                        using `normalizeDoc`
    */
   static normalizeDoctypeRawApi(doctype) {
-    return function(data, response) {
+    return function (data, response) {
```
paultranvan, line 95: The CI lint does not seem happy with those 2 changes

paultranvan (commented):
A lint issue, but LGTM, thanks!

paultranvan (approved):

trollepierre (approved):

PR #1198, opened 2022-07-20
fix: Mark the query as loaded when the document does not exist

doubleface:


packages/cozy-client/src/store/queries.js
```diff
@@ -165,7 +165,12 @@ const query = (state = queryInitialState, action, documents) => {
     case RECEIVE_QUERY_RESULT: {
       const response = action.response
       if (!response.data) {
-        return state
```
paultranvan, line 168: nit: I think it would be worth adding a comment here saying that `data` might be null when querying a single doc that does not exist

PR #1201, opened 2022-07-21
fix: Lodash.merge is not creating a new object

Crash--:
By not creating a new object, we kept the same
referencial equality and then some components
that checked this equality was not re-rendering.


trollepierre (approved):
👌

paultranvan (approved):
Nice catch :clap:

PR #1203, opened 2022-07-25
chore(docs): Explicit param description

trollepierre:


packages/cozy-client/src/authentication/mobile.js
```diff
@@ -16,7 +16,7 @@ const win = typeof window !== 'undefined' ? window : null
 /**
  * Open a SafariView Controller and resolve with the URL containing the token
  *
- * @param {string} url
+ * @param {string} url - Url containing the token
```
Crash--, line 19: This is not the token yet. 

So URL containing: access_code, state

packages/cozy-client/src/authentication/mobile.js
```diff
@@ -55,7 +55,7 @@ const authenticateWithSafari = url => {
 /**
  * Opens an InAppBrowser and resolves with the URL containing the token
  *
- * @param {string} url
+ * @param {string} url - Url containing the token
```
Crash--, line 58: same here

PR #1204, opened 2022-07-25
chore: Browserlist-update

trollepierre:


Crash--:
Did you have trouble with the current version of caniuse-lite? 

It seems that it comes from parcel: 

```
 yarn why caniuse-lite
yarn why v1.22.18
[1/4] 🤔  Why do we have the module "caniuse-lite"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "caniuse-lite@1.0.30001296"
info Reasons this module exists
   - "_project_#cozy-pouch-link#parcel#browserslist" depends on it
   - Hoisted from "_project_#cozy-pouch-link#parcel#browserslist#caniuse-lite"
   - Hoisted from "_project_#cozy-pouch-link#parcel#cssnano#cssnano-preset-default#postcss-merge-rules#caniuse-api#caniuse-lite"
info Disk size without dependencies: "3.26MB"
info Disk size with unique dependencies: "3.26MB"
info Disk size with transitive dependencies: "3.26MB"
info Number of shared dependencies: 0
✨  Done in 0.39s.
```

So we should upgrade parcel instead? Or remove it if we don't use it anymore to dev?

PR #1202, opened 2022-07-22
feat: Add permissions's translation

Noemie21:

PR #1209, opened 2022-07-27
Add permissions translation

Noemie21:
I forgot to export my translations in my last PR which is already merged : https://github.com/cozy/cozy-client/pull/1202


Crash--:
@Noemie21 can you rebase your branch please?

packages/cozy-client/src/models/doctypes/index.js
```diff
@@ -0,0 +1,3 @@
+import * as locales from './locales'
```
trollepierre, line 1: Capture d’écran 2022-07-28 à 11 52 04

Are we sure this is working?

tests are failing

looks like we should do something like

import * as locales from `./locales/${lang}`

Crash--: We should create an index.js file inside ./locales that export local by local

Noemie21: done

packages/cozy-client/src/models/doctypes/locales/index.js
```diff
@@ -0,0 +1,15 @@
...
+for (const lang of langs) {
+  let locales = {}
+  try {
+    locales = require(`./${lang}.json`)
+  } catch (e) {
+    // eslint-disable-line no-empty-block
+  }
+  const polyglot = new Polyglot()
+  polyglot.extend(locales)
+  polyglots[lang] = polyglot
```
Noemie21, line 14: @Merkur39 should we add 'node-polyglot' here?

docs/api/cozy-client/modules/models.doctypes.locales.md
```diff
@@ -0,0 +1,105 @@
+[cozy-client](../README.md) / [models](models.md) / [doctypes](models.doctypes.md) / locales
```
trollepierre, line 1: i wonder if we should document that part.

Crash-- (approved):

PR #1210, opened 2022-07-28
chore(travis): Stop running on node 12

trollepierre:


Crash--:
I'll wait the end of august before approving. I want to be sure that the code we produce during that time is still compatible with node 12.

trollepierre:
@Crash-- : is it ready?

Crash-- (approved):

PR #1211, opened 2022-08-01
feat(background-fetching): Introduce isFetching in QueryState to handle refetch

trollepierre:
##  Context: 

We want to persist data from Cozy-Client, inside a storage, in order to reuse them for next loading of the app (without send any request to Pouch / Stack). Redux-persist helps us to persist redux store.

However, when the data got rehydrated, the requests have already a fetchStatus "loaded",  so they are not updated by  Cozy-Client, because the data already existed. 

That's why we used fetchPolicy, to refresh data older than a certain amount of hours / days.
However there are somme drawbacks:
- a delay value is totally arbitrary, so not reliable
- when fetchPolicy is not respected, then the query is reexecuted, passing the fetchStatus to loading again, that make application displaying a Spinner, which is bad for User Experience.

This PR follows the same concept of background fetching from React Query (https://react-query.tanstack.com/guides/background-fetching-indicators & https://react-query.tanstack.com/guides/window-focus-refetching) 

## Implementation

- a query has a loaded fetchStatus
- this query is restarted (by another component or a rehydratation)
- then the query keeps the loaded fetchStatus, the applicaations keep displaying the data, while the request got executing in background. A `isFetching: true` attribute is added inside the QueryState
- When CozyClient finally load the data to the store, the application refreshes instantly with fresh data

This behavior should only concern React part of Cozy Client (useQuery, ObservableQuery). The client.query() should not implement this, because inside Node usage, we will have issues.


packages/cozy-client/src/store/queries.js
```diff
@@ -48,6 +48,7 @@ const queryInitialState = {
   id: null,
   definition: null,
   fetchStatus: 'pending',
+  isFetching: false,
```
trollepierre, line 51: I am afraid this implementation will affect Node part. But I wonder how to add the feature elsewhere. 🤔

packages/cozy-client/src/utils.js
```diff
@@ -39,7 +39,11 @@ export const isQueryLoading = col => {
     console.warn('isQueryLoading called on falsy value.') // eslint-disable-line no-console
     return false
   }
-  return col.fetchStatus === 'loading' || col.fetchStatus === 'pending'
+  return (
+    col.fetchStatus === 'loading' ||
+    col.fetchStatus === 'pending' ||
+    col.isFetching
```
trollepierre, line 45: I need to have a deeper look on the CozyClient code, to find if there is other place like this, what need a change.

packages/cozy-client/src/store/queries.js
```diff
@@ -148,19 +149,24 @@ const query = (state = queryInitialState, action, documents) => {
         id: action.queryId,
         definition: action.queryDefinition,
         options: action.options,
-        // When the query is new, we set "fetchStatus" to "loading"
-        // directly since we know it will be loaded right away.
-        // This way, the loadQuery action will have no effect, and
-        // we save an additional render.
```
trollepierre, line 154: this comment was outdated

packages/cozy-client/src/store/queries.js
```diff
@@ -148,19 +149,24 @@ const query = (state = queryInitialState, action, documents) => {
...
       }
     case LOAD_QUERY:
       if (state.fetchStatus === 'loading') {
         return state
       }
+      if (state.fetchStatus === 'loaded') {
+        return {
+          ...state,
+          fetchStatus: 'loaded',
+          isFetching: true
```
trollepierre, line 163: the heart of the feature is here: once the query has already been loaded, we don't set fetchStatus to loading.

packages/cozy-client/src/types.js
```diff
@@ -54,6 +54,7 @@ import { QueryDefinition } from './queries/dsl'
  * @property {string} id
  * @property {QueryDefinition} definition
  * @property {QueryFetchStatus} fetchStatus
+ * @property {boolean} isFetching
```
trollepierre, line 57: I should also add directly here, the documentation, about isFetching (and take the time to document a little bit, the attribute of QueryState

packages/cozy-client/src/store/queries.js
trollepierre: Maybe my implementation should stay closer to fetchPolicy. But in my opinion, add a queryResult here, is practical for usage. (see in the examples of react-integration.md)

docs/react-integration.md
```diff
@@ -172,31 +175,38 @@ The following props will be given to your wrapped component:
  - `fetchStatus`: the status of the fetch (`pending`, `loading`, `loaded` or `error`)
  - `lastFetch`: when the last fetch occurred
  - `hasMore`: the fetches being paginated, this property indicates if there are more documents to load
+ - `isFetching`: a boolean that you can use to show that it's in a fetching state, regardless of the state of the `fetchStatus` variable.
```
Crash--, line 178: isFetching should only be there if we have a "background sync" attribute somewhere. Maybe based on an attribute of useQuery() or because we've a global backgroundSync attribute. 

In other case, I don't think this attribute should be set.

docs/react-integration.md
```diff
@@ -172,31 +175,38 @@ The following props will be given to your wrapped component:
...
-    {
-      ({ data, fetchStatus }) =>
+const TodoList = ({ props }) => (
+  <Query query={queryDefinition} checked={props.checked}>
+    {({ data, fetchStatus, isFetching }) => (
       fetchStatus !== 'loaded'
         ? <h1>Loading...</h1>
-        : <ul>{data.map(todo => <li>{todo.label}</li>)}</ul>
+        : <>
+          {isFetching && <h1>Refreshing data...</h1>}
```
Crash--, line 201: See my previous comment. Since the QDef is not talking about backgroundSync, isFetching should not be set to true

docs/react-integration.md
```diff
@@ -386,16 +407,20 @@ import React from 'react'
...
+
+              <ul>
+                {data.map(todo => (
+                        <li key={todo.id}>{todo.label}</li>
+                ))}
+              </ul>
+            </>
     )
   }
 }
```
Crash--, line 426: I can't comment on the next commit, the rename one. But all the React stuff should be inside JSX file, not inside JS file.

packages/cozy-client/src/store/queries.js
Crash--: It'll affect since we'll have a new QueryState attribute, but it's fine.

packages/cozy-client/src/utils.js
Crash--: This is exactly what I don't want to see in Cozy-Client. Your query has data loaded, you should display them. `isFetching` should only be there to add a little information on the UI side. This is a different concept.

Crash-- (commented):
Few remarks: 
- Maybe we should add a per query attribute to enable background Fetching. And maybe if this attribute is here, we should override the fetchPolicy?
- Maybe we should add a global attribute to enable background fetching. I think we already have something to declare global error handler.

Crash--:
>  do you think we should support both options (the global attribute AND the "per query" attribute)?

I think we should follow what we have done for the onError callback: We can add it globally or having one per query (https://github.com/cozy/cozy-client/commit/c5f65d33e0ab29a22b409365151173a3b7645b56)

packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -1395,6 +1396,17 @@ describe('CozyClient', () => {
...
+    it('should not fetch policy when background fetching enabled', async () => {
+      jest.spyOn(client, 'requestQuery')
+      const fetchPolicy = jest.fn(() => false)
+      await client.query(query, {
+        as: 'allTodos',
+        fetchPolicy,
+        backgroundFetching: true
+      })
+      expect(fetchPolicy).not.toHaveBeenCalled()
+    })
```
Crash--, line 1408: I'm really not sure about that. `fetch policy` is useful for caching and to avoid fetch "spam" when rendering several components. 

`background fetching` will be more used for refreshing data when they already exists. 

It should not be one or the other. We should find a way to make them live together. 

Maybe that, by default, background fetching should respect the fetch policy. If it returns false, then we should not run the query. 

After that, we can create a plugin to force the background fetch, for instance, when giving the focus or something. 

What is your opinion about this feature?

trollepierre: After reconsideration:

I see 3 different features that can be totally independant:

- **fetchPolicies**, exactly what you define above

- **backgroundFetching** that handles queryState.fetchStatus + isFetching, to prevent the Spinner and the bad UX. Implemented in this PR, if I revert that change  line 917 in CozyClient.js and update this test    `expect(fetchPolicy).toHaveBeenCalled()`

- **forcingRefetch** - or something called [refetchOnWindowFocus in reactQuery](https://tanstack.com/query/v4/docs/guides/window-focus-refetching?from=reactQueryV3&original=https%3A%2F%2Freact-query-v3.tanstack.com%2Fguides%2Fwindow-focus-refetching), when giving the focus, when an app needs to refetch whatever the fetch policies.


I am pretty sure, the background fetching will make the fetchPolicies decrease. Maybe the `forcingRefetch` is not that urgent. Personally I think we should try to see how the 2 first features will collaborate, after considering implementing the `forcingRefetch`.

So I agree 100% with you.

Crash-- (approved):
Nice! I'll try it on a cozy app!

PR #1217, opened 2022-08-05
feat: Add `deconstructCozyWebLinkWithSlug` to extract data from Cozy url

Ldoppea:
Sometimes we have to manipulate Cozy's URLs as strings and we want to
deconstruct them to retrieve data like slug, or Cozy's name

This is the case on `cozy-notes` as we can retrieve a `returnUrl` from
the URL's searchParams. In a mobile app, we wan't to extract the target
slug from the `returnUrl` so we can call `openApp()` for this slug

In order to ease the process, `deconstructCozyWebLinkWithSlug()`
requires the Cozy's URL to contain a slug

Without knowing if the URL contains a slug or not, it would be not
possible to correctly extract data from the URL as we can't guess if a
part of the URL should be a subdomain, or if it contains a slug

___

Related PRs:

- https://github.com/cozy/cozy-notes/pull/318
- https://github.com/cozy/cozy-ui/pull/2206

### DIFF
--- a/packages/cozy-client/src/helpers.js
+++ b/packages/cozy-client/src/helpers.js
@@ -1,5 +1,7 @@
 import { Association } from './associations'
 
+import { CozyLinkData, SubdomainType } from './types'
+
 export const dehydrate = document => {
   const dehydrated = Object.entries(document).reduce(
     (documentArg, [key, value]) => {
@@ -74,6 +76,47 @@ export const generateWebLink = ({
   }
 
   return url.toString()
+}
+
+/**
+ * Deconstruct the given link in order to retrieve useful data like Cozy's name, domain, or slug
+ *
+ * The given link MUST contain a slug
+ *
+ * @param {string} webLink - link to deconstruct. It should be a link from a Cozy and containing a slug
+ * @param {SubdomainType} [subDomainType=flat] - whether the cozy is using flat or nested subdomains.
+ * @returns {CozyLinkData} Deconstructed link
+ */
+export const deconstructCozyWebLinkWithSlug = (
+  webLink,
+  subDomainType = 'flat'
+) => {
+  const url = new URL(webLink)
+
+  const slug =
+    subDomainType === 'nested'
+      ? url.host.split('.')[0]
+      : url.host.split('.')[0].split('-')[1]
+
+  const cozyName =
+    subDomainType === 'nested'
+      ? url.host.split('.')[1]
+      : url.host.split('.')[0].split('-')[0]
+
+  const cozyBaseDomain = url.host
+    .split('.')
+    .slice(subDomainType === 'nested' ? 2 : 1)
+    .join('.')
+
+  return {
+    cozyBaseDomain,
+    cozyName,
+    pathname: url.pathname,
+    hash: url.hash,
+    protocol: url.protocol,
+    searchParams: url.searchParams.toString(),
+    slug
+  }
 }
 
 export class InvalidProtocolError extends Error {
--- a/packages/cozy-client/src/helpers.spec.js
+++ b/packages/cozy-client/src/helpers.spec.js
@@ -1,6 +1,7 @@
 import { enableFetchMocks, disableFetchMocks } from 'jest-fetch-mock'
 
 import {
+  deconstructCozyWebLinkWithSlug,
   dehydrate,
   generateWebLink,
   rootCozyUrl,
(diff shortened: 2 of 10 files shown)


packages/cozy-client/src/helpers.spec.js
```diff
@@ -150,6 +151,83 @@ describe('generateWebLink', () => {
...
 
+fdescribe('deconstructWebLink', () => {
+  it.each([
+    [
+      'https://claude-notes.mycozy.cloud',
+      'flat',
+      {
+        protocol: 'https:',
+        cozyName: 'claude',
+        cozyDomain: 'mycozy.cloud',
```
Crash--, line 162: I'm not confortable with the `cozyDomain` notion. To me, in that case, the cozyDomain is : `claude.mycozy.cloud`

packages/cozy-client/src/helpers.spec.js
```diff
@@ -150,6 +151,83 @@ describe('generateWebLink', () => {
   })
 })
 
+fdescribe('deconstructWebLink', () => {
```
Crash--, line 154:
```suggestion
describe('deconstructWebLink', () => {
```

packages/cozy-client/src/types.js
```diff
@@ -380,4 +380,26 @@ import { QueryDefinition } from './queries/dsl'
...
+ * Subdomain type for a Cozy. Can be flat or nested subdomains
+ *
+ * Example of 'flat' domain: https://claude-notes.somedomain.fr
+ * Example of 'nested' domain: https://notes.claude.somedomain.fr
+ *
+ * @typedef {'flat'|'nested'} SubdomainType
+ */
+
+/**
+ * Represents the differents parts of a deconstructed Cozy link
```
trollepierre, line 393:
```suggestion
 * Represents the different parts of a deconstructed Cozy link
```

packages/cozy-client/src/types.js
```diff
@@ -380,4 +380,26 @@ import { QueryDefinition } from './queries/dsl'
...
+ *
+ * @typedef {object} CozyLinkData
```
trollepierre, line 395: nit: Why not calling this CozyLink or CozyLinkParams or FlattenCozyUrl?

packages/cozy-client/src/helpers.js
```diff
@@ -76,6 +78,47 @@ export const generateWebLink = ({
   return url.toString()
 }
 
+/**
+ * Deconstruct the given link in order to retrieve useful data like Cozy's name, domain, or slug
+ *
+ * The given link MUST contain a slug
+ *
+ * @param {string} webLink - link to deconstruct. It should be a link from a Cozy and containing a slug
+ * @param {SubdomainType} subDomainType - whether the cozy is using flat or nested subdomains. Defaults to flat.
```
trollepierre, line 87:
```suggestion
 * @param {SubdomainType} [subDomainType=flat] - whether the cozy is using flat or nested subdomains. 
```

trollepierre: https://jsdoc.app/tags-param.html

packages/cozy-client/src/helpers.js
```diff
@@ -76,6 +78,47 @@ export const generateWebLink = ({
...
+ * @returns {CozyLinkData} Deconstructed link
+ */
+export const deconstructCozyWebLinkWithSlug = (
```
trollepierre, line 90: nit: Why no calling it simply `deconstructCozyWebLink`

Ldoppea: I wanted to emphasise the fact that this method only works if the link contains a slug.

packages/cozy-client/src/helpers.spec.js
Ldoppea: Do you have any suggestion on how to name that part? Would `cozyBaseDomain` work?

packages/cozy-client/src/helpers.js
Ldoppea: Fixed in https://github.com/cozy/cozy-client/compare/637711a2193c5077ee2ada1cb7eec31c843ee6fc..f14039dd1b109d01d61f354cbd854fdb7804edf8

Thanks

packages/cozy-client/src/types.js
Ldoppea: Fixed in https://github.com/cozy/cozy-client/compare/637711a2193c5077ee2ada1cb7eec31c843ee6fc..f14039dd1b109d01d61f354cbd854fdb7804edf8

Thanks

packages/cozy-client/src/helpers.spec.js
Ldoppea: Fixed in https://github.com/cozy/cozy-client/compare/637711a2193c5077ee2ada1cb7eec31c843ee6fc..f14039dd1b109d01d61f354cbd854fdb7804edf8

Thanks

packages/cozy-client/src/types.js
Ldoppea: I'm wondering if `CozyLink` would bring confusion to what it contains. For me a link is a `string` or a `URL` object, I would not expect it to contain structured data. That said, the `URL` object contains those kind of data, so I'm not sure anymore.

Regarding `Params`, those are not params, but the description of the entire link. So I would not use this term.

`SplittedCozyUrl` would work maybe, but here again I'm wondering if someone can expect to find an array here like for the output from `string.split()`.

Maybe we should use `CozyLink` as you suggested but using a full class that allows to manipulate a link like for `URL`?

packages/cozy-client/src/helpers.spec.js
Crash--: Let's go for cozyBaseDomain

packages/cozy-client/src/types.js
trollepierre: After the whole discussion, I would go for CozyLinkData

packages/cozy-client/src/helpers.spec.js
Ldoppea: Fixed in https://github.com/cozy/cozy-client/compare/38c31b33efee7fe4485930f29698112f9be495c2..ada77f37c863df55c5e555de4960603ce5a1d48c

Crash-- (approved):

PR #1220, opened 2022-08-10
feat: Add warning when selector fields are not indexed

paultranvan:
When no indexFields option is given in a query definition, the fields to
index are automatically generated to create the index.
However, this might be dangerous as the indexed fields might actually
not entirely comply with the query objective.

### DIFF
--- a/packages/cozy-stack-client/src/ContactsCollection.spec.js
+++ b/packages/cozy-stack-client/src/ContactsCollection.spec.js
@@ -32,6 +32,9 @@ describe('ContactsCollection', () => {
     const col = new ContactsCollection('io.cozy.contacts', stackClient)
     return { stackClient, col }
   }
+  beforeEach(() => {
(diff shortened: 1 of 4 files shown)


Merkur39 (approved):
LGTM

trollepierre (approved):
thanks

Crash--:
Can you give us a use case where this is not complying with the query objective? 

Because with this PR we will add a lot of warning inside our applications. 

Also, should we do the same on pouch-link? cc @paultranvan

paultranvan:
This kind of query is problematic for instance: 
```
  const query = Q('io.cozy.files').where({
    $and: [
      {
        created_at: {
          $gt: '2021-01-01'
        }
      },
      {
        type: 'file'
      }
    ]
  })
```
We will actually try to index the `$and`.
Good point for pouch-link, we should add it as well...

PR #1222, opened 2022-08-12
feat(scripts): Add examples of scripts that adds and deletes files

trollepierre:


packages/cozy-client/examples/helpers/file.js
```diff
@@ -0,0 +1,116 @@
...
+const deleteAllFilesWithoutPath = async (client, dryRun) => {
+  const result = await getAllFiles(client)
+  console.log(result.length, 'files found inside this Drive')
+  const undefinedPathFiles = result.filter(({ path }) => !path)
+  console.log(
+    undefinedPathFiles.length,
+    'files without path found inside this Drive'
+  )
+
+  if (dryRun) {
```
paultranvan, line 86: I don't get why there is a dry run here? And it seems to be the reverse of what's expected: a dry run it not supposed to actually modify anything

packages/cozy-client/examples/delete-files.js
```diff
@@ -0,0 +1,43 @@
...
+    oauth: {
+      softwareID: 'io.cozy.client.cli'
+    }
+  })
+
+  const count = 10
+
+  const folder = await createFolderWithFilesBulk(client, count)
+  await deleteFile(client, folder)
+  await deleteAllFilesWithoutPath(client, true)
```
paultranvan, line 39: Do we need this every time? Have you tried adding a sleep in case this is the stack that is trying to trash the files in the background?

packages/cozy-client/examples/delete-files.js
```diff
@@ -0,0 +1,43 @@
...
+  await getAllFiles(client)
```
paultranvan, line 40: Why do we get all the files without reading the response?

packages/cozy-client/examples/helpers/file.js
```diff
@@ -0,0 +1,116 @@
...
+
+const schema = {
+  files: {
+    doctype: 'io.cozy.files'
+  }
+}
+
+const generateFakeFolder = () => ({
+  name: 'folder-name' + Math.random(),
+  icon: 'icon',
```
paultranvan, line 13: I don't think `icon` is an expected directory field: https://github.com/cozy/cozy-stack/blob/master/docs/files.md#post-filesdir-id

packages/cozy-client/examples/helpers/file.js
```diff
@@ -0,0 +1,116 @@
...
+  type: 'directory',
+  dirId: 'io.cozy.files.root-dir'
+})
+
+const generateFakeFile = (folder, i, file) => ({
+  name: `${folder.name}-${i}.txt`,
+  contentType: 'text/plain',
+  icon: 'icon',
```
paultranvan, line 21: Likewise, I don't think the icon is useful for files?

packages/cozy-client/examples/helpers/file.js
```diff
@@ -0,0 +1,116 @@
...
+
+const createFolder = async client => {
+  const { data: folder } = await client.stackClient
+    .collection('io.cozy.files')
+    .create(generateFakeFolder())
+  return folder
+}
+
+const createFileInsideFolder = async (client, count, folder) => {
+  const readme = fs.readFileSync('../README.md')
```
paultranvan, line 36: This makes an assumption on the current path,  I don't think it will work if we do `node packages/cozy-client/examples/delete-files.js`
`readFileSync` returns a Buffer, so maybe `const fakeFile = Buffer.from('this is a test')` might work?

packages/cozy-client/examples/helpers/file.js
```diff
@@ -0,0 +1,116 @@
...
+})
+
+const createFolder = async client => {
+  const { data: folder } = await client.stackClient
+    .collection('io.cozy.files')
+    .create(generateFakeFolder())
+  return folder
+}
+
+const createFileInsideFolder = async (client, count, folder) => {
```
paultranvan, line 35: Since there are several files to create, it should be plural ;) `createFilesInsideFolder`

packages/cozy-client/examples/helpers/file.js
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @param {CozyClient} client - Cozy client instance
+ * @returns {Promise<FileDocument>}
+ */
+const getAllFiles = async client => {
+  const query = new QueryDefinition({
+    doctype: 'io.cozy.files',
+    limit: null
+  })
+  const resp = await client.query(query)
```
paultranvan, line 106: You probably want `queryAll` instead of `query` if you want to get everything

paultranvan (commented):
Some comments, but this is nice to have!

packages/cozy-client/examples/helpers/file.js
trollepierre: ok removed

packages/cozy-client/examples/delete-files.js
trollepierre: ok removed

packages/cozy-client/examples/delete-files.js
trollepierre: ok removed

trollepierre:
@paultranvan or @Crash-- 🙏

packages/cozy-client/examples/README.md
```diff
@@ -0,0 +1,20 @@
...
+Example: to run the script file named `perfs`:
+
+```
+cd cozy-client/packages/cozy-client/examples
+node perfs.js
+```
+
+And then accept permissions / login.
+
+During the first login, it might crash with a `ERR_UNHANDLED_REJECTION`
```
paultranvan, line 16: We do not know why this error?

packages/cozy-client/examples/helpers/file.js
```diff
@@ -0,0 +1,110 @@
...
+
+const createFolder = async client => {
+  const { data: folder } = await client.stackClient
+    .collection('io.cozy.files')
+    .create(generateFakeFolder())
+  return folder
+}
+
+const createFilesInsideFolder = async (client, count, folder) => {
+  const readme = Buffer.from('this is a test')
```
paultranvan, line 33: nit: this var name does not seem relevant anymore ;)

paultranvan (approved):

packages/cozy-client/examples/README.md
trollepierre: No, I would have mentioned the reason if so.

And I didn't take the time to investigate.

Issue #866, opened 2021-01-27
console.warn is interpreted as an error by the stack in a service 


paultranvan:
In a node context, e.g. a service, the output of a console.warn is sent to stderr and logged [as an error](https://github.com/cozy/cozy-stack/blob/master/worker/exec/common.go#L87-L94) by the stack. To avoid this, it should be sent to stdout and be formatted in the JSON format [expected by the stack](https://github.com/cozy/cozy-stack/blob/master/worker/exec/service.go#L199-L205)

The issue doesn't exist with the konnectors, because it uses [cozy-logger](https://github.com/cozy/cozy-libs/tree/master/packages/cozy-logger) that does this formatting.
So, in order to have this behaviour in cozy-client, we could either: 

* Make cozy-client uses cozy-logger and adapt the format depending on the context execution (node vs web)
* Re-implement this logic directly in cozy-client, as it does not seem very complex

PR #1224, opened 2022-08-17
refactor: Use logger instead of console

Crash--:
Cozy-Client can be used in a web app but also in konnectors or services. 

When run in the browser, there is no issue using console.* directly. But when it comes to run it on the server (services or server side konnectors), logs need to have a special format that cozy-client doesn't respect for now. See this issue https://github.com/cozy/cozy-client/issues/866 

The idea is to not use console.* anymore in CC, but use the logger. This is the first step. After that, we'll be able to format correctly the log.

### DIFF
--- a/packages/cozy-stack-client/src/logger.js
+++ b/packages/cozy-stack-client/src/logger.js
@@ -1,0 +1,6 @@
+import minilog from '@cozy/minilog'
+
+const logger = minilog('cozy-stack-client')
+minilog.suggest.deny('cozy-stack-client', 'info')
+
+export default logger
--- a/docs/api/cozy-client/README.md
+++ b/docs/api/cozy-client/README.md
@@ -54,7 +54,7 @@ cozy-client
 
 *Defined in*
 
-[packages/cozy-client/src/queries/dsl.js:495](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/queries/dsl.js#L495)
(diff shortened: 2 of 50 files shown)


Merkur39 (approved):
LGTM

trollepierre (commented):
could we update `"no-console": 0,` into `"no-console": "error", // Use logger instead of console (see https://github.com/cozy/cozy-client/pull/1224 )`

packages/cozy-stack-client/src/logger.js
```diff
@@ -0,0 +1,6 @@
+import minilog from '@cozy/minilog'
```
paultranvan, line 1: In services/connectors we use cozy-logger: https://github.com/cozy/cozy-libs/tree/master/packages/cozy-logger . Eventually, do we plan to use only one of them to log them all?

paultranvan (approved):
Very nice initiative :clap:

Crash--:
> could we update "no-console": 0, into "no-console": "error", // Use logger instead of console (see https://github.com/cozy/cozy-client/pull/1224 ) inside .eslintrc.json?

I don't like that. There is few places where console.* are useful 

I prefer to be careful and let the dev think if the code will be used in production or not instead of forcing the use of logger  without understanding what is going on.

packages/cozy-stack-client/src/logger.js
Crash--: That's the goal at the end, to have only one logger that handle correctly browser & node. Though,  I don't know the path to do that

PR #1225, opened 2022-08-17
fix: Use find method where args are for find

### LINKED ISSUE #1216: .where selector is mandatory when using .select or .partialIndex
When a query is run without a selector and sort, the Collection.all will be used: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/StackLink.js#L86-88.
For instance:

```
client.find(Q(doctype).partialIndex({trashed: false}).fields(['trashed', 'dir_id', 'name']))
```

Furthermore, a sort cannot be used without a selector according to https://github.com/cozy/cozy-client/issues/405

When the Collection.all is used, some query options cannot work such as `.partialIndex` or `.select`, as the  `_all_docs` API will be used , that does not support such features.

Crash--:
The idea is to check if there is some options
related to _find. If yes, then we should use
this route instead of {normal|all}_docs

Fix #1216

### DIFF
--- a/packages/cozy-client/src/StackLink.js
+++ b/packages/cozy-client/src/StackLink.js
@@ -1,10 +1,24 @@
-import { MutationTypes } from './queries/dsl'
+import { MutationTypes, QueryDefinition } from './queries/dsl'
 import CozyLink from './CozyLink'
-import { CozyClientDocument, CouchDBBulkResult } from './types'
+import { CozyClientDocument, CouchDBBulkResult, ClientResponse } from './types'
 import { DOCTYPE_FILES } from './const'
 import { BulkEditError } from './errors'
 import zipWith from 'lodash/zipWith'
-
+/**
+ *
+ * To know if cozy-client should use Document.find()
+ * or Document.all()
+ * Similar to what is done in CozyPouchLink executeQuery()
+ *
+ * @param {QueryDefinition} queryDefinition - QueryDefinition to check
+ * @returns {boolean} If has find options
+ *
+ */
+const hasFindOptions = queryDefinition => {
+  const { selector, partialFilter, sort, fields } = queryDefinition
+  if (selector || partialFilter || sort || fields) return true
+  return false
+}
 /**
  * Returns full documents after a bulk update
  *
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/StackLink.js
```diff
@@ -1,10 +1,24 @@
...
+ * Similar to what is done in CozyPouchLink executeQuery()
+ *
+ * @param {QueryDefinition} queryDefinition - QueryDefinition to check
+ * @returns {boolean} If has find options
+ *
+ */
+const hasFindOptions = queryDefinition => {
+  const { selector, partialFilter, sort, fields } = queryDefinition
+  if (selector || partialFilter || sort || fields) return true
+  return false
```
paultranvan, line 20: don't you prefer `return selector || partialFilter || sort || fields` ?

packages/cozy-client/src/StackLink.spec.js
```diff
@@ -46,6 +46,22 @@ describe('StackLink', () => {
...
+      expect(stackClient.collection).toHaveBeenCalledWith('io.cozy.todos')
+    })
+
+    it('should use find if fields are given', async () => {
+      const query = Q('io.cozy.todos').select(['trashed'])
+      stackClient.collection().find.mockReset()
+      await link.request(query)
+      expect(stackClient.collection().find).toHaveBeenCalled()
+      expect(stackClient.collection).toHaveBeenCalledWith('io.cozy.todos')
+    })
```
paultranvan, line 64: It might be worth adding a test for `all`
```
it('should use all when the whole doctype is queried',  async () => {
    const query = Q('io.cozy.todos')
    stackClient.collection().all.mockReset()
    await link.request(query)
    expect(stackClient.collection().all).toHaveBeenCalled()
    expect(stackClient.collection).toHaveBeenCalledWith('io.cozy.todos')
  })
 ```

paultranvan (approved):
Thanks!

packages/cozy-client/src/StackLink.spec.js
Crash--: https://github.com/cozy/cozy-client/blob/4cdda31aa62a39e569defa737aef0bc33aa4901c/packages/cozy-client/src/StackLink.spec.js#L42-L48

already there ;)

paultranvan: Uh, my bad

packages/cozy-client/src/StackLink.js
Crash--: I do not have any pref. At first, this method was more complicated than just returning the fields. Since I wrote it that way, I will commit that way. Maybe next time you'll change it :p

packages/cozy-client/src/StackLink.spec.js
Crash--: Though, next time, I'll rename the test! it's not relevant ATM.

PR #1227, opened 2022-08-17
feat: Fetch capabilities on OAuth login

Ldoppea:
Current cozy-client implementation calls `loadInstanceOptionsFromDOM`
when initialized

This method only works for web app clients, but would fail for OAuth
clients as those cannot rely on any DOM

To fix this we should retrieve client's options by calling the
cozy-stack

For now we only need to retrieve capabilities. But other props should
be implemented later

As this method is async, we don't want to call it from the constructor.
Instead we chose to call it at the end of `login()`. Which would also
allow to access token data when we will implement this part

___

This PR also modify `useCapabilities` hook to get `client.capabilities` if
it exists instead of fetching info from cozy-stack.

This introduce a breaking change (described in the commit)

___

After merging this, we should update the following code:

- [cozy-drive/HistoryModal.jsx](https://github.com/cozy/cozy-drive/blob/6d07b39067a867e6092ff82dbb0bce3a24acda7d/src/components/FileHistory/HistoryModal.jsx#L33-L37)
- [cozy-drive/CreateNoteItem.jsx](https://github.com/cozy/cozy-drive/blob/6d07b39067a867e6092ff82dbb0bce3a24acda7d/src/drive/web/modules/drive/Toolbar/components/CreateNoteItem.jsx#L22-L26)
- [cozy-drive/FolderViewBody.jsx](https://github.com/cozy/cozy-drive/blob/6d07b39067a867e6092ff82dbb0bce3a24acda7d/src/drive/web/modules/views/Folder/FolderViewBody.jsx#L66-L70)
- [cozy-ui/ForwardButton.jsx](https://github.com/cozy/cozy-ui/blob/84000049dddf367701e44597ce7f944e19953942/react/Viewer/Footer/ForwardButton.jsx#L36)

___

Related PRs:

- cozy/cozy-react-native#391

### DIFF
--- a/docs/api/cozy-client/README.md
+++ b/docs/api/cozy-client/README.md
@@ -645,7 +645,7 @@ An array with all apps in maintenance
 
 *Defined in*
 
-[packages/cozy-client/src/hooks/useCapabilities.jsx:4](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/hooks/useCapabilities.jsx#L4)
+[packages/cozy-client/src/hooks/useCapabilities.jsx:5](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/hooks/useCapabilities.jsx#L5)
 
 ***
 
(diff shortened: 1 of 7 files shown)


Crash-- (approved):
Thank!

PR #1228, opened 2022-08-17
feat: Add energy contract qualification

KillianCourvoisier:
Adding specific qualification for contracts with energy providers


paultranvan (approved):
:sunglasses:

PR #1229, opened 2022-08-17
fix: Export `deconstructCozyWebLinkWithSlug` from `index.node.js`

Ldoppea:
In #1217 we forgot to expose this method from `index.node.js`


Merkur39 (approved):
😗

PR #1233, opened 2022-08-31
fix: Limit DocsIds according to limit instead of queryState.count

trollepierre:
Bugs happen when creating a shortcut, inside a new folder in Drive
queryState returns a data as an empty array, so count was 0
And sliceCount removes the new shortcut ids create, because it is 0
The fix using a limit - to 100 in our case, handles that case.
And should not break other scenarii.

### DIFF
--- a/packages/cozy-client/src/store.spec.js
+++ b/packages/cozy-client/src/store.spec.js
@@ -216,7 +216,7 @@ describe('Store', () => {
       expect(queryResult.data.map(x => x._id)).toEqual(['todo_1', 'todo_2'])
       queryResult = getQueryFromStore(store, 'Q2')
       expect(queryResult.data.length).toBe(2)
-      expect(queryResult.data.map(x => x._id)).toEqual(['todo_3', 'todo_4'])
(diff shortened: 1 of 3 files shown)


trollepierre:
cc @paultranvan

packages/cozy-client/src/store.spec.js
```diff
@@ -216,7 +216,7 @@ describe('Store', () => {
       expect(queryResult.data.map(x => x._id)).toEqual(['todo_1', 'todo_2'])
       queryResult = getQueryFromStore(store, 'Q2')
       expect(queryResult.data.length).toBe(2)
-      expect(queryResult.data.map(x => x._id)).toEqual(['todo_3', 'todo_4'])
+      expect(queryResult.data.map(x => x._id)).toEqual(['todo_1', 'todo_2'])
```
trollepierre, line 219: failing job : https://app.travis-ci.com/github/cozy/cozy-client/builds/255039439

i am not convinced about that change

Crash--:
See https://github.com/cozy/cozy-client/pull/1236 for the fix

PR #1236, opened 2022-09-01
fix: Fix store issue with creation mutation

Crash--:
We had an issue when using the creation mutation.
Document's slice was correctly updated but the
queries' one was not correctly updated. Some of
this date was right, not the other.

This is because since
0f2e8a3f2f13157723fa6395929476d66b860dec
&
d986f59781557a97ff2545b228153325d18307b9

We try to respect the Query Definition even with
a limitBy attribute. But we were using the wrong
data.

Since we're in a mutation context, we need to
use the mutated data, not the one we got during the
request. So instead of relying on old `count` from
the previous query, we rely on `updatedData`.length
to get the right information.

### DIFF
--- a/packages/cozy-client/src/store/queries.js
+++ b/packages/cozy-client/src/store/queries.js
@@ -355,7 +355,7 @@ export const makeSorterFromDefinition = definition => {
  * @param  {DocumentsStateSlice} documents - A reference to the documents slice
  * @returns {QueryState} - Updated query state
  */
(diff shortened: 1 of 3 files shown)


trollepierre (approved):
nit: are you sure this test is failing without the fix added in queries?

Crash--:
> nit: are you sure this test is failing without the fix added in queries?

Yep.

PR #1238, opened 2022-09-06
fix: MatchingIndex with partialFilter

Crash--:
If we had an existing index with partial filter
but if we made a request with the same fields but
without partial filter, then isMatchingIndex
returned true and then we used it and then we
renamed it and we destroyed the existing index...

We fixed isMatchingIndex to check if
partialFilters are the same.


paultranvan:
Good catch and fast fix, what a guy :clap:

paultranvan (approved):

PR #1241, opened 2022-09-12
Fix/index checking

Crash--:
It is possible to create an index, but this index is not ready to use
yet.

In that case, we query again with the same fields to check if the index
is ready to be used or not.

We forgot to use partialFilter in that check.

Also added a few types fix.

### DIFF
--- a/packages/cozy-stack-client/src/mangoIndex.js
+++ b/packages/cozy-stack-client/src/mangoIndex.js
@@ -25,6 +25,8 @@ import isEqual from 'lodash/isEqual'
  * @property {string|null} [indexId] The _id of the CouchDB index to use for this request
  * @property {string|null} [bookmark] For bookmark-based pagination, the document _id to start from
  * @property {Array<string>} [indexedFields]
+ * @property {string} [use_index] Name of the index to use
+ * @property {boolean} [execution_stats] If true, we request the stats from Couch
  * @property {MangoPartialFilter|null} [partialFilter] An optional partial filter
  */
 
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/mangoIndex.js
```diff
@@ -80,14 +82,18 @@ export const transformSort = sort => {
...
+ * @param {MangoSelector} [options.selector] - Mango selector
+ * @param {MangoPartialFilter} [options.partialFilter] - An optional partial filter
  * @param {MangoSort} [options.sort] - The sorting parameters
  *
  * @returns {Array} - Fields to index
  */
-export const getIndexFields = ({ selector, partialFilter, sort = [] }) => {
+export const getIndexFields = ({
+  selector = [],
+  partialFilter = [],
```
paultranvan, line 94: selector and partialFilter are not array :thinking:

Crash--: Thanks fixed. Always blown by `const a = []; typeof a` 

We should create a real type for selector with a more restrictive defintion (and so for partialFilter since partialFilter is a selector)

packages/cozy-stack-client/src/mangoIndex.js
```diff
@@ -3,7 +3,7 @@ import get from 'lodash/get'
 import isEqual from 'lodash/isEqual'
 
 /**
- * @typedef {object} MangoPartialFilter
+ * @typedef {Object} MangoPartialFilter
```
paultranvan, line 6: Is it really `Object`?

paultranvan (approved):
Looks good, thanks!

packages/cozy-stack-client/src/mangoIndex.js
Crash--: You can do what you want... But please be patient, something bigger is coming ... ;)

paultranvan: O is certainly bigger than o, but I feel like you're talking about something else...

PR #1243, opened 2022-09-12
fix: QueryAll return empty array if error

Crash--:
Sometimes query can error and its legitimate to do so. For instance, if the database doesn't exist yet or else. In that case, since queryAll() is about fetching all the docs and that API doesn't respect query() API, let's return an empty array without throwing anything.

queryAll should ba as close as possible than query() itself

### DIFF
--- a/docs/api/cozy-client/classes/CozyClient.md
+++ b/docs/api/cozy-client/classes/CozyClient.md
@@ -83,7 +83,7 @@ Cozy-Client will automatically call `this.login()` if provided with a token and 
 
 *Defined in*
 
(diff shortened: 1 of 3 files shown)


paultranvan:
> queryAll should ba as close as possible than query() itself

Eventually, shouldn't we change this API so queryAll returns a `.data` just like query? Even though this would be a big BC...

paultranvan (approved):
CI needs to be pleased, but looks good!

Crash--:
Seriously, I don't know what to do with this API. Same with useQueryAll... 

I think I should not return an empty array every time. I should use something like https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/Collection.js#L10-L22

But that means that queryAll should respect "json api" response.

PR #1245, opened 2022-09-20
feat: Do not use NODE_ENV=development to detect authentication

doubleface:
Prefer to use process.env.COZY_CREDENTIALS format to choose
authentication mode sincecozy-jobs-cli does not set NODE_ENV=development
for cozy-konnector-dev command
[anymore](https://github.com/konnectors/libs/pull/851)

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -369,15 +369,16 @@ class CozyClient {
    */
   static fromEnv(envArg, options = {}) {
     const env = envArg || (typeof process !== 'undefined' ? process.env : {})
-    const { COZY_URL, COZY_CREDENTIALS, NODE_ENV } = env
+    const { COZY_URL, COZY_CREDENTIALS } = env
     if (!COZY_URL || !COZY_CREDENTIALS) {
       throw new Error(
         'Env used to instantiate CozyClient must have COZY_URL and COZY_CREDENTIALS'
       )
     }
-    if (NODE_ENV === 'development') {
+
+    try {
       options.oauth = JSON.parse(COZY_CREDENTIALS)
-    } else {
+    } catch (err) {
       options.token = COZY_CREDENTIALS.trim()
     }
     options.uri = COZY_URL.trim()
--- a/docs/api/cozy-client/classes/CozyClient.md
+++ b/docs/api/cozy-client/classes/CozyClient.md
@@ -83,7 +83,7 @@ Cozy-Client will automatically call `this.login()` if provided with a token and 
 
 *Defined in*
 
-[packages/cozy-client/src/CozyClient.js:1620](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L1620)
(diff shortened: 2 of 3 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -369,16 +369,19 @@ class CozyClient {
...
-    if (NODE_ENV === 'development') {
-      options.oauth = JSON.parse(COZY_CREDENTIALS)
-    } else {
-      options.token = COZY_CREDENTIALS.trim()
+
+    if (COZY_CREDENTIALS) {
+      try {
+        options.oauth = JSON.parse(COZY_CREDENTIALS)
+      } catch (err) {
+        options.token = env.COZY_CREDENTIALS.trim()
```
Crash--, line 383:
```suggestion
        options.token = COZY_CREDENTIALS.trim()
```

packages/cozy-client/src/CozyClient.js
```diff
@@ -369,16 +369,19 @@ class CozyClient {
...
       throw new Error(
         'Env used to instantiate CozyClient must have COZY_URL and COZY_CREDENTIALS'
       )
     }
-    if (NODE_ENV === 'development') {
-      options.oauth = JSON.parse(COZY_CREDENTIALS)
-    } else {
-      options.token = COZY_CREDENTIALS.trim()
+
+    if (COZY_CREDENTIALS) {
```
Crash--, line 379: We can remove this check since we already do it in https://github.com/cozy/cozy-client/pull/1245/files#diff-9d26647e8d41ba1414842f5d3dca72250075f14efed52997e19584197a66e2a6R373

packages/cozy-client/src/CozyClient.js
```diff
@@ -369,15 +369,16 @@ class CozyClient {
...
   static fromEnv(envArg, options = {}) {
     const env = envArg || (typeof process !== 'undefined' ? process.env : {})
-    const { COZY_URL, COZY_CREDENTIALS, NODE_ENV } = env
+    const { COZY_URL, COZY_CREDENTIALS } = env
     if (!COZY_URL || !COZY_CREDENTIALS) {
       throw new Error(
         'Env used to instantiate CozyClient must have COZY_URL and COZY_CREDENTIALS'
       )
     }
-    if (NODE_ENV === 'development') {
```
paultranvan, line 378: Isn't it a breaking change? Or do we consider that is not used anywhere anymore?

doubleface: This is supposed to no be used anymore since this NODE_ENV=development only comes from dev mode in connectors.

But it may be a breaking change. @Crash-- what do you think ?

Crash--: I don't think we should add a BC here. The previous code will still work as expected.

The only downside I saw with this approach is from a performance POV. In production, we now try to JSON.parse a string then we catch the error and the we trim the string. This is not optimal but this code is generally not in an hot path.

Crash-- (approved):

PR #1247, opened 2022-09-22
feat(qualification): Add more logic to getBoundT function

Merkur39:
The qualifications labels can be suffixed with an emoji (country flag) via a second parameter corresponding to the country

Today we limit ourselves to:
- country == 'fr' => (🇫🇷)
- country == 'stranger' => (stranger)
- country == undefined | something else => no suffix


Crash-- (approved):

paultranvan (approved):
Nice!

PR #1248, opened 2022-09-22
fix: Get apps method was returning wrong type for not-registry source

vdnj:
Any call to `collection('io.cozy.apps').getById('...')` with not-registry
source was failing to properly return data.


paultranvan:
The commit title and description could be improved as they are quite vague and does not quite describe the situation and fix.

Proposition:
 
 ```
 fix: Get apps method was returning wrong type for not-registry source
 
Any call to `collection('io.cozy.apps').getById('...')` with not-registry source was failing to properly return data.

 ```

Crash-- (approved):
thanks

PR #1257, opened 2022-10-26
feat: Add copy in FileCollection

zatteo:
Will be used here https://github.com/cozy/cozy-drive/pull/2727

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -418,6 +418,31 @@ class FileCollection extends DocumentCollection {
   }
 
   /**
+   * Copy a file.
+   *
+   * @param {string} id   - The file's id
+   * @param {string} [name]   - The file copy name
+   * @param {string} [dirId]   - The destination directory id
+   * @returns {Promise<object>}   - A promise that returns the copied file if resolved.
+   * @throws {FetchError}
+   *
+   */
+  async copy(id, name, dirId) {
+    const params = {
+      name: name === undefined ? undefined : sanitizeAndValidateFileName(name),
+      dirId
+    }
+    const path = uri`/files/${id}/copy`
+    const url = querystring.buildURL(path, params)
+
+    const resp = await this.stackClient.fetchJSON('POST', url)
+
+    return {
+      data: normalizeFile(resp.data)
+    }
+  }
+
+  /**
    * async deleteFilePermanently - Definitely delete a file
    *
    * @param  {string} id - The id of the file to delete
--- a/packages/cozy-stack-client/src/FileCollection.spec.js
+++ b/packages/cozy-stack-client/src/FileCollection.spec.js
@@ -923,6 +923,74 @@ describe('FileCollection', () => {
     })
   })
 
+  describe('copy', () => {
+    it('should copy a file', async () => {
+      const FILE_ID = 'd04ab491-2fc6'
+      const COPIED_FILE_ID = '59140416-b95f'
+      client.fetchJSON.mockReturnValue({
+        data: {
+          id: COPIED_FILE_ID,
+          type: 'io.cozy.files'
+        }
+      })
+      const result = await collection.copy(FILE_ID)
+      expect(client.fetchJSON).toHaveBeenCalledWith(
+        'POST',
+        '/files/d04ab491-2fc6/copy'
+      )
+      expect(result).toEqual({
+        data: {
+          id: COPIED_FILE_ID,
+          type: 'io.cozy.files',
+          _id: COPIED_FILE_ID,
+          _rev: undefined,
+          _type: 'io.cozy.files'
+        }
+      })
+    })
+
+    it('should copy a file with custom name and directory', async () => {
+      const FILE_ID = 'd04ab491-2fc6'
+      const COPIED_FILE_ID = '59140416-b95f'
+      const COPIED_FILE_NAME = 'newName'
+      const COPIED_DIR_ID = '41686c35-9d8e'
+      client.fetchJSON.mockReturnValue({
+        data: {
+          id: COPIED_FILE_ID,
+          type: 'io.cozy.files',
+          attributes: {
+            name: COPIED_FILE_NAME,
+            dir_id: COPIED_DIR_ID
+          }
+        }
+      })
+      const result = await collection.copy(
+        FILE_ID,
+        COPIED_FILE_NAME,
+        COPIED_DIR_ID
+      )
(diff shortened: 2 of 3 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -417,6 +417,18 @@ class FileCollection extends DocumentCollection {
...
+   * Copy a file.
+   *
+   * @param {string} id   - The file's id
+   * @returns {Promise}   - A promise that returns the copied file if resolved.
+   * @throws {FetchError}
+   *
+   */
+  async copy(id) {
+    return await this.stackClient.fetchJSON('POST', `/files/${id}/copy`)
+  }
```
Crash--, line 430: The route can take more parameters than just this `id`. See https://docs.cozy.io/en/cozy-stack/files/#post-filesfile-idcopy 

Since this part of CozyClient is to consume the stack's APIs, can you add this capabilities?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -417,6 +417,18 @@ class FileCollection extends DocumentCollection {
...
+  /**
+   * Copy a file.
+   *
+   * @param {string} id   - The file's id
+   * @returns {Promise}   - A promise that returns the copied file if resolved.
+   * @throws {FetchError}
+   *
+   */
+  async copy(id) {
+    return await this.stackClient.fetchJSON('POST', `/files/${id}/copy`)
```
Crash--, line 429: Also if we don't need to wait for the reponse in the method, we don't await it.

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -417,6 +417,18 @@ class FileCollection extends DocumentCollection {
     return this.stackClient.fetchJSON('POST', uri`/files/trash/${id}`)
   }
 
+  /**
+   * Copy a file.
+   *
+   * @param {string} id   - The file's id
```
Crash--, line 423:
```
@param {FileAttributes['id']} id   - The file's id
```

I know this is not used anywhere else, but let's make the code better ;)

zatteo: I tried this way but got

`JSDOC_ERROR: ERROR: Unable to parse a tag's type expression for source file /Users/theo/Cozy/cozy-client/packages/cozy-stack-client/src/FileCollection.js in line 420 with tag title "param" and text "{FileAttributes['id']} id   - The file's id": Invalid type expression "FileAttributes['id']": Expected "!", "#", "$", "(", "-", ".", "/", "0", ":", "<", "=", "?", "@", "[]", "\\", "_", "|", "~", "‌", "‍", Unicode combining mark, Unicode decimal number, Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode punctuation connector, Unicode titlecase letter, Unicode uppercase letter, [1-9], or end of input but "[" found.`

Tried some other ways too without success for the moment

packages/cozy-stack-client/src/FileCollection.js
zatteo: Updated!

paultranvan (approved):

packages/cozy-stack-client/src/FileCollection.spec.js
```diff
@@ -923,6 +923,30 @@ describe('FileCollection', () => {
     })
   })
 
+  describe('copy', () => {
```
Crash--, line 926: @zatteo can you add a test with name & dirId 🙏

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -417,6 +417,27 @@ class FileCollection extends DocumentCollection {
...
+   * @param {string} name   - The file copy name
+   * @param {string} dirId   - The destination directory id
```
Crash--, line 425: Since those params are optional you can mark them like this: 
```
 * @param {string} [name]   - The file copy name
 * @param {string} [dirId]   - The destination directory id
 ```

File: packages/cozy-stack-client/src/FileCollection.js [hunk 6]
```diff
@@ -417,6 +417,27 @@ class FileCollection extends DocumentCollection {
...
+   */
+  copy(id, name, dirId) {
+    const params = {
+      name: name === undefined ? undefined : sanitizeAndValidateFileName(name),
+      dirId
+    }
+    const path = uri`/files/${id}/copy`
+    const url = querystring.buildURL(path, params)
+
+    return this.stackClient.fetchJSON('POST', url)
```
Crash--, line 438: At first, when reading your code, I though that the stack doesn't return any content. But this is not the case : https://docs.cozy.io/en/cozy-stack/files/#response_9 

If its a success, then the stack respond with the new created document. And this is important to deal with it in CozyClient even if for this method, this is not very important **yet**. 

I will not take the time to write a fully detailed comment but: 
1/ If we use the DSL and not make a request through the Collection directly, we await the response of the request
2/ The, we dispatched it to the cozy-client's redux store. 
3/ Redux store is updated with this new document (without making any stack request) 
4/ Cozy-Client replays all the queries, in memory aka without requesting the stack (through http) , already made and add this new document to the result 
5/ All your concerned queries by this document will be updated 
6/ You'll have your application up to date without making any http request 

But why does it works in Drive, because it works and there is no http request to get the result? Good question @zatteo! 

There is something called `realtime` on Cozy : https://docs.cozy.io/en/cozy-stack/realtime/ . When something happens in CouchDB, the stack send a message to say the change. 

Cozy-Drive is linked to this realtime by using the `RealtimeQueries` component: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/RealTimeQueries.jsx like what we do in "2/". 

This component listen for a message within the websocket, and when there is something, dispatch the result like what we do in "2/". 

So why do the works twice? 

As I said, the mechanism from the redux store directly only works when we do not bypass it. We bypass it by calling the collection directly. We need to work on making accessible those method through the DSL: https://github.com/cozy/cozy-client/pull/650 

But we also need to rely on the realtime for drive: There is a lot of application that can manipulate files on your cozy (konnectors, but also other apps). And in that case, cozy-client can't handle that for you since this is not the same instance. 

And I do not want to exclusively rely on the realtime because it adds latence (since stack sends a message). Message can be missed and so one.

Crash--: So we shoudl wait the response, normalize it and return the normalized data. 

Like that, you can also change the return type to `FileDocument`

Crash-- (approved):

PR #1259, opened 2022-11-03
fix: Remove isReferencedByAlbum

zatteo:
BREAKING CHANGE: Since the introduction of the photos' clustering, each photo is at least referenced by an automatic album. Therefore, this method which was used to know if a file was referenced by an album, will always return true, does not make sense anymore because it is not what we want to test.

Instead, we now want to see if a file is not referenced by an album that has not been created automatically. This implementation has been done in Drive directly in eb76c831b3deabd0857b9472451830831df5fb3a

Linked to https://github.com/cozy/cozy-drive/pull/2739

### DIFF
--- a/docs/api/cozy-client/interfaces/models.file.FileUploadOptions.md
+++ b/docs/api/cozy-client/interfaces/models.file.FileUploadOptions.md
@@ -14,7 +14,7 @@ Erase / rename
 
 *Defined in*
 
-[packages/cozy-client/src/models/file.js:466](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/file.js#L466)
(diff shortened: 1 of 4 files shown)


Crash-- (approved):
Its so enjoyable to remove code! <3

paultranvan:
> Therefore, this method which was used to know if a file was referenced by an album, will always return true, does not make sense anymore because it is not what we want to test.

It is not entirely correct: an *image* file will always return true. But a PDF, for instance, will never return true. That being said, it is clear that this deleted method was used for photos files and can be deleted if it is not used anymore in Drive/Photos.

paultranvan (approved):
Very nice initiative to take the time to delete dead code :clap:

PR #1261, opened 2022-11-04
fix: Properly display objects in warning logs

paultranvan:
We stringify the logged objects to avoid having `[object Object]`


Crash--:
You got [Object object] on Node env, right? On the browser you got something you can unfold I think, and I find that very useful instead of a string. If you can confirm that, maybe you should test the env we're in?

paultranvan:
> You got [Object object] on Node env, right? 

Yes, it's actually from logs reported by @clochix in ecolyo services. I'll try to test if it is correctly displays in browser

paultranvan:
@Crash-- Actually, this is the same in browser
image

Crash-- (approved):

PR #1262, opened 2022-11-09
feat: Method fetchPermissionsByLink normalizes permissions

zatteo:
It is now more coherent because we expose only methods returning normalized permissions.


Crash-- (approved):
Personally, I would have tagged this commit as fix but it's a nit.

PR #1264, opened 2022-11-10
fix: Query definition replay with only a partialFilter

zatteo:
When we had a query with no selector but a partial index,
we were missing the sift evaluation.


paultranvan:
I think it's worth adding a description in the commit about the why: in this case it's because if we had a query with no selector but a partial index, we were missing the sift evaluation

Crash-- (approved):

PR #1266, opened 2022-11-15
fix: Async login

Crash--:
We added to the login, the fetch of the cozy capabilities here https://github.com/cozy/cozy-client/pull/1227

Since then, we had error (not failure!) in tests saying we were missing async call...

So this commit fix this issue by:
- waiting the loginPromise if needed in the fromOldClient() method
- mocking the collection.get method to have the test passing. (we can't not mock stackclient by passing stackclient as the 2nd arg of this method because we'll bypass the all test because of that, so I had to mock it globally...)

### DIFF
--- a/docs/api/cozy-client/classes/CozyClient.md
+++ b/docs/api/cozy-client/classes/CozyClient.md
@@ -83,7 +83,7 @@ Cozy-Client will automatically call `this.login()` if provided with a token and 
 
 *Defined in*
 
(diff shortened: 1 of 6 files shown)


Crash--:
@Ldoppea can you review it please?

Ldoppea (approved):

Ldoppea:
Done, sorry i didn't see this PR

PR #1271, opened 2022-11-24
fix(ContactsCollection): Check selector to prevent undefined

cballevre:
This problem appears when you only have partialIndex so the selector which is where return undefined. This problem was found into : https://github.com/cozy/cozy-contacts/pull/898

### DIFF
--- a/packages/cozy-stack-client/src/ContactsCollection.spec.js
+++ b/packages/cozy-stack-client/src/ContactsCollection.spec.js
@@ -36,6 +36,16 @@ describe('ContactsCollection', () => {
     jest.spyOn(console, 'warn').mockImplementation(() => {})
   })
 
+  it('call find route if there is no selector', async () => {
+    const { col, stackClient } = setup()
+    await col.find()
+    expect(stackClient.fetchJSON).toHaveBeenCalledWith(
+      'POST',
+      '/data/io.cozy.contacts/_find',
+      expect.any(Object)
+    )
+  })
+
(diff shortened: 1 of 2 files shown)


Crash-- (commented):
Can you add a test please?

packages/cozy-stack-client/src/ContactsCollection.spec.js
```diff
@@ -36,6 +36,16 @@ describe('ContactsCollection', () => {
     jest.spyOn(console, 'warn').mockImplementation(() => {})
   })
 
+  it('should handle undefined selector', async () => {
+    const { col, stackClient } = setup()
+    await col.find()
+    expect(stackClient.fetchJSON).toHaveBeenCalledWith(
+      'POST',
+      '/data/io.cozy.contacts/_find',
+      expect.objectContaining({ selector: undefined })
```
paultranvan, line 45: Does that mean we actually send `selector: undefined` to the stack, or the selector attribute is not there?

cballevre: not if we have an partial index with combine with selector. I didn't define partialFilter for this test so it's return undefined

Crash--: The test should be : 
`it('call find route if there is no selector')`

and we should just test that `_find` has been called. We don't need to test the object

Crash--:
ping @cballevre

Crash-- (approved):

PR #1274, opened 2022-11-29
Add expiration helpers

PolariTOON:
This extracts and reimplements many helpers from `mespapiers` that were initially implemented in cozy/mespapiers#329 and also add new ones.
The rationale behind this change is to be able to use these helpers in the viewer (`cozy-ui` and thus also `mespapiers` and `cozy-drive`) and also `cozy-mespapiers-lib`, instead of only be able to use them for `mespapiers` services.


packages/cozy-client/src/models/file.js
```diff
@@ -1,3 +1,5 @@
+import sub from 'date-fns/sub'
+import add from 'date-fns/add'
```
Crash--: `date-fns` is only a devDep currently. I don't know why... Also, devDep is currently 2.X while in our front app we use 1.X version. 

But since you added it to `fileModel` we should absolutely do something about that point. 
At first, it should be added to the `dep` section. But doing that we'll increase the size of the cozy-client lib. 

Maybe it should be a peerDep?

packages/cozy-client/src/models/file.js
```diff
@@ -16,6 +18,29 @@ export const ALBUMS_DOCTYPE = 'io.cozy.photos.albums'
...
+        label: 'residence_permit',
+        expirationDateAttribute: 'expirationDate'
+      },
+      {
+        label: 'personal_sporting_licence',
+        expirationDateAttribute: 'referencedDate'
+      }
+    ]
+  }
+}
```
Crash--: Can you create a file called `paper.js` or something like that and put inside all the stuff related to `paper`? 

Models are very domain oriented. Since here we want to deal with `paper` even if `paper` is a subset of `file` I don't think it should belong to `file` but have its own "domain"

packages/cozy-client/src/models/file.js
```diff
@@ -1,5 +1,4 @@
-import sub from 'date-fns/sub'
-import add from 'date-fns/add'
+import { add, sub } from 'date-fns'
```
Crash--, line 1: We will not import from the root for this kind of lib. Too much stuff can comes with that kind of import.

packages/cozy-client/src/models/file.js
```diff
@@ -622,6 +622,68 @@ export const fetchBlobFileById = async (client, fileId) => {
...
+ */
+const isExpiringPersonalSportingLicense = file => {
+  const label = get(file, 'metadata.qualification.label', null)
+  const referencedDate = get(file, 'metadata.referencedDate', null)
+  const created_at = get(file, 'created_at', null)
+  if (label === 'personal_sporting_licence' && (referencedDate || created_at)) {
+    return true
+  }
+  return false
+}
```
Crash--, line 648: We should start coding differently, something more declarative instead of having those imperative `if`... We should only rely on a configuration file and the code will do the rest: 


```
const config = [
'frenchNationalCard: {
'isExpiring': true, 
'expiringCondition': file => .... 
}
]
```

Like that there is no need to update the code if we want to add a new paper, change expiring condition or else. We just have to edit this config. If we think even further this config can be a json file served directly by the cozy. Like that even the PO will be able to edit by him/herself the content.

packages/cozy-client/src/models/file.js
```diff
@@ -622,6 +622,68 @@ export const fetchBlobFileById = async (client, fileId) => {
...
+/**
+ * @param {IOCozyFile} file - io.cozy.files document
+ * @returns {boolean}
+ */
+const isExpiringFrenchNationalIdCard = file => {
+  const label = get(file, 'metadata.qualification.label', null)
+  const country = get(file, 'metadata.qualification.country', null)
+  const expirationDate = get(file, 'metadata.expirationDate', null)
+  if (
+    label === 'national_id_card' &&
```
Crash--, line 614: We should not put this check at the same level as the `expirationDate`. 

Either we make sure that we can't pass wrong data here or we can do something like: 
```
const isExpiringFrenchNationalIdCard = file => {
  const label = get(file, 'metadata.qualification.label', null)
  const country = get(file, 'metadata.qualification.country', null)
  const expirationDate = get(file, 'metadata.expirationDate', null)
if(label !== 'national_id_card'){
logger.log(`You called isExpiringFrenchNationalIdCard for a non national id card. File: ` JSON.stringify(file)`)
}
if(country && country !== 'fr'){logger.log... }

File: packages/cozy-client/src/models/file.js [hunk 6]
```diff
@@ -665,50 +665,55 @@ export const isExpiring = file => {
...
+export const computeExpirationDate = file => {
+  if (isExpiringFrenchNationalIdCard(file) || isExpiringResidencePermit(file)) {
+    const expirationDate = get(file, 'metadata.expirationDate', null)
+    return new Date(expirationDate)
   }
+  if (isExpiringPersonalSportingLicense(file)) {
+    const referencedDate = get(file, 'metadata.referencedDate', null)
+    const created_at = get(file, 'created_at', null)
+    return add(new Date(referencedDate ?? created_at), {
+      days: 365
```
Crash--, line 680: this 365 should be added to the `config` we talked early.

packages/cozy-client/src/models/file.js
```diff
@@ -717,3 +717,20 @@ export const computeExpirationNoticeDate = file => {
...
+    return 'https://www.service-public.fr/particuliers/vosdroits/N358'
+  }
+  if (isExpiringResidencePermit(file)) {
+    return 'https://www.service-public.fr/particuliers/vosdroits/N110'
+  }
+  if (isExpiringPersonalSportingLicense(file)) {
+    return 'https://www.service-public.fr/particuliers/vosdroits/F1029'
+  }
+  return null
+}
```
Crash--, line 736: Same here. Those code should be inside the config.

packages/cozy-client/src/models/file.js
PolariTOON: The goal of this function is not to test if a given national id card is french with an expiration date, but instead if a given file is a national id card that is French and with an expiration date.
Being able to be called on any file is on purpose.

packages/cozy-client/src/models/file.js
PolariTOON: UI has `date-fns@^1.28.5`, Drive has `date-fns@1.30.1`, MesPapiers (lib) has `date-fns@2.29.3` and MesPapiers (app) has `date-fns@2.23.0`.
So we have different major versions in the libs / apps that will consume this helpers.
Thus i don't think we can currently set it as a peer dependency.
I would be inclined to put in `dep` dependency in `cozy-client`, otherwise we would have to update the version in each consumer (perhaps we should).
Anyway, i will move the helpers in a `paper.js` file as you suggested.

packages/cozy-client/src/models/file.js
PolariTOON: The original approach was indeed to declare everything in a JSON (`paperDefinitions` in the first commit, that is later removed), but we discussed extensively between devs and decided to move away from a declarative coding since the conditions to determine if a given paper expires and how to compute their expiration and notice dates are already very different. 

For each relevant paper, the expiring condition is different because we use different attributes and not as many. Moreover the way we compute dates is also already different, with different fallback behavior and default delays. Making something purely declarative and thus generic would need many parameters and add noise in my opinion. As you first suggested it is easier to define dedicated methods for each paper anyway, so at this point i don't really see the benefit to define either a very complex JSON config that won't really be extensible in practice or methods instead of functions.

Note that we also discussed this with the PO and confirmed that were was no need for them to be able to edit themselves a config.

JF-Cozy: > we discussed extensively between devs and decided to move away from a declarative coding

Attention à ne pas mélanger tous les sujets. A mon sens lors de nos échanges nous n'avons pas exclu spécifiquement le fait du faire du déclaratif de manière absolu. Là dessus, par défaut, perso je suis pas fermé.

Maintenant le fait que ce soit facile ou pas, et pertinent ou non (ce que tu évoques) est un autre point.

packages/cozy-client/src/models/file.js
PolariTOON: Isolated as a `PERSONAL_SPORTING_LICENCE_PERIOD_DAYS` constant for now, i'll try to see how i can integrate that to a config file.

packages/cozy-client/src/models/file.js
Crash--: This is not clear. You should add it to the description.

packages/cozy-client/src/models/file.js
Crash--: Nice let's have a duplicate package...

>  I would be inclined to put in dep dependency in cozy-client, otherwise we would have to update the version in each consumer (perhaps we should).

Yep, but we can't easily. See https://github.com/cozy/cozy-banks/pull/2522#issuecomment-1274636844 . We also have applications that remove duplicate package. It'll be a nightmare but I imagine this is what we need to do...

JF-Cozy:
exemple de déclaratif : 

```
import { national, passport } from 'identityLabels'
import { wedding } from 'familyLabels'

const papersExpirationConfig = {
  national_id_card: {
    expiration: {
      condition: file => national.condition(file),
      date: file => national.date(file)
    },
    notification: {
      sendingDate: () => {}
    },
    alert: {
      renewalLink: () => {}
    }
  },
  passport: {
    expiration: {
      condition: file => national.condition(file),
      date: file => passport.date(file)
    },
    notification: {
      sendingDate: () => {}
    },
    alert: {
      renewalLink: () => {}
    }
  }
}

export const cIsExpiring = file => {}
export const cExpirationDate = file => {
  if cIsExpiring(file)
  else
}
export const cSendingDate = file => {
  if cIsExpiring(file)
  else
}
export const cRenewalLink = file => {
  if cIsExpiring(file)
  else
}
```

packages/cozy-client/src/models/file.js
JF-Cozy: on peut simplement
- passer date-fns en dep sur la dernière version
- mettre à jour la lib
- faire l'import profond

Crash--: Dans ce cas là, si la version de `date-fns` importe peu, alors une peerDep fera l'affaire.

PolariTOON:
Some of the changes since the latest review:
- added `date-fns` to the direct dependencies and updated it
- moved the helpers to `paper.js`
- removed the link for the personal sporting license
- fixed how the country is retrieved
- removed the default notice period for the french national id card and the residence permit
- used optional chaining instead of lodash
- added missing types definition for attributes used in these helpers

I also opened an alternate PR with a final refactor at #1281 that could land instead of this one, or separately afterwards.
I think this is ready for another round of reviews. @Crash--, could you take another look ?

packages/cozy-client/src/models/file.js
Crash--: Je ré-ouvre ce point. Si on n'a pas besoin d'une version spécifique de date-fns alors peerDep. Sinon on a tous les bundlemon de toutes les app qui vont péter 🙏

Crash--: > Sans trop que je comprenne pourquoi, l'erreur qui m'a motivé à faire ce fix a disparu. J'ai quand même mis à jour date-fns et l'ai ajouté en dep. Aucun changement dans la lib entre les deux versions qui nous concerne apparemment.

Je ne faisais que reprendre ça pour dire d'ajouter en peerDep. Si on a besoin de add et de sub alors mettons en dep. Et laissons bundlemon nous rappeler à l'ordre.

packages/cozy-client/src/models/paper.js
```diff
@@ -0,0 +1,97 @@
+import add from 'date-fns/add'
+import sub from 'date-fns/sub'
+
+import { IOCozyFile } from '../types'
```
Crash--, line 4: Since you use it only for the typing, you should remove this JS import and replace it by: 
```
@typedef {import("../types").IOCozyFile} IOCozyFile
```

packages/cozy-client/src/models/paper.js
```diff
@@ -0,0 +1,97 @@
...
+    return new Date(file.metadata[dateLabel])
+  }
+
+  return null
+}
+
+/**
+ * @param {IOCozyFile} file - An CozyFile
+ * @param {string} dateLabel - Label of date
+ * @returns {Date} Notice date
```
Crash--: same

packages/cozy-client/src/models/paper.js
```diff
@@ -1,5 +1,6 @@
 import add from 'date-fns/add'
 import sub from 'date-fns/sub'
+import get from 'lodash/get'
```
Crash--: We avoid to use `lodash/get` now and use [Nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)

packages/cozy-client/src/types.js
```diff
@@ -170,6 +171,11 @@ import { QueryDefinition } from './queries/dsl'
  * @property {object} [schema] - the schema used by prosemirror (with notes and marks serialized as arrays to preserve the order).
  * @property {string} [title] - the initial title of the note (that will also be used for the file name)
  * @property {number} [version] - Number of a note
+ * @property {Qualification} [qualification] - Qualification of the file
+ * @property {string} [country] - Country of the paper
+ * @property {string} [expirationDate] - Expiration date of the paper
```
Crash--, line 176: String? Are-you sure?

packages/cozy-client/src/types.js
```diff
@@ -170,6 +171,11 @@ import { QueryDefinition } from './queries/dsl'
...
+ * @property {string} [referencedDate] - Reference date of the paper
```
Crash--, line 177: Same here?

Crash--:
Can you squash few commits please? I think it's not useful here to have the refactor between `lodash.get` and `?` for instance. 

I think having 3 commits can be great: 
- one to add date-fns as dep 
- one to add all the expiration stuff 
- one to add the missing jsdoc

packages/cozy-client/src/models/file.js
Merkur39: > est-ce que Yarn installe une peerDep s'il ne la trouve nulle part ?

Non, avec Yarn il faut passer par la commande `npx check-peer-dependencies --yarn --install` pour maj ou installer les perrDeps manquantes.

> Sur quoi on part ?
nit: Si c'est les fonctions les plus récentes, alors on pourrait être en peerDeps >= 2.9.0, et ainsi éviter les feux d'artifices ?

PS: Désolé d'intervenir si tardivement, cette PR était passée sous mon radar...

Crash--: Non pas de peerDep si on a besoin de date-fns en 2.9. On a plein d'app qui sont en 1.X sur date-fns et j'ai pas envie de me retrouver avec un BC à gérer là-dessus. 

Donc `dep`. Et le jour ou toutes les app seront passées à date-fns >= 2.9 alors on pourra regarder quoi faire. 


(ps: vu que Banks a de la dedup de package sur date-fns et que ça ne fait pas attention à la version, ça va être fun... J'espère qu'on ne passera pas dans ces méthodes depuis banks)

packages/cozy-client/src/models/paper.js
PolariTOON: Same² :smile:

packages/cozy-client/src/types.js
PolariTOON: The expiration date is currently set by the user in an date field in Mes Papiers and stored as is in the metadata. So yes this is a string.

packages/cozy-client/src/types.js
PolariTOON: The reference date is also provided by the user as a string.

PolariTOON:
I added a description for each helper to clarify what they do, and imported `IOCozyFile` and `Qualification` as JSDOC types instead of these "ghost" variables. I squashed the whole thing into 3 commits to eliminate the intermediate refactors as @Crash-- suggested. Can you take one (final?) look?

packages/cozy-client/src/models/file.js
PolariTOON: Du coup ça reste en `dep` pour l'instant.

packages/cozy-client/src/models/file.js
PolariTOON: I added a description to each function that should hopefully clarify that.

packages/cozy-client/src/models/paper.js
PolariTOON: Done. Did the same for `Qualification`

packages/cozy-client/package.json
```diff
@@ -51,7 +52,7 @@
     "cozy-flags": "2.10.2",
     "cozy-logger": "1.7.0",
     "cozy-ui": "48.0.0",
-    "date-fns": "2.27.0",
+    "date-fns": "2.29.3",
```
Crash--, line 55: nit: Vu qu'on l'ajoute en dep, aucune raison de la garder en devDep

packages/cozy-client/src/models/paper.js
```diff
@@ -0,0 +1,177 @@
+import add from 'date-fns/add'
+import sub from 'date-fns/sub'
+
+/**
+ * @typedef {import("../types").IOCozyFile} IOCozyFile
+ */
```
Crash--, line 6: <3

Crash-- (approved):

PR #1298, opened 2023-01-11
[WIP] feat: Remove the export from models/index

Crash--:
In order to enhance the cozy-client tree shaking,
we should avoind importing * and exporting.

We had an issue with this PR
https://github.com/cozy/cozy-client/pull/1274

because this PR adds date-fns v2 as dep, but even
if the consuming app doesn't use the concerned model it was imported.

BREAKING CHANGE: You can not use
import { models } from cozy-client
...
models.files.func()

You have to change it to:
import { func } from cozy-client/dist/models/files


Crash--:
still need to work on this subject to have a better understanding of the three shaking mechanism.

PR #1300, opened 2023-01-13
fix: Do not modify selector with lodash merge

paultranvan:
Lodash merge modify the source object to apply the merge
on it.
That was causing errors on queries without `.indexfields` but with
selector and a partial filter.
In such situation, the final selector is a merge between the actual
selector and the partial index.
After the first query, the selector became the merge one.
Then, the query is replayed when the index is missing, this time with
a merged selector.
As the indexed fields were not specified, and we are using the
selector fields to guess them, the resulting index was wrong, as
partially based on the partial filter.

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -533,7 +533,7 @@ The returned documents are paginated by the stack.
     // We need to pass the partialFilter in the selector, otherwise CouchDB might
     // fallback on another index if it does not exist yet
     const mergedSelector = partialFilter
-      ? merge(selector, partialFilter)
+      ? merge({ ...selector }, partialFilter)
       : selector
 
     return {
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -533,7 +533,7 @@ The returned documents are paginated by the stack.
     // We need to pass the partialFilter in the selector, otherwise CouchDB might
     // fallback on another index if it does not exist yet
     const mergedSelector = partialFilter
-      ? merge(selector, partialFilter)
+      ? merge({ ...selector }, partialFilter)
```
Crash--, line 536: can we add a test for that? Having a test will help us... next time to remove the need of lodash without fearing any regression on this kind of stuff

paultranvan: Fair enough, I'll add it

paultranvan: 16ddffb6e3751bd150e0df93cf992e5a2ac0315d
Note the merge itself was already tested

Crash-- (approved):

PR #1305, opened 2023-01-23
fix(useQuery): Enabled option must throw an error

Merkur39:
If a non boolean value passed to it

One approach, among others but which seems to me the most appropriate


Merkur39:
Best implementation here https://github.com/cozy/cozy-client/pull/1306

PR #1306, opened 2023-01-24
fix: Throw error if enabled is not a boolean

Crash--:
We had an issue on `mespapiers-lib` because mespapiers was sending an empty string instead of a boolean. And since this commit https://github.com/cozy/cozy-client/commit/c5b602256f3e4cd747734fa213500191eeb2e3c9 , we don't "resolve" the query definition. And since the `enabled` was not false, the query was executed https://github.com/cozy/cozy-client/blob/c5b602256f3e4cd747734fa213500191eeb2e3c9/packages/cozy-client/src/hooks/useQuery.js#L69-L72 (see https://github.com/cozy/cozy-libs/pull/1983/files) 

This is an alternative implementation of https://github.com/cozy/cozy-client/pull/1305 

I prefer the current approach because `query()` is the central point of cozy-client. If we use `<Query>` `client.query` `useQuery` `useQueryAll` , all these method pass to `CozyClient.query()`. It makes more sense to check it there.

### DIFF
--- a/docs/api/cozy-client/classes/CozyClient.md
+++ b/docs/api/cozy-client/classes/CozyClient.md
@@ -83,7 +83,7 @@ Cozy-Client will automatically call `this.login()` if provided with a token and 
 
 *Defined in*
 
-[packages/cozy-client/src/CozyClient.js:1625](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L1625)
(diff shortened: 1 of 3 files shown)


Merkur39:
Je comprends et préfère effectivement cette approche, plus globale 👍

Merkur39 (approved):

PR #1307, opened 2023-01-31
feat: Add sourceAccount & sourceAccountIdentifier to createFile

Crash--:
Konnectors need to pass the sourceAccount & sourceAccountIdentifier to POST /files/ in order to let the stack set the right cozyMetadata.


LucsT (approved):

paultranvan (approved):
Thanks! It would indeed be nice to deal with empty params :+1:

PR #1309, opened 2023-02-03
Update doc metadata

LucsT:
Update doc metadata with new possibility for sourceAccount and source AccountIdentifier

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -48,7 +48,7 @@ Cozy-Client will set:
 * `metadataVersion` <string>: to the last version known to Cozy-Client. It's the version of the schema for the metadata block itself.
 * `doctypeVersion` <string>: if it's provided in the `schema` at Cozy-Client initialization
 * `slug` <string>: Slug of the app
-* `sourceAccount`:
+* `sourceAccount`: <string>: if it's provided in the `appMetadata` at initialization. Note it won't work for files (see below).
 * creation date, app and version
 * updated date, app and version
 
@@ -76,4 +76,25 @@ Cozy-Client will:
 * Add your app and version in the `updatedByApps` list
 
 ⚠️ These values will override any values that your document may already have.
+
+### Specific case of io.cozy.files
+
+The io.cozy.files doctype is protected by cozy-stack, that restricts the attributes that can be saved by the client, including the cozyMetadata.
+
+This is the case for :
+* `cozyMetadata.sourceAccount`
+* `cozyMetadata.sourceAccountIdentifier`
+
+To save those attributes, you can pass them directly to the save method, like this:
+
+```
+const client = new client()
+client.save({
+  _type: 'io.cozy.files',
+  sourceAcount: '12ab3c',
+  sourceAccountIdentifier: 'john@example.org',
+  ...
+})
+```
+
 
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -43,15 +43,15 @@ When you provide this information, cozy-client will automatically update metadat
 
 ### For new documents
 
-Cozy-Client will set: 
+Cozy-Client will set:
 
 * `metadataVersion` <string>: to the last version known to Cozy-Client. It's the version of the schema for the metadata block itself.
 * `doctypeVersion` <string>: if it's provided in the `schema` at Cozy-Client initialization
 * `slug` <string>: Slug of the app
-* `sourceAccount`: 
+* `sourceAccount`:
 * creation date, app and version
 * updated date, app and version
-  
+
 ```
 const client = new CozyClient({
   appMetadata: {


docs/metadata.md
```diff
@@ -77,3 +77,24 @@ Cozy-Client will:
...
+Since cozy-client 34.11.0, you can use a direct attribut in the document to save it.
+
+```
+const client = new client()
+client.save(
+  _type: 'io.cozy.files',
+  sourceAcount: '12ab3c',
+  sourceAccountIdentifier: 'john@example.org',
+  ...
+)
```
Merkur39, line 97: Plutôt quelque chose du genre je pense ?
```suggestion
client.save({
  _type: 'io.cozy.files',
  sourceAcount: '12ab3c',
  sourceAccountIdentifier: 'john@example.org'
})
```

docs/metadata.md
```diff
@@ -77,3 +77,24 @@ Cozy-Client will:
 
 ⚠️ These values will override any values that your document may already have.
 
+### Specific case of io.cozy.files
+
+`io.cozy.files` attribut cozyMetadata is protected by cozy-stack and can't be saved directly.
```
paultranvan, line 82: Reformulation suggestion: 
The `io.cozy.files` doctype is protected by cozy-stack, that restricts the attributes that can be saved by the client, including the `cozyMetadata`.

docs/metadata.md
```diff
@@ -77,3 +77,24 @@ Cozy-Client will:
...
+
+This is the case for :
+* `cozyMetadata.sourceAccount`
+* `cozyMetadata.sourceAccountIdentifier`
+
+Since cozy-client 34.11.0, you can use a direct attribut in the document to save it.
```
paultranvan, line 88: I'm not sure it's worth specifying the version for every change, the doc is supposed to be relevant only for the latest version.

paultranvan: Reformulation suggestion:
To save those attributes, you can pass them directly to the `save` method, like this:

docs/metadata.md
paultranvan: By seeing this, I realize this is troublesome to handle for an update:
```
const doc = await client.query(Q('io.cozy.files').getById('myid'))
const newDoc = updateDocWithStuff(doc)
await client.save(newDoc) 
```

If the `cozyMetadata.sourceAccount` was changed in the process,  the dev must think to pass it directly as a `newDoc.sourceAccount` to update it, and not the received `newDoc.cozyMetadata.sourceAccount`. 
Maybe passing `cozyMetadata.sourceAccount` in the [FileCollection](https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/FileCollection.js#L543) would make more sense.
What do you think @Crash-- ?

docs/metadata.md
```diff
@@ -43,15 +43,15 @@ When you provide this information, cozy-client will automatically update metadat
...
 ### For new documents
 
-Cozy-Client will set: 
+Cozy-Client will set:
 
 * `metadataVersion` <string>: to the last version known to Cozy-Client. It's the version of the schema for the metadata block itself.
 * `doctypeVersion` <string>: if it's provided in the `schema` at Cozy-Client initialization
 * `slug` <string>: Slug of the app
-* `sourceAccount`: 
+* `sourceAccount`:
```
paultranvan, line 51: You probably can remove it, since there is no description and you mention it afterwards

docs/metadata.md
Crash--: You can not update the sourceAccount / sourceAccountIdentifier. Cozy-Stack doesn't allow it. This data is too much sensitive.

docs/metadata.md
LucsT: I'm not sure about it, because you can still pass a sourceAccount via the constructor of a client in the cozy-client source code. (ex: for other doctype)

docs/metadata.md
LucsT: Done

docs/metadata.md
LucsT: Yeah, thank's, I was not sure too. Changed

docs/metadata.md
LucsT: Forgotten brackets added.

For the other point, that's the main reason of this problem, you can't pass cozyMetadata.sourceAccount.
So the second best option to pass sourceAccount directly.

docs/metadata.md
paultranvan: Thus maybe a description like this: 
``` 
`sourceAccount`: <string>: if it's provided in the `appMetadata` at initialization. Note it won't work for files (see below).
```

paultranvan (approved):

paultranvan:
Thanks for the doc improvement!

PR #1312, opened 2023-02-07
docs: Correct login logout event name mistake

zatteo:


Crash-- (approved):
Let's fix the CI and then feel free to merge it ;)

PR #1313, opened 2023-02-09
fix: Do not overwrite mandatory fields in updateInformation

zatteo:
According to the function specification, mandatory fields can not be updated.

So we invert information and mandatoryFields.

### DIFF
--- a/packages/cozy-stack-client/src/OAuthClient.js
+++ b/packages/cozy-stack-client/src/OAuthClient.js
@@ -227,8 +227,8 @@ class OAuthClient extends CozyStackClient {
       softwareID: this.oauthOptions.softwareID
     }
     const data = this.snakeCaseOAuthData({
-      ...mandatoryFields,
-      ...information
+      ...information,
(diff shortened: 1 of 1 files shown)


packages/cozy-stack-client/src/OAuthClient.js
```diff
@@ -227,8 +227,8 @@ class OAuthClient extends CozyStackClient {
       softwareID: this.oauthOptions.softwareID
     }
     const data = this.snakeCaseOAuthData({
-      ...mandatoryFields,
-      ...information
+      ...information,
+      ...mandatoryFields
```
Crash--, line 231: in fact @zatteo I think we misunderstood what's happening. This is not fields that we can't override, this is juste mandatoryFields. So we can override them if wanted, but they need to be there. 

So I think we can close the issue

zatteo: Oh you're right. I close.

PR #1314, opened 2023-02-09
feat: Sanitize fields in manifest lib

doubleface:
And also add the `getIdentifier` method to be able de get the
`sourceAccountIdentifier` whatever format of the fields in the manifest

This code is directly imported from cozy-harvest-lib and will be used by cozy-harvest-lib once these modifications are in cozy-client

- feat: Sanitize fields in manifest lib
- docs: Add ts-check + corresponding fixes


packages/cozy-client/src/manifest.js
```diff
@@ -84,5 +114,112 @@ export function sanitize(manifest) {
...
+ *
+ * @param  {Object} [fields={}] Manifest fields
+ * @returns {Object}            Sanitized manifest fields
+ */
+const sanitizeFields = _flow([
+  removeOldFields,
+  sanitizeIdentifier,
+  sanitizeRequired,
+  sanitizeEncrypted
+])
```
Crash--, line 225: I didn't even know about this file. I think we should move it to /models directory. Can you move it and create a BC for that?

packages/cozy-client/src/manifest.js
```diff
@@ -84,5 +114,112 @@ export function sanitize(manifest) {
...
+    sanitized.fields = sanitizeFields(manifest.fields)
+  }
+
   return sanitized
 }
+
+/**
+ * Ensures that fields has at least one field with the role 'identifier'
+ *
+ * @param  {Object} [fields={}] Manifest fields
```
Crash--, line 127: Can you, at least, create a typedef for fields?

packages/cozy-client/src/manifest.spec.js
```diff
@@ -133,3 +188,262 @@ describe('sanitizeManifest', () => {
...
+      id: {
+        type: 'text'
+      }
+    }
+
+    expect(getIdentifier(fields)).toBe('username')
+  })
+})
+
+describe('sanitizeFields', () => {
```
Crash--, line 220: sanitize ? 

You're not testing `sanitizeFields`. If you want to test it, you need to export it. This is my prefer-ed way instead of having to test the full `sanitize` to expect that `sanitizeFields` is working correctly.

packages/cozy-client/src/manifest.spec.js
```diff
@@ -133,3 +188,262 @@ describe('sanitizeManifest', () => {
...
+          }
+        }
+      }
+    }
+
+    const result = sanitizeManifest(oldManifest)
+    expect(result.fields.advancedFields).toBeUndefined()
+  })
+})
+describe('sanitizeIdentifier', () => {
```
Crash--, line 243: same here

packages/cozy-client/src/manifest.js
```diff
@@ -84,5 +114,112 @@ export function sanitize(manifest) {
...
+    }
+
+  return sanitized
+}
+
+/**
+ * Returns the key for the field having the role=identifier attribute
+ *
+ * @param  {Object} fields Konnector fields
+ * @returns {string}        The key for the identifier field, example 'login'
```
Crash--, line 159: can return undefined if there is no match.

packages/cozy-client/src/manifest.js
doubleface: Done, but this actually not a BC since the manifest lib is called via client.manifest and I fixed index.js and index.node.js. Don't you think ?

Crash--: > Done, but this actually not a BC since the manifest lib is called via client.manifest and I fixed index.js and index.node.js. Don't you think ?

We've a lot of deep import since our tree shaking is not working correctly, so yes I consider it as a BC.

packages/cozy-client/src/manifest.js
```diff
@@ -84,5 +114,112 @@ export function sanitize(manifest) {
...
 }
+
+/**
+ * Ensures that fields has at least one field with the role 'identifier'
+ *
+ * @param  {Object} [fields={}] Manifest fields
+ * @returns {Object}             Sanitized manifest fields
+ */
+export const sanitizeIdentifier = fields => {
+  const sanitized = _cloneDeep(fields)
```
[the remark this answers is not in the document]
doubleface: Replacing cloneDeep with object spread makes the `should not mutate source manifest` test fail

doubleface:
@Crash-- It should be ok now

packages/cozy-client/src/manifest.js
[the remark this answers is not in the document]
paultranvan: The spread only makes a copy for the top-level elements of the object. Any nested object will be still copied by reference, :
not value:
```
const x1 = {a: 1, b: { c: 2}}
const x2 = {...x1}
x2.a = 3
x2.b.c = 42
// x1 = {a: 1, b: { c: 42}}
```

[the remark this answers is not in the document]
Crash--: > The spread only makes a copy for the top-level elements of the object. Any nested object will be still copied by reference, not value

This is the idea of creating a type for that, to be able to know is we're dealing with one level objet or not ;)

packages/cozy-client/src/index.js
```diff
@@ -37,7 +37,7 @@ export { getQueryFromState } from './store'
 export { default as Registry } from './registry'
```
Crash--, line 37: Inside this commit message, you should write how to handle the BC.

Something like: 
```
If you were importing manifest from 'cozy-client/dist/manifest` you have to either import it from the root .... or from the models.

Crash-- (approved):
Nice! thanks. Please take my comment in consideration before merging :)

PR #1315, opened 2023-02-09
feat: Import accounts and triggers features from harvest

doubleface:
These features will be used by harvest and any app willing to create accounts or triggers

I also tried to add proper jsdoc for those doctypes (+io.cozy.konnectors)

- feat: Import accounts and triggers features from harvest
- docs: Add ts-check + corresponding fixes
- docs: Build doc

### DIFF
--- a/packages/cozy-client/src/models/account.js
+++ b/packages/cozy-client/src/models/account.js
@@ -3,26 +3,23 @@ import merge from 'lodash/merge'
 
 import { getHasManyItem, updateHasManyItem } from '../associations/HasMany'
 import { legacyLoginFields, getIdentifier } from './manifest'
-/**
- * @typedef {object} CozyAccount
- */
 
 /**
  * getMutedErrors - Returns the list of errors that have been muted for the given account
  *
- * @param {object} account io.cozy.accounts
+ * @param {import('../types').IOCozyAccount} account io.cozy.accounts
  *
  * @returns {Array} An array of errors with a `type` and `mutedAt` field
  */
-export const getMutedErrors = account => account?.mutedErrors ?? []
+export const getMutedErrors = account => account?.mutedErrors || []
 
 /**
  * muteError - Adds an error to the list of muted errors for the given account
  *
- * @param {CozyAccount} account   io.cozy.accounts
+ * @param {import('../types').IOCozyAccount} account   io.cozy.accounts
  * @param {string} errorType The type of the error to mute
  *
- * @returns {CozyAccount} An updated io.cozy.accounts
+ * @returns {import('../types').IOCozyAccount} An updated io.cozy.accounts
  */
 export const muteError = (account, errorType) => {
   const mutedErrors = getMutedErrors(account)
@@ -39,7 +36,9 @@ const DEFAULT_CONTRACT_SYNC_STATUS = true
 /**
  * Returns whether a contract is synced from account relationship
  *
- * @param  {CozyAccount} account - Cozy account
+ * @param  {import('../types').IOCozyAccount} account - Cozy account
+ * @param  {String} contractId - contract identifier
+ * @returns  {Boolean} synchronisation status
  */
 export const getContractSyncStatusFromAccount = (account, contractId) => {
   const relItem = getHasManyItem(account, 'contracts', contractId)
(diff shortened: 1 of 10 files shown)


packages/cozy-client/src/models/account.js
```diff
@@ -39,7 +36,9 @@ const DEFAULT_CONTRACT_SYNC_STATUS = true
 /**
  * Returns whether a contract is synced from account relationship
  *
- * @param  {CozyAccount} account - Cozy account
+ * @param  {import('../types').IOCozyAccount} account - Cozy account
+ * @param  {String} contractId - contract identifier
+ * @returns  {String} synchronisation status
```
Crash--, line 41: is it really a `string`? It seems that it can return at least a `true` boolean

packages/cozy-client/src/types.js
```diff
@@ -5,18 +5,62 @@ import { QueryDefinition } from './queries/dsl'
...
+ * @property {String} worker - type of worker. Can be "konnector" or "sendmail"
+ * @property {object} message - Parameters to pass to the the worker. For example, when the worker is set to konnector, message contains the related konnector and the related account.
+ * @property {String} [arguments] - Arguments related to the type attribute. For example it's a cron configuration when the type is set to @cron.
+ * @property {TriggerState} [current_state] - state of the last executed jobs related to this trigger
+ * @typedef {CozyClientDocument & TriggersDocument} IOCozyTrigger - An io.cozy.konnectors document
+ */
+
+/**
+ * @typedef {object} TriggerState
+ * @property {String} status - Global status of the trigger (queued, running, done, errored)
```
Crash--, line 52: You can make an enum: 
```
* @property {'queued'|'running'|'done'|'errored'} status - Global status of the trigger 
```

packages/cozy-client/src/models/trigger.js
```diff
@@ -59,7 +59,7 @@ const triggerStates = {
   /**
    * Returns the type of the last error to occur
    *
-   * @param {import('../types').TriggersDoctype} triggerState - trigger object
+   * @param {import('../types').IOCozyTrigger} triggerState - trigger object
```
Crash--, line 62: Are you sur that is a triggerState here? 

The method reads: 
```
getLastErrorType: triggerState =>
    get(triggerState, 'current_state.last_error')
```

But `current_state` is a property of `TriggersDocument`if it was really a triggerState that was passed, then we'll just returned `  get(triggerState, 'last_error')`

Crash--: OK I got it, TS is lost because of lodash `get`. If you change it by `triggerState?.current_state...` you'll see the error. 

Also, I don't know why, but this `current_state` property is not generated in the `types.d.ts` file. I had to change the order inside the types.js to be able to make it works.
```
/**
 * @typedef {object} TriggersDocument
 * @property {String} [_id] - document identifier
 * @property {String} type - type of the trigger. Can be "at", "cron", "event", "every", "in", "webhook", "client"
 * @property {String} worker - type of worker. Can be "konnector" or "sendmail"
 * @property {object} message - Parameters to pass to the the worker. For example, when the worker is set to konnector, message contains the related konnector and the related account.
 * @property {TriggerState} [current_state] - state of the last executed jobs related to this trigger
 * @property {String} [arguments] - Arguments related to the type attribute. For example it's a cron configuration when the type is set to @cron.
 *
 * @typedef {CozyClientDocument & TriggersDocument} IOCozyTrigger - An io.cozy.konnectors document
 */
```

Crash--: (maybe because of `@cron`)

Crash-- (commented):
Nice, thanks! Let's fix few typing issues, but really nice! Thanks

packages/cozy-client/src/models/account.js
doubleface: My mistake, in every case, this is a boolean

packages/cozy-client/src/models/trigger.js
doubleface: I removed all the get() calls in this file and renamed the triggerState variable everywhere since it is actually a trigger document. I then fixed the corresponding jsdoc issues

doubleface:
@Crash-- Should be ok now

Crash-- (approved):
<3

PR #1320, opened 2023-02-21
fix: Do not rely on potentially missing datetime

paultranvan:
Some images such as bitmaps do not have any `metadata` object.
However, we were relying on the existence of a `metadata.datetime` date
in order to build the cursor for fetching HasManyFiles relationship,
which was causing errors in the photos clustering service.

Kudos to the inspector @Crash-- for finding out this bug :clap:

### DIFF
--- a/packages/cozy-client/src/associations/HasManyFiles.js
+++ b/packages/cozy-client/src/associations/HasManyFiles.js
@@ -23,6 +23,17 @@ const newCursor = ([doctype, id, lastDatetime], startDocId) => {
 }
 
 /**
+ * Get the file datetime
+ *
+ * @param  {import('../types').IOCozyFile} file - io.cozy.files document
+ * @returns {string} The file datetime
+ */
+export const getFileDatetime = file => {
+  // Some files do not have any metadata, e.g. bitmap files.
+  return file.metadata?.datetime || file.created_at
+}
+
+/**
  *  This class is only used for photos albums relationships.
  *  Behind the hood, the queries uses a view returning the files sorted
  *  by datetime, with a cursor-based pagination.
@@ -48,8 +59,7 @@ export default class HasManyFiles extends HasMany {
         lastRelationship._type,
         lastRelationship._id
       )
-      // Photos always have a datetime field in metadata
-      const lastDatetime = lastRelDoc.attributes.metadata.datetime
+      const lastDatetime = getFileDatetime(lastRelDoc.attributes)
       // cursor-based pagination
       const cursor = newCursor(
         [this.target._type, this.target._id, lastDatetime],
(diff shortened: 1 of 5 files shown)


packages/cozy-client/types/associations/helpers.d.ts
```diff
@@ -11,6 +11,7 @@ export function isReferencedBy(file: IOCozyFile, referencedBy: Doctype): boolean
 export function isReferencedById(file: IOCozyFile, referencedBy: Doctype, referencedId: string): boolean;
 export function getReferencedBy(file: IOCozyFile, referencedBy: Doctype): Reference[];
 export function getReferencedById(file: IOCozyFile, referencedBy: Doctype, referencedId: string): Reference[];
+export function getFileDatetime(file: IOCozyFile): string;
```
Crash--: Does it belongs really to this commit?

packages/cozy-client/src/associations/HasManyFiles.js
```diff
@@ -7,7 +7,7 @@ import { CozyClientDocument } from '../types'
 import { DOCTYPE_FILES } from '../const'
 import Association from './Association'
 import HasMany from './HasMany'
-import { CouchDBViewCursor, DocId, ViewKey } from '../types'
+import { CouchDBViewCursor, DocId, ViewKey, IOCozyFile } from '../types'
```
Crash--, line 10: If this is just for the jsdoc, you should not import it that way but rather something like: 
```
@param  {import('../types').IOCozyFile} file
```

packages/cozy-client/src/associations/HasManyFiles.js
```diff
@@ -48,8 +59,8 @@ export default class HasManyFiles extends HasMany {
         lastRelationship._type,
         lastRelationship._id
       )
-      // Photos always have a datetime field in metadata
-      const lastDatetime = lastRelDoc.attributes.metadata.datetime
+      console.log('go get dateimte')
```
Crash--, line 62: to remove?

packages/cozy-client/src/associations/HasManyFiles.js
paultranvan: Why is it better? Does it allow to make the import only if the function is actually called from the app?

packages/cozy-client/types/associations/helpers.d.ts
paultranvan: Good catch, I rebased

packages/cozy-client/src/associations/HasManyFiles.js
Crash--: because `import` is used by the transpiler / bundler. If you import it only for documentation, you import it. And the file will be available in the bundle. If you /*@param import('')*/ then the file will not be available in the bundle.

paultranvan: Ok nice ; but don't we exclusively use the types for documentation, i.e. in the JSDocs ? Is there legitimate use for importing types at the beginning of the file ?

Crash--: No.

Crash-- (approved):

PR #1321, opened 2023-02-22
fix: Normalize doc for fetchChanges

Crash--:
Recently, we changed in Banks from
BankTransaction.updateAll() to client.saveAll().

But we had errors in our log saying that we were
calling collection without a doctype
(CozyStackClient.collection() called without a doctype)

In fact, this is because saveAll need to read the
_type for the documents to update them cf:
https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/StackLink.js#L122

Sometime it worked for Banks sometime not. It was due to the fact that we have a condition here:
https://github.com/cozy/cozy-banks/blob/master/src/targets/services/helpers/helpers.js#L19-L45

The fix, is to normalize the document returned by fetchChanges() like that we add _type & id automatically.

We have a fetchChangesRaw() to get the real changes without any edition.

### DIFF
--- a/packages/cozy-stack-client/src/DocumentCollection.js
+++ b/packages/cozy-stack-client/src/DocumentCollection.js
@@ -827,7 +827,10 @@ The returned documents are paginated by the stack.
     if (!options.includeDeleted) {
       docs = docs.filter(doc => !doc._deleted)
     }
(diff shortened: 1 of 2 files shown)


paultranvan (approved):
:+1:

PR #1322, opened 2023-02-22
feat: Add deconstruct redirect link helper

zatteo:
To be used when cozy-stack return redirect link (i.e. 'slug/#/path') for example in the onboarded_redirection
field from settings.


packages/cozy-client/src/helpers.js
```diff
@@ -119,6 +119,24 @@ export const deconstructCozyWebLinkWithSlug = (
...
 
+/**
+ * Deconstruct the given redirect link in order to retrieve slug and hash
+ *
+ * @param {string} redirectLink - redirect link to deconstruct (i.e. 'slug/#/path')
+ * @returns {RedirectLinkData} Deconstructed link
+ */
+
+export const deconstructRedirectLink = redirectLink => {
+  const [slugWithEndingSlash, hash] = redirectLink.split('#')
```
Crash--, line 130: Be careful, you can have : 

`slug/path/#/hash`

zatteo: Oh! Like what? `drive/public/#/folder/abcd` ?

Crash--: Yep you don't know what can be there. 

See https://github.com/cozy/cozy-stack/blob/63782e4e45faed57e3fa5fe0ae5c49480a5f6aaf/model/instance/instance.go#L531-L532

packages/cozy-client/src/helpers.spec.js
```diff
@@ -228,6 +229,53 @@ describe('deconstructWebLink', () => {
...
+        hash: '/long/hash'
+      }
+    ],
+    [
+      'contacts/long/path/#/long/hash',
+      {
+        slug: 'contacts',
+        path: 'long/path/',
+        hash: '/long/hash'
+      }
```
zatteo, line 272: @Crash-- What do you think about the new version ? Is it okay to have a trailing slash at the end of path ?

PR #1324, opened 2023-02-28
feat: Add account.isAccountWithTrigger

doubleface:
To know if a given account is associated to a trigger or not.

### DIFF
--- a/packages/cozy-client/src/models/account.js
+++ b/packages/cozy-client/src/models/account.js
@@ -1,5 +1,6 @@
 // @ts-check
 import merge from 'lodash/merge'
+import { Q } from '../queries/dsl'
 
 import { getHasManyItem, updateHasManyItem } from '../associations/HasMany'
 import { legacyLoginFields, getIdentifier } from './manifest'
@@ -113,4 +114,23 @@ export const buildAccount = (konnector, authData) => {
     identifier: getIdentifier(konnector.fields),
     state: null
   }
+}
+
+/**
+ * Look if the given account has an associated trigger or not.
+ *
+ * @param {import('../CozyClient').default} client - CozyClient instance
+ * @param {import('../types').IOCozyAccount} account - account document
+ * @returns {Promise<Boolean>}
+ */
+export const isAccountWithTrigger = async (client, account) => {
+  const result = await client.query(
+    Q('io.cozy.triggers')
+      .where({
+        'message.account': account._id
+      })
+      .indexFields(['message.account'])
+      .limitBy(1)
+  )
+  return result.data.length > 0
 }
--- a/docs/api/cozy-client/modules/models.account.md
+++ b/docs/api/cozy-client/modules/models.account.md
@@ -27,7 +27,7 @@ io.cozy.accounts attributes
 
 *Defined in*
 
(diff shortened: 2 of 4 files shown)


packages/cozy-client/src/models/account.js
```diff
@@ -1,5 +1,7 @@
 // @ts-check
 import merge from 'lodash/merge'
+import CozyClient from '../CozyClient'
```
Crash--, line 3: If this is just for the JSDoc, don't import it like that. https://github.com/cozy/cozy-client/pull/1320#discussion_r1113034764

packages/cozy-client/src/models/account.js
```diff
@@ -114,3 +116,22 @@ export const buildAccount = (konnector, authData) => {
...
+
+/**
+ * Look if the given account is connected or not. Does a trigger which is referencing this account
+ * exist
+ *
+ * @param {CozyClient} client - CozyClient instance
+ * @param {String} accountId - id of the account
+ * @returns {Promise<Boolean>}
+ */
+export const isAccountConnected = async (client, accountId) => {
```
Crash--, line 128: I'd prefer you to pass the entire account object. Not just the id. 

Cozy-Client's models are about a "functional object", not just a subpart. It's easier to remember that all the methods from models take the object

packages/cozy-client/src/models/account.js
```diff
@@ -114,3 +116,22 @@ export const buildAccount = (konnector, authData) => {
...
+  const result = await client.query(
+    Q('io.cozy.triggers')
+      .where({
+        'message.account': accountId
+      })
+      .indexFields(['message.account'])
```
Crash--, line 134: You can add a limitBy(1). Like that, the query will be more optimized

packages/cozy-client/src/models/account.js
```diff
@@ -114,3 +116,22 @@ export const buildAccount = (konnector, authData) => {
...
+  )
+  return result?.data?.length > 0
```
Crash--, line 136: result can't be empty. data neither. No?

packages/cozy-client/src/models/account.js
```diff
@@ -114,3 +115,23 @@ export const buildAccount = (konnector, authData) => {
...
+
+/**
+ * Look if the given account is connected or not. Does a trigger which is referencing this account
+ * exist
+ *
+ * @param {import('../CozyClient').default} client - CozyClient instance
+ * @param {import('../types').IOCozyAccount} account - account document
+ * @returns {Promise<Boolean>}
+ */
+export const isAccountConnected = async (client, account) => {
```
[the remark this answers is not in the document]
doubleface: But since the the connected/disconnected state is displayed to the user in harvest, I think this has more meaning than isAccountWithTrigger, no ?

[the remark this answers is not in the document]
paultranvan: The meaning of a function should makes sense to the developer, not the user. And for me it's always better to reflect the actual technical meaning, rather than a "high level" meaning, that will be scope-dependent. Furthermore, we may now use this function out of the scope of harvest ;)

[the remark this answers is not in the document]
doubleface: Ok, let's fo for isAccountWithTrigger

PR #1325, opened 2023-02-28
feat: Add getCronFromFrequency to manifest models

doubleface:
This feature is imported from harvest and allows to get the cron string
from a given frequency like 'daily'.

This cron string will configure the trigger associated to the konnector


packages/cozy-client/src/models/cron.js
```diff
@@ -0,0 +1,69 @@
...
+  randomDayTimeFn = randomDayTime
+) =>
+  fromFrequency(konnector.frequency, {
+    ...randomDayTimeFn.apply(
+      null,
+      konnector.time_interval || DEFAULT_TIME_INTERVAL
+    ),
+    dayOfWeek: startDate.getDay(),
+    dayOfMonth: startDate.getDate()
+  })
```
Crash--, line 69: models are related to functional object. `cron` doesn't exist. `cron` is needed for the manifest, so let's move it there?

packages/cozy-client/src/models/daytime.js
```diff
@@ -0,0 +1,37 @@
...
+    throw new Error('randomize function returns invalid hour value')
+
+  return { hours, minutes }
+}
+
+/**
+ * @typedef {object} randomDayTimeResult
+ * @property {number} hours - hours
+ * @property {number} minutes - minutes
+ */
```
Crash--, line 37: same

paultranvan (commented):
You can now create daily, weekly and monthly triggers, with the randomness directly handled by the stack: https://github.com/cozy/cozy-stack/blob/master/docs/jobs.md#daily-syntax
I assume we could eventually directly use this? Although hourly is not supported yet (but should it?)

doubleface:
> You can now create daily, weekly and monthly triggers, with the randomness directly handled by the stack: [cozy/cozy-stack@master/docs/jobs.md#daily-syntax](https://github.com/cozy/cozy-stack/blob/master/docs/jobs.md?rgh-link-date=2023-03-06T14%3A03%3A15Z#daily-syntax)

Oh I did not know that. Let's use it then

doubleface:
> Although hourly is not supported yet (but should it?) @paultranvan 

[I just saw](https://github.com/search?q=org%3Akonnectors%20hourly&type=code) that some some banking konnectors used by self hosted users use the "hourly" frequency (because the konnector is only run for them is their own machine I suppose). Maybe we should add the hourly frequency to the stack, what do you think ?

Crash--:
Let's see https://github.com/cozy/cozy-stack/pull/3835

Crash--:
@doubleface `@hourly` has been added to the stack. But I think we should use it with a lot of precaution. For the two examples you have listed, I think we should move them `@daily`.

Crash--:
ping @doubleface

Crash--:
@doubleface 🙏

Crash--:
@doubleface

Crash--:
@doubleface . If not needed anymore, please close.

PR #1328, opened 2023-03-05
fix: Correctly type nullable return for `useClient()`

Ldoppea:
If no `CozyProvider` is set on the components tree, then `useClient()` would return no `client`

### DIFF
--- a/packages/cozy-client/types/context.d.ts
+++ b/packages/cozy-client/types/context.d.ts
@@ -1,6 +1,6 @@
 export const clientContext: import("react").Context<{
-    /** @type {CozyClient}  */
-    client: CozyClient;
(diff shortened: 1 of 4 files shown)


acezard (approved):

packages/cozy-client/types/context.d.ts
```diff
@@ -1,6 +1,6 @@
 export const clientContext: import("react").Context<{
```
acezard, line 1: those types are created automatically, right? why are they versioned instead of built before deployment? this question is not strictly related to the current PR obviously and doesn't block it

Ldoppea: I'm not sure why this has been done.

PR #1234, opened 2022-08-31
feat: Add .d.ts files generation on cozy-stack-client

Ldoppea:
On our Flagship mobile app I need type declaration for `client.getStackClient()` so I'm trying to add `.d.ts` files generation on cozy-stack-client like it is already set for cozy-client

There are a lot of TS errors, so this PR may take a bit of time. It is draft for now, so you can help if you want to.

PR #1333, opened 2023-03-17
fix: Fix jsdocs in queryAll

LucsT:
The jsdoc concerning queryAll wasn't showing the correct return (Array of docs)

I propose this correction.
What do you think ?

Linked to https://mattermost.cozycloud.cc/cozycloud/pl/jw7dfmxq8fdbdmemjbnuxxjuwa


Crash-- (changes requested):
I would prefer `QueryResult[]` since CozyClientDocument[]` is about document, not JSON API related return. 

There is a lot of work done in this https://github.com/cozy/cozy-client/pull/1234 in order to try to fix types issue.

LucsT:
@Crash-- 
Ok so you want me to add a commit in this PR instead with ```@returns {Promise<QueryResult[]>}``` 
and close this one ?

PR #1334, opened 2023-03-21
fix: Use id instead of path in SettingsCollection.get()

zatteo:
Previously, get was receiveing a path like "instance". This path was used to call a correct route like "/settings/instance" but was also used internally as an id for store stuff. But the id of the document behind "/settings/instance" is
"io.cozy.settings.instance" and not "instance" leading to store issue.

### DIFF
--- a/packages/cozy-stack-client/src/SettingsCollection.js
+++ b/packages/cozy-stack-client/src/SettingsCollection.js
@@ -1,5 +1,6 @@
-import DocumentCollection from './DocumentCollection'
+import DocumentCollection, { normalizeDoc } from './DocumentCollection'
 import logger from './logger'
+import { uri } from './utils'
 
 export const SETTINGS_DOCTYPE = 'io.cozy.settings'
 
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/SettingsCollection.js
```diff
@@ -13,10 +14,26 @@ class SettingsCollection extends DocumentCollection {
...
    * @returns {object} The response from the route
    */
-  async get(path) {
+  async get(id) {
+    let path
+
+    if (id === 'io.cozy.settings.instance') {
+      path = 'instance'
+    } else if (id === 'io.cozy.settings.context') {
+      path = 'context'
```
Crash--, line 26: We've a lot of possible path, and not just `instance` or `context`. https://docs.cozy.io/en/cozy-stack/settings/ 

I'm wondering if we should not do something like: 

```
if(id.includes('io.cozy.settings.')){
path = substring(...)
} else {
// `Deprecated: in next versions of cozy-client, it will not be possible to query settings with an incomplete id

}
```

packages/cozy-stack-client/src/SettingsCollection.js
```diff
@@ -25,6 +41,33 @@ class SettingsCollection extends DocumentCollection {
...
+
+  /**
+   * Updates a settings document
+   *
+   * @param {object} document - Document to update. Do not forget the _id attribute
+   */
+  async update(document) {
+    let resp
+
+    if (document._id === 'io.cozy.settings.instance') {
```
zatteo, line 53: I did not see other PUT which could be used like this in cozy stack documentation.

Crash-- (approved):

cballevre (approved):

PR #1335, opened 2023-03-24
fix: Remove doctypes locales type

cballevre:
These exports were causing errors because typescript cannot find the module "./fr.json" or "./en.json". The
se locales are only used with a deep import into cozy-settings: https://github.com/cozy/cozy-settings/blob/71a21962f4fb48f84561db845bd5f4f5c3e858e8/src/lib/withAllLocales.js#L4. So I decided to remove export and import which block the cozy-client update on cozy-pass-web

image


Crash--:
can you add a breaking change to the commit please?

cballevre:
@Crash-- I added the breaking change 🙂 Can I have a review ? Pls 🙏

PR #1337, opened 2023-03-30
feat: Expose setAppMetadata

Crash--:
In the context of the flagship app, we create a new CozyClient for the clisk. We set the appMetadata we have when we do the instanciation aka slug & version. But in some scenario, we don't have yet the accountIdentifier. So we can't pass it at this moment.

In order to still write this metadata, we now expose a setAppMetadata() to be able to use it later after the creation of the client.

Related PR https://github.com/cozy/cozy-flagship-app/pull/706


paultranvan (approved):
A bit disappointing you had to do a test...

PR #1338, opened 2023-03-30
fix: Handle undefined `pathname` in `generateWebLink()` method

Ldoppea:
The URL javascript object does not support undefined `pathname`. This member should be an empty string (transformed to `/`) or have a value

```js
const url = new URL('https://cozy.io')
console.log(url.pathname) // prints '/'
console.log(url.toString()) // prints 'https://cozy.io/'

url.pathname = undefined
console.log(url.pathname) // prints '/undefined'
console.log(url.toString()) // prints 'https://cozy.io/undefined'

url.pathname = ''
console.log(url.pathname) // prints '/'
console.log(url.toString()) // prints 'https://cozy.io/'

url.pathname = ensureFirstSlash(undefined)
console.log(url.pathname) // prints '/'
console.log(url.toString()) // prints 'https://cozy.io/'
```

By using 'ensureFirstSlash' method we ensure that the caller can pass undefined `pathname` without negative side effect

### DIFF
--- a/packages/cozy-client/src/helpers/urlHelper.js
+++ b/packages/cozy-client/src/helpers/urlHelper.js
@@ -44,7 +44,7 @@ export const generateWebLink = ({
           .split('.')
           .map((x, i) => (i === 0 ? x + '-' + slug : x))
           .join('.')
-  url.pathname = pathname
+  url.pathname = ensureFirstSlash(pathname)
   url.hash = ensureFirstSlash(hash)
 
   for (const [param, value] of searchParams) {
--- a/packages/cozy-client/src/helpers/urlHelper.spec.js
+++ b/packages/cozy-client/src/helpers/urlHelper.spec.js
@@ -40,6 +40,18 @@ describe('generateWebLink', () => {
       })
     ).toEqual(`https://drive.alice.cozy.tools/public/#/files/432432`)
   })
+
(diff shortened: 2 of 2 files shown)


packages/cozy-client/src/helpers/urlHelper.js
```diff
@@ -44,7 +44,7 @@ export const generateWebLink = ({
           .split('.')
           .map((x, i) => (i === 0 ? x + '-' + slug : x))
           .join('.')
-  url.pathname = pathname
+  url.pathname = pathname || ''
```
Crash--, line 47: shouldn't we use `ensureFirstSlash` as just below for the `hash`?

Ldoppea: The result would be the same, `URL` will internally convert `''` to `'/'`

Do you prefer the ensureFirstSlash version?

Crash--: to be consistent with the line just below your code, yep. Also `/` speaks to me for a path while `''`  doesn't.

Ldoppea: Done, I also updated the commit and PR messages

Ldoppea: I'll merge just after I get the CI result

PR #1345, opened 2023-04-19
refactor: Remove JS import for the JSDoc one

Crash--:
In order to avoid warning with some bundlers:
"Require circule"

First step, changed the import from 'types'. We still have work to do when we import class only for the type too. But it ll be for a next time.


paultranvan (approved):
I applause the tedious work! :clap:

PR #1352, opened 2023-04-26
fix: Handle `503 Blocked` response when calling `rootCozyUrl`

Ldoppea:
When calling `rootCozyUrl` we want the method to throw a specific exception if the Cozy is in a "blocked" state

If a `503 Blocked` is returned we conclude that there is a Cozy behind the URL but we cannot check if the URL points to the Cozy's root or to a specific slug. So it is safer to throw an exception that should then be handled by the caller

BREAKING CHANGE: rootCozyUrl() can now throw a BlockedCozyError exception if the target Cozy is blocked

### DIFF
--- a/docs/api/cozy-client/README.md
+++ b/docs/api/cozy-client/README.md
@@ -10,6 +10,7 @@ cozy-client
 ## Classes
 
 *   [Association](classes/Association.md)
+*   [BlockedCozyError](classes/BlockedCozyError.md)
 *   [BulkEditError](classes/BulkEditError.md)
 *   [CozyClient](classes/CozyClient.md)
 *   [CozyLink](classes/CozyLink.md)
@@ -615,7 +616,7 @@ The root Cozy URL
 
 *Defined in*
 
-[packages/cozy-client/src/helpers/urlHelper.js:249](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/helpers/urlHelper.js#L249)
+[packages/cozy-client/src/helpers/urlHelper.js:269](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/helpers/urlHelper.js#L269)
 
 ***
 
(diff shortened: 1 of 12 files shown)


Ldoppea:
> **Warning**
> Blocked cozy will return a 503 even if the slug is present in the URL which would produce the wrong root URL, we are discussing about this behavior, we should find a solution before merging this PR

Crash--:
@Ldoppea is it still a WIP (because there is a unchecked todo)

Ldoppea:
Yes this is still WIP, last Friday we chose to change the way we will handle this error.

Instead of returning a `valid` url if a `503 - Blocked` error is detected, we will throw a custom error containing the `Blocked` info

This is because a locked cozy would also return a `503 - Blocked` on `claude-drive.mycozy.cloud/.well-known/change-password` instead of the usual `401` error.
So with the current PR implementation we would return `claude-drive.mycozy.cloud` as a valid cozy URL (which is not as it contains a slug)

So instead we will throw on `503 - Blocked` error

Ldoppea:
This PR has been reworked to throw an exception in case of `503 - Blocked` cozy

New commit message is

> When calling `rootCozyUrl` we want the method to throw a specific exception if the Cozy is in a "blocked" state
> 
> If a `503 Blocked` is returned we conclude that there is a Cozy behind the URL but we cannot check if the URL points to the Cozy's root or to a specific slug. So it is safer to throw an exception that should then be handled by the caller
> 
> BREAKING CHANGE: rootCozyUrl() can now throw a BlockedCozyError exception if the target Cozy is blocked

Old commit message was

> When calling `rootCozyUrl` we want to retrieve the root URL of a Cozy even if this Cozy is in a "blocked" state
> 
> With previous implementation a blocked Cozy would be considered as non-existing which was incorrect
> 
> Instead it should be considered as existing and valid. The "blocked" state should be handled somewhere else (i.e when initializing the client on caller side)
> 
> BREAKING CHANGE: rootCozyUrl() now returns a valid URL even if the target Cozy is blocked. The "blocked" state should be verified by the caller (i.e. when registering CozyClient)

The PR is now ready for review

PR #1353, opened 2023-05-15
Add `others` theme and `other_admin_doc` qualification

JF-Cozy:
PR d'usage : https://github.com/cozy/cozy-libs/pull/2152


packages/cozy-client/src/models/document/documentTypeData.js
```diff
@@ -170,6 +171,23 @@ const invoiceLabels = [
...
+ */
+const healthTheme = flag('hide.healthTheme.enabled')
+  ? undefined
+  : {
+      id: 'theme4',
+      label: 'health',
+      icon: 'heart',
+      items: buildItemsByLabel(healthLabels)
+    }
+
```
paultranvan, line 190: I feel like the `healthTheme` should always be here, and it's the calling app that should use the flag to decide if it should be included or not. Otherwise, that sounds like are we putting some business logic in a shared lib

packages/cozy-client/src/models/document/documentTypeData.js
```diff
@@ -230,5 +235,11 @@ export const themesList = [
     label: 'invoice',
     icon: 'bill',
     items: buildItemsByLabel(invoiceLabels)
+  },
+  {
+    id: 'theme10',
+    label: 'others',
+    icon: 'dots',
+    items: buildItemsByLabel(othersLabels)
```
paultranvan, line 251: If we create an "other" theme, don't we want to include all the others labels? Like "other_work_document", etc?

packages/cozy-client/src/models/document/documentTypeData.js
JF-Cozy: C'était la question, on voulait éviter de filtrer sur toutes les apps, mais si ça te semble plus pertinent comme ça je peux changer 👍

packages/cozy-client/src/models/document/documentTypeData.js
BenjaminMty: Good legitimate question, but not really, the idea is to place "unclassifiable" documents here and documents like "other_work_document" have an associated theme.

PR #1356, opened 2023-06-13
feat: Add missing type to konnectors document

cballevre:
I realised that the name field was missing from the konnectorsDocument type when I did my [PR](https://github.com/cozy/cozy-libs/pull/2190) on cozy-harvest-lib. 

On the [doctype](https://github.com/cozy/cozy-doctypes/blob/07fa847eac938523e5b60a7a77037b14c56375a0/docs/io.cozy.konnectors.md), there are even more fields, can I add them? Or have they been omitted on purpose?

### DIFF
--- a/packages/cozy-client/src/types.js
+++ b/packages/cozy-client/src/types.js
@@ -30,9 +30,42 @@ import { QueryDefinition } from './queries/dsl'
 
 /**
  * @typedef {object} KonnectorsDocument
- * @property {String} slug - slug of konnector
- * @property {ManifestFields} fields - konnector fields
- * @property {Boolean} clientSide - whether the konnector runs on client or not
+ * See https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.konnectors/
+ *
+ * @property {object} [aggregator] - object containing aggregator data
+ * @property {String[]} [categories] - list of categories (default: ['others'])
+ * @property {Boolean} [clientSide] - whether the konnector runs on client or not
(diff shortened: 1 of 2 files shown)


Crash--:
You can add them :)

cballevre:
@Crash-- I added them can I have a review pls 🙂

packages/cozy-client/src/types.js
```diff
@@ -30,9 +30,42 @@ import { QueryDefinition } from './queries/dsl'
...
  * @typedef {object} KonnectorsDocument
- * @property {String} slug - slug of konnector
- * @property {ManifestFields} fields - konnector fields
- * @property {Boolean} clientSide - whether the konnector runs on client or not
+ * See https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.konnectors/
+ *
+ * @property {object} [aggregator] - object containing aggregator data
+ * @property {Array} [categories] - list of categories (default: ['others'])
+ * @property {Boolean} [clientSide] - whether the konnector runs on client or not
+ * @property {Array} [data_types] - list of data types managed by the connector
```
Crash--, line 38: You can't do that? `Array.<string>` ? Same for categories and other array

Crash-- (approved):

packages/cozy-client/src/types.js
cballevre: You're right after checking JSDoc documentation I've specified the type of each Array by adding `[]` to the type contained in the Array (e.g. String[])

PR #1357, opened 2023-06-14
Add directory metadata support 

zatteo:
Following this PR https://github.com/cozy/cozy-stack/pull/3968, we can now add metadata to a directory. This PR add metadata support to `createDirectory` and `createDirectoryByPath` methods exposed by cozy-client.


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -35,6 +35,7 @@ import logger from './logger'
  * @property {string} dirId - Id of the parent directory.
  * @property {boolean} name - Name of the created directory.
  * @property {boolean} executable - Indicates whether the file will be executable.
+ * @property {object} [metadata] io.cozy.files.metadata to attach to the file
```
Crash--, line 38: nit: I try to talk about file as opposed to folder. And if we're talking about both, in this case I'm trying to talk about io.cozy.files. Since here we are in a DirectoryAttribute, I would have use `folder` instead of file.

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -862,7 +870,7 @@ class FileCollection extends DocumentCollection {
     return this.specialDirectories[path]
   }
 
-  async getDirectoryOrCreate(name, parentDirectory) {
+  async getDirectoryOrCreate(name, parentDirectory, directoryAttributes = {}) {
```
Crash--, line 873: jsdoc ? 🙏 😘

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -891,15 +900,24 @@ class FileCollection extends DocumentCollection {
...
+    for (const [i, part] of parts.entries()) {
+      if (i === parts.length - 1) {
+        parentDir = await this.getDirectoryOrCreate(
+          part,
+          parentDir.data,
+          attributes
+        )
+      } else {
+        parentDir = await this.getDirectoryOrCreate(part, parentDir.data)
+      }
```
Crash--, line 920: Can you explain why you have to make this check and this different call for this one?

paultranvan:
I suppose we need this change to attach new metadata to a folder? If yes, it would be great to document what are those metadata in https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.files_metadata.md :)

packages/cozy-stack-client/src/FileCollection.js
zatteo: It is because we want to add attributes only to the final directory (the "target" directory). 

I updated the js doc to be more explicit.

Crash--: The explanation seems complicated to me, so I find the API hard to understand. I wonder if we need a method that does both at the same time, why not simply have: 
- a method that creates the folder 
- a method that puts the metadata you want on this folder 

Or a method "set attributes on the root of this path" or something like that?

zatteo: **Summary of discussions and work** 

/!\ It is not possible to update the metadata of a folder and it is not possible to add metadata after folder creation. 

So if we want to create a folder _foo/bar/backup_ with attributes on _backup_
- we have a method to create _/foo/bar_ (createDirectoryByPath)
- we have a method to create _backup_ inside _/foo/bar_ with the folder attributes (createDirectory)
=> result is easy to understand and there is not a lot of code inside cozy-client

Implementation differences https://github.com/cozy/cozy-flagship-app/compare/a0b184e5c9d125496e9006026be43702e2a697b8..5c05e7a07a9fc265384ee2d18b2e8a61f40bc6e3

Crash-- (approved):

packages/cozy-stack-client/src/FileCollection.js
Crash--: I know that this is not optimal for now, maybe we should accept to update the metadata of a folder even after its creation?

PR #1360, opened 2023-06-22
fix: Do not call refreshToken several times in parallel

Crash--:
We can have several fetchJSON() in parallel and if the token is old, then we can call several times refreshToken.

This refresh token mechanism update the io.cozy.oauth-client document and can lead to couchdb conflict.

In order to avoid this issue, we store the first promise and use this one for other call. Like that, the refreshToken method is only called once.

I copy / paster the promise-cache file, because I don't want to include a dep from cozy-client inside stack-client. There is already issue talking about that.


packages/cozy-stack-client/src/CozyStackClient.spec.js
```diff
@@ -597,6 +597,41 @@ describe('CozyStackClient', () => {
       const newToken = client.refreshToken()
       await expect(newToken).rejects.toThrow()
     })
+
+    fit(`should NOT try to refresh several times in parallele`, async () => {
```
Ldoppea, line 601:
```suggestion
    it(`should NOT try to refresh several times in parallel`, async () => {
```

Ldoppea (approved):
> I copy / paster the promise-cache file, because I don't want to include a dep from cozy-client inside stack-client. There is already issue talking about that.

If we don't want duplication, we can also move the method to stack-client and keep a raw import/export in cozy-client for backward compatibility

packages/cozy-stack-client/src/CozyStackClient.spec.js
Ldoppea, line 601: nit: in order to fully test this method we may want to test that 2 consecutive calls correctly call the refreshToken 2 times

packages/cozy-stack-client/src/CozyStackClient.spec.js
```diff
@@ -597,6 +597,41 @@ describe('CozyStackClient', () => {
...
+          client.fetchJSON('GET', '/test'),
+          client.fetchJSON('GET', '/test')
+        ])
+      } catch (e) {
+        expect(e).toBeInstanceOf(FetchError)
+        expect(client.fetch).toHaveBeenCalledTimes(2)
+
+        expect(client.refreshToken).toHaveBeenCalledTimes(1)
+        const token = client.getAccessToken()
+        expect(token).toEqual(FAKE_INIT_OPTIONS.token)
```
nono, line 632: I would move those 3 lines outside of the `catch`, to be sure to not have a green test only because we don't raise an exception on ` await Promise.all` (lines 622-625).

packages/cozy-stack-client/src/CozyStackClient.spec.js
Crash--: it's @nono fault! I was late on a meeting, I forgot to remove the `fit` 🤣

Crash--: You mean call the `this._promiseCache.exec` 2 times, right?

Crash--: Done

packages/cozy-stack-client/src/CozyStackClient.spec.js
Crash--: done

nono (approved):

PR #1361, opened 2023-06-22
feat: Add hasPassword helper

cballevre:
This helper is intended to harmonise the way we check whether the user has a password

### DIFF
--- a/packages/cozy-client/src/models/bitwarden.js
+++ b/packages/cozy-client/src/models/bitwarden.js
@@ -1,0 +1,25 @@
+import { Q } from '../queries/dsl'
+
+/**
+ * Checks the value of the extension_installed attribute
+ *
+ * @param {import("../CozyClient").default} client - The CozyClient instance
+ * @returns {Promise<boolean>} - Returns the value of the extension_installed attribute
+ */
+export const hasExtensionInstalledAttribute = async client => {
+  try {
+    const {
+      data: { extension_installed }
+    } = await client.fetchQueryAndGetFromState({
+      definition: Q('io.cozy.settings').getById('io.cozy.settings.bitwarden'),
+      options: {
+        as: 'io.cozy.settings/io.cozy.settings.bitwarden',
+        singleDocData: true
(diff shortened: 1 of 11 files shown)


packages/cozy-client/src/models/bitwarden.js
```diff
@@ -0,0 +1,26 @@
+import { Q } from '../queries/dsl'
+import CozyClient from '../CozyClient'
```
Crash--, line 2: If this file is only imported for the JSDoc, then you should remove it and use the jsdoc import that way: 

```
   * @param {import("../types").CozyClientDocument} document - Document to query
```

for instance.

packages/cozy-client/src/models/capability.js
```diff
@@ -0,0 +1,21 @@
+import CozyClient from '../CozyClient'
```
Crash--, line 1: same here

packages/cozy-client/src/models/bitwarden.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Checks the value of the extension_installed attribute
+ *
+ * @param {CozyClient} client - The CozyClient instance
+ * @returns {Promise<boolean>} - Returns the value of the extension_installed attribute
+ */
+export const hasExtensionInstalledAttribute = async client => {
```
Crash--, line 10: why the `attribute` stuff?

Crash-- (approved):

packages/cozy-client/src/models/bitwarden.js
cballevre: I've updated all the imports in the JSDOC for CozyClient

packages/cozy-client/src/models/bitwarden.js
cballevre: We have added `attribute` to avoid confusion between hasPassword and hasPasswordDefined and we have followed the same convention for hasExtensionInstalled

PR #1362, opened 2023-06-29
Refactor/lodash

Crash--:
The idea was to reduce the size when importing dsl file. (see https://github.com/konnectors/libs/pull/952) 

I didn't success to remove the dep to findKey ATM. Will open another PR for that.

### DIFF
--- a/docs/api/cozy-client/README.md
+++ b/docs/api/cozy-client/README.md
@@ -55,7 +55,7 @@ cozy-client
 
 *Defined in*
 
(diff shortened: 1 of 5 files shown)


paultranvan:
For posterity: https://github.com/lodash/lodash/blob/npm/isArray.js

paultranvan (approved):
Nice! Do we have an idea of the lodash impact on the bundle size?

PR #1363, opened 2023-07-03
fix(SettingsCollection): Normalize a document spreads attributes to root

cballevre:
Realtime provides an object with attributes at the root, whereas in query result, attributes are a property in themselves. When cozy-client tries to merge them, only the root attributes are updated. When normalising, I now spread attributes at the root so that I can access to a data that is always up to date.property. I decided to flatten the attributes property to have a single data source.

### DIFF
--- a/packages/cozy-stack-client/src/SettingsCollection.js
+++ b/packages/cozy-stack-client/src/SettingsCollection.js
@@ -5,6 +5,20 @@ import { uri } from './utils'
 export const SETTINGS_DOCTYPE = 'io.cozy.settings'
 
 /**
+ * Normalizing a document for SettingsCollection context
+ *
+ * @param {object} doc - Document to normalize
+ * @returns {object} normalized document
+ */
+export const normalizeSettings = doc => {
+  const normDoc = normalizeDoc(doc, SETTINGS_DOCTYPE)
+  return {
+    ...normDoc,
+    ...normDoc.attributes
+  }
+}
+
+/**
  * Implements `DocumentCollection` API to interact with the /settings endpoint of the stack
  */
 class SettingsCollection extends DocumentCollection {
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/SettingsCollection.js
```diff
@@ -50,10 +64,7 @@ class SettingsCollection extends DocumentCollection {
 
     const resp = await this.stackClient.fetchJSON('GET', `/settings/${path}`)
     return {
-      data: DocumentCollection.normalizeDoctypeJsonApi(SETTINGS_DOCTYPE)(
-        { id: `/settings/${path}`, ...resp.data },
-        resp
-      )
+      data: normalizeSettings({ id: `/settings/${path}`, ...resp.data })
```
paultranvan, line 67: I feel like we should improve [normalizeDoctypeJsonApi](https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/DocumentCollection.js#L85) to deal with the `attributes` rather than having to re-implement this normalize function for almost every collection. @Crash-- what do you think?

cballevre: The idea was to have a breaking change with a limited impact so that we could move in gradually. But I agree with your view too

paultranvan (approved):
Approving to avoid blocking this ; the discussion can continue afterwards here, or with an issue :)

Crash--:
Je ne sais pas si c'est le bon plan de faire un BC pour juste settings et donc avoir partout ailleurs accès à `attributes` mais pas ici. 

Dans cette PR, je ferais plutôt le fait de dupliquer, pour que ce soit comme sur les autres doctypes. Et par contre, on pourra ensuite faire un BC global pour dégager ça partout.

Crash-- (approved):

PR #1366, opened 2023-07-07
feat: Add diskUsage helper

cballevre:
These helpers originally come from cozy-settings (cf: [code](https://github.com/cozy/cozy-settings/blob/master/src/lib/makeDiskInfos.ts)), I imported them to share the code with cozy-drive

### DIFF
--- a/docs/api/cozy-client/interfaces/models.instance.DiskInfos.md
+++ b/docs/api/cozy-client/interfaces/models.instance.DiskInfos.md
@@ -1,0 +1,41 @@
+[cozy-client](../README.md) / [models](../modules/models.md) / [instance](../modules/models.instance.md) / DiskInfos
+
+# Interface: DiskInfos<>
+
+[models](../modules/models.md).[instance](../modules/models.instance.md).DiskInfos
+
+## Properties
+
+### humanDiskQuota
+
+• **humanDiskQuota**: `string`
+
+Space used in GB rounded
+
+*Defined in*
+
+[packages/cozy-client/src/models/instance.js:121](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/instance.js#L121)
+
+***
+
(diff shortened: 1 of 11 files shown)


packages/cozy-client/src/models/diskUsage.js
```diff
@@ -0,0 +1,65 @@
...
+  const { diskQuota, diskUsage, percentUsage } = computeDiskInfos(
+    +usage,
+    quota ? +quota : undefined
+  )
+  return {
+    humanDiskQuota: formatDecimals(diskQuota),
+    humanDiskUsage: formatDecimals(diskUsage),
+    percentUsage: Math.round(percentUsage).toString()
+  }
+}
```
Crash--, line 65: Models should be related to a domain. It's not a domain by its own. I think it is related to the `instance` model, so this code should be moved there.

cballevre: I moved my code to the `instance` model so it related to a domain 🙂

packages/cozy-client/types/models/diskUsage.d.ts
```diff
@@ -0,0 +1,29 @@
...
+    humanDiskQuota: string;
+    /**
+     * - Maximum space available in GB rounded
+     */
+    humanDiskUsage: string;
+    /**
+     * - Usage percent of the disk rounded
+     */
+    percentUsage: string;
+};
```
Crash--, line 29: To remove?

cballevre: Why? 🤔 I'm currently using it in drive show my banner. Do you think it's the responsibility of each app to make this calculation according to their needs?

cballevre: Fixed 🔧

Crash-- (approved):

PR #1371, opened 2023-07-17
fix(HasManyFiles): Update after removing a relation

cballevre:
The interface was not updated after deleting an image because only the mutation causes React rendering. The data was updated afterwards, which is why I changed the order of execution

Before we `save` the target after remove the relationships which led to a React re-render. This mechanism has been removed because it was duplicated (cf: [PR](https://github.com/cozy/cozy-client/pull/1048))


paultranvan (approved):
Nice catch :+1:

PR #1376, opened 2023-08-01
feat: Add AppsRegistryCollection

Merkur39:
Using the "registry" source via the `io.cozy.apps` doctype is discouraged as it may skew the `io.cozy.apps` entry of the `cozy->documents` store.

For example, if we request data from the registry, the application will be present in `cozy->documents->io.cozy.apps`, which will falsify the return of the `isInstalled()` helper.

That's why we add `AppsRegistryCollection`, in order to separate the 2 sources in 2 different store entries `cozy.documents['io.cozy.apps']` & `cozy.documents['io.cozy.apps_registry']`


Crash-- (approved):
Nice ;)

PR #1379, opened 2023-08-30
feat: Do not throw when an app or konnector is not installed

Crash--:
Instead, we return data: null

BREAKING CHANGE: before this change AppCollection.getById() and KonnectorCollection.getById() were throwing an error if the stack responded a 404 Application not installed. Now it returns a `data: null`.

### DIFF
--- a/packages/cozy-stack-client/src/AppCollection.js
+++ b/packages/cozy-stack-client/src/AppCollection.js
@@ -60,7 +60,8 @@ class AppCollection extends DocumentCollection {
           this.stackClient,
           `${this.endpoint}${encodeURIComponent(id)}`,
           {
-            normalize: data => normalizeApp(data, this.doctype)
+            normalize: data => normalizeApp(data, this.doctype),
+            dataForNotFound: null
           }
         ),
       registry: () => this.stackClient.fetchJSON('GET', registryEndpoint + id)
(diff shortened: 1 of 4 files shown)


packages/cozy-stack-client/src/AppCollection.js
```diff
@@ -60,7 +60,8 @@ class AppCollection extends DocumentCollection {
           this.stackClient,
           `${this.endpoint}${encodeURIComponent(id)}`,
           {
-            normalize: data => normalizeApp(data, this.doctype)
+            normalize: data => normalizeApp(data, this.doctype),
+            dataForNotFound: null
```
Merkur39, line 64: Any specific reason to add this option?

Merkur39 (approved):
Great! good job ❤️

packages/cozy-stack-client/src/AppCollection.js
Crash--: At first, I wanted to return an empty array for data as the same we do when we `/_find` nothing. But after that, I saw that `Collection.get()` returned a `null` result. 

In fact, I still don't be confortable with the PR since : 
- `FilesCollection.get()` use `statById` which call directly `/files/${id}`. This route returns the same errors as /apps or /konnectors. So it should still throw after this PR. But there is no test that check that behavior. 
- I don't know if we should throw or not in fact in that case instead of returning a null value. I think that it can depends on the context : 
   - If I know I should have the file or the app, if a not found is raised, I should throw. 
   - If I know that it a possibility that I don't have the app installed, then I should not throw?

PR #1380, opened 2023-08-31
feat: Add instance data in loadInstanceOptionsFromStack

zatteo:
We need `locale` attribute on flagship app.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -1676,7 +1676,7 @@ instantiation of the client.`
   /**
    * loadInstanceOptionsFromStack - Loads the instance options from cozy-stack and exposes it through getInstanceOptions
    *
-   * For now only retrieving capabilities is supported
+   * This method is not iso with loadInstanceOptionsFromDOM for now.
    *
    * @returns {Promise<void>}
    */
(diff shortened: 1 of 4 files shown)


paultranvan (approved):
The doc needs to be updated, but looks good!

packages/cozy-client/src/CozyClient.js
```diff
@@ -1685,8 +1685,14 @@ instantiation of the client.`
       Q('io.cozy.settings').getById('io.cozy.settings.capabilities')
     )
 
+    const { data: instanceData } = await this.query(
```
Crash--, line 1688: Should be getOrFechFromState (I don’t know the name exactly) and you should add a fetchPolicy

zatteo: Updated. It should be always fresh so no fetchPolicy no?

paultranvan: @zatteo I think using `fetchQueryAndGetFromState` as proposed by @Crash-- is only useful if we add a fetchPolicy. That being said, I'm not sure we want this kind of logic at this level, I don't think I like the idea of a forced fetchPolicy that could lead to unexpected results from the caller.
In any case, the behaviour should be the same for the other query just above on io.cozy.settings.capabilities.

zatteo: So :
- no fetchPolicy because we need it fresh
- same method for the two calls : decided to stay on query for the moment

PR #1384, opened 2023-09-19
feat: Add parenthesis option to generateNewFileNameOnConflict

zatteo:
For photo backup feature, we often have name issues with underscore renaming. For example, an image named IMG_768.JPG will be renamed IMG_769.JPG instead of IMG_768_1.JPG.

We choose to implement a new version of generateNewFileNameOnConflict more adapted to pictures :
- avoid any problems with _X which are regular on pictures by using (X)
- return a more readable name : IMG_768 (1).JPG is more readable than IMG_768_1.JPG


zatteo:
New version https://github.com/cozy/cozy-client/pull/1385

We do not want to introduce a new way of generated conflict name.

PR #1385, opened 2023-09-19
feat: Allow to pass a name format regex to avoid bad renaming

zatteo:
Sometimes, files are named like they already had a conflict like IMG_001.JPG. It is a normal name but in this case, previous implementation would consider _001 as a conflict suffix and would return IMG_2.JPG whereas we expect IMG_001_1.JPG.

By allowing to pass a name format regex, we can tell the new implementation that the name is the original one and it should not try to increment a number even if it corresponds to the conflict format.


Crash-- (approved):
Nice!

PR #1386, opened 2023-09-20
feat: Update cozy-ui from 48.0.0 to 93.1.1

cballevre:
This migration is necessary as part of the version upgrade of cozy-ui on cozy-drive to keep the DevTools compatible


packages/cozy-client/package.json
Crash--, line 1: You should not put a space after CHANGE. I think it'll not work.

Crash--: Is it needed only if the app use CozyDevTools? If yes, maybe we should add that to the commit message

packages/cozy-client/package.json
```diff
@@ -52,6 +52,7 @@
     "cozy-flags": "2.10.2",
     "cozy-logger": "1.7.0",
     "cozy-ui": "93.1.1",
+    "cozy-intent": "1.17.3",
```
Crash--, line 55: same here for the commit message

Crash-- (approved):
small things to change, but LGTM. Thanks.

packages/cozy-client/package.json
cballevre: Thx I did it the french way 🙁

PR #1387, opened 2023-09-20
Feat: Add `all` & `getAll` methods to `AppsRegistryCollection`

Merkur39:
For an initial need in the `cozy-banks` app, we need to add the `all` & `getAll` methods to the `AppsRegistryCollection` collection.


Crash--:
@Merkur39 Sorry, I missed that one. If still needed, can you rebase it?

PR #1388, opened 2023-09-25
feat: Add useMutation hook

cballevre:


Crash--:
Looks good! Maybe we can add a chapter on the doc. Maybe here https://github.com/cozy/cozy-client/blob/master/docs/getting-started.md#mutate-the-data ? I know that we don't talk about useQuery here https://github.com/cozy/cozy-client/blob/master/docs/getting-started.md#fetch-data but maybe we should. Or here ? https://github.com/cozy/cozy-client/blob/master/docs/react-integration.md 

I don't know, but we should add it somewhere ;)

cballevre:
@Crash-- I've added documentation in https://github.com/cozy/cozy-client/blob/master/docs/react-integration.md that I think is most relevant

Crash-- (approved):

PR #1390, opened 2023-09-27
feat: Now allow to update a trigger with cozy-client

doubleface:
To use it :

```javascript
await client.save({
  _id: 'triggerid',
  _rev: 'triggerrev',
  message: {
    konnector: 'konnectorslug',
    account: 'accountid',
    folder_to_save: 'folderid'
  }  
})
```

Only 'arguments' and message attributes are allowed to be modified other attributes will trigger an error.

### DIFF
--- a/packages/cozy-stack-client/src/TriggerCollection.js
+++ b/packages/cozy-stack-client/src/TriggerCollection.js
@@ -171,8 +171,47 @@ class TriggerCollection extends DocumentCollection {
     }
   }
 
-  async update() {
-    throw new Error('update() method is not available for triggers')
+  /**
+   * Updates a Trigger document. Only updatable attributes plus _id are allowed.
+   *
+   * @param  {object}  trigger Trigger's attributes to update + id
+   * @returns {object}  Stack response, containing resulting trigger document under `data` attribute.
+   */
+  async update(trigger) {
+    for (const key in trigger) {
+      if (
+        ![
+          '_id',
+          '_rev',
+          '_type',
+          'arguments',
+          'message',
+          'cozyMetadata'
+        ].includes(key)
+      ) {
+        throw new Error(
+          `TriggerCollection.update only works for 'arguments', and 'message' attributes.`
+        )
+      }
+    }
+
+    const attributes = {
+      ...(trigger.arguments ? { arguments: trigger.arguments } : {}),
+      ...(trigger.message ? { message: trigger.message } : {})
+    }
+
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/TriggerCollection.js
```diff
@@ -171,8 +171,38 @@ class TriggerCollection extends DocumentCollection {
...
+   * @returns {object}  Stack response, containing resulting trigger document under `data` attribute.
+   */
+  async update(trigger) {
+    for (const key in trigger) {
+      if (!['_id', 'arguments', 'message'].includes(key)) {
+        throw new Error(
+          `TriggerCollection.update only works for 'arguments', and 'message' attributes.`
+        )
+      }
+    }
```
Crash--, line 187: you should not use `update`. You should use `save` in order to not bypass cozy-client links.

packages/cozy-stack-client/src/TriggerCollection.spec.js
```diff
@@ -332,6 +351,70 @@ describe('TriggerCollection', () => {
...
+
+    beforeEach(() => {
+      stackClient.fetchJSON.mockResolvedValue(UPDATE_RESPONSE_FIXTURE)
+    })
+    afterEach(() => {
+      jest.resetAllMocks()
+    })
+
+    it('should call the right route', async () => {
+      await collection.update({
```
Crash--, line 365: you should not use `collection.update`, you should use client.save(). Can you update the tests and the PR please?

doubleface: @Crash-- Since the code and the tests are in cozy-stack-client package, shouldn't I use the collection in the unit tests, and the `collection.save()` here ?

Crash--: Oh my bad. You're right, when unit testing, you can call directly the collection.update(). And it should be named update, not save https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/StackLink.js#L126 (I think you didn't see the deletion of my first comment. Next time I'll update it) 

However, in the PR description, you should not mention collection.update(), you should mention client.save(). This is the way you should use it within the caller anyway.

doubleface: Ok I understand now

paultranvan (approved):

Crash-- (approved):

PR #1394, opened 2023-10-03
fix: Remove the trailing slash for the URI

Crash--:
When navigating on the Flagship app, we can pass into the `securiseUri` method with  window['cozy']?.isSecureProtocol === true.

Result: we call to the toString() method, and then we have a trailing slash.

And since this uri is used when called by client.options.uri some apps can concat that with other string without checking if there is a trailing slash or not.

Apps should not do that, but since we have two different behavior between the flagship and the no flagship app, we fix that here.


Merkur39 (approved):
You will indeed have to be more careful with URL construction in apps. 👍

Ldoppea (approved):

PR #1397, opened 2023-10-05
feat: Add geographic methods

paultranvan:
This adds methods to compute latitude and longitude deltas from a distance, which is typically useful to be able to query close coordinates.
We also add a centroid computation in a sphere, that is uesful to determine the center location of a set of geo points.

### DIFF
--- a/packages/cozy-client/src/models/geo.js
+++ b/packages/cozy-client/src/models/geo.js
@@ -1,3 +1,4 @@
+// @ts-check
 const EARTH_RADIUS_M = 6378137
 const EARTH_CIRCUMFERENCE_M = 40075000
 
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/models/geo.js
```diff
@@ -16,6 +19,15 @@ const degreesToRadians = degrees => {
...
 }
 
+const radiansToDegrees = radians => {
+  return (radians * 180) / Math.PI
+}
+
+const roundToNDecimals = (value, n) => {
+  const multiplier = Math.pow(10, n)
+  return Math.round(value * multiplier) / multiplier
+}
```
Crash--, line 29: let's add the jsdoc for both of the new methods! (also you can add // ts-check on at the top)

Crash-- (approved):

packages/cozy-client/src/models/geo.js
paultranvan: Fair enough, I didn't do it as it was small internal methods

paultranvan: [c7d07ce](https://github.com/cozy/cozy-client/pull/1397/commits/c7d07ce4d8ab58d6ac6964615121a763f48e2818)

PR #1398, opened 2023-10-06
fix: Add trigger message details documentation

doubleface:


doubleface:
@Crash-- @paultranvan @Merkur39 Can I have a review for this please ?

packages/cozy-client/src/types.js
```diff
@@ -75,6 +75,10 @@ import { QueryDefinition } from './queries/dsl'
  * @property {String} type - type of the trigger. Can be "at", "cron", "event", "every", "in", "webhook", "client"
  * @property {String} worker - type of worker. Can be "konnector" or "sendmail"
  * @property {object} message - Parameters to pass to the the worker. For example, when the worker is set to konnector, message contains the related konnector and the related account.
+ * @property {String} [message.account] - Id of the account associated to this trigger, if any
+ * @property {String} [message.konnector] - Slug of the konnector associated to this trigger, if any
+ * @property {String} [message.folder_to_save] - Id of the destination folder of the konnector associated to this trigger, if any
```
Crash--, line 80: Let's use something like this: 
```
* @property {IOCozyAccount['id']} [message.account] - Id of the account associated to this trigger, if any
* @property {IOCozyKonnector['slug']} [message.konnector] - Slug of the konnector associated to this trigger, if any
* @property {IOCozyFolder['path']} [message.folder_to_save] - Id of the destination folder of the konnector associated to this trigger, if any
```

Like that, we add the "relation" between data.

doubleface: Ok I see.
I just fixed the message.folder_to_save attribute which is actually the id of the folder and the the path.

Crash--: OK cool 👍

Crash--:
Last thing... If you tag a commit `docs` then we'll not publish a new version on npm. 

Since you are editing the types and that types are used everywhere, you can tag it as a `feat` or `fix` as you want :)

Crash-- (approved):

PR #1402, opened 2023-10-13
test: Add a test with a partialIndex

Crash--:
We didn't have a hight unit test for partial index and I think it's important to have one because we rely more or more on partialIndexes


packages/cozy-client/src/store/queries.spec.js
```diff
@@ -208,7 +208,23 @@ describe('queries reducer', () => {
       )
       expect(state).toMatchSnapshot()
     })
-
+    it('should update a query with a partiel index', () => {
```
paultranvan, line 211: still partial ;)

packages/cozy-client/src/store/__snapshots__/queries.spec.js.snap
```diff
@@ -663,3 +663,79 @@ Object {
   },
 }
 `;
+
+exports[`queries reducer updates should update a query with a partiel index 1`] = `
```
paultranvan, line 667: s/partiel/partial ;)

paultranvan (approved):
Good idea :+1:

PR #1405, opened 2023-10-18
fix: Inconsistency in Flagship flag

cballevre:
A small mistake 😟 By convention, we use hyphen instead of underscore


Crash-- (approved):

cballevre:
thx @Crash-- 🙏

PR #1409, opened 2023-10-26
feat: Add countries model

Merkur39:
For OCR needs on MesPapiers

### DIFF
--- a/packages/cozy-client/src/models/countries.spec.js
+++ b/packages/cozy-client/src/models/countries.spec.js
@@ -1,0 +1,18 @@
+import { checkCountryCode } from './countries'
+
+describe('checkCountryCode', () => {
+  it('should return true for valid country codes', () => {
+    expect(checkCountryCode('FR')).toBe(true)
+    expect(checkCountryCode('fr')).toBe(true)
+    expect(checkCountryCode('FRA')).toBe(true)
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/models/countries.spec.js
```diff
@@ -0,0 +1,16 @@
...
+describe('checkCountryCode', () => {
+  it('should return true for valid country codes', () => {
+    expect(checkCountryCode('FR')).toBe(true)
+    expect(checkCountryCode('US')).toBe(true)
+    expect(checkCountryCode('FRA')).toBe(true)
+    expect(checkCountryCode('USA')).toBe(true)
+  })
+
+  it('should return false for invalid country codes', () => {
+    expect(checkCountryCode('fr')).toBe(false)
```
paultranvan, line 12: `fr` seems pretty legit to me,  I don't think this should be considered as invalid. We might force capitalization in `checkCountryCode` imo

paultranvan (approved):

packages/cozy-client/src/models/countries.spec.js
Merkur39: I made the modification, the function is no longer case sensitive 👍

paultranvan (approved):

PR #1411, opened 2023-10-27
feat: Apply data enhancement to realtime results for specific doctypes

cballevre:
This enrichment closes the gap between the results of queries from the cozy-stack and realtime. In fact, realtime returns the document directly from the database, unlike the stack, which can add computed fields. Without them, it can lead to instable behaviour in applications. Some applications fix some issue by reimplementing the RealTimeQueries component.

This commit aims to provide unifed api to avoid duplicated code. It can also be used as a reference when we want to close the gap. 

**Applications reimplementation :**
- cozy-settings : [SettingsRealTimeQueries](https://github.com/cozy/cozy-settings/blob/master/src/components/SettingsRealTimeQueries.jsx)
- cozy-drive : [FilesRealTimeQueries](https://github.com/cozy/cozy-drive/blob/master/src/components/FilesRealTimeQueries.jsx)

**Note :**
A function similar to `ensureFilePath` exists in the models, but I chose to implement a new one to avoid dependencies with the models because we want to extract them from cozy-client

### DIFF
--- a/packages/cozy-client/src/RealTimeQueries.jsx
+++ b/packages/cozy-client/src/RealTimeQueries.jsx
@@ -5,13 +5,14 @@ import {
   dispatchDelete,
   dispatchUpdate
 } from './store/realtime'
+import { ensureFilePath } from './helpers/realtime'
 
 /**
  * Component that subscribes to a doctype changes and keep the
  * internal store updated.
  *
- * @param  {object} options - Options
- * @param  {import("./types").Doctype} options.doctype - The doctype to watch
+ * @param {object} options - Options
+ * @param {import("./types").Doctype} options.doctype - The doctype to watch
  * @returns {null} The component does not display anything.
  */
 const RealTimeQueries = ({ doctype }) => {
@@ -26,16 +27,27 @@ const RealTimeQueries = ({ doctype }) => {
       )
     }
 
+    let options = {}
+    if (doctype === 'io.cozy.files') {
+      options.enhanceDocFn = ensureFilePath
+    }
+
+    const handleCreated = data => {
+      dispatchCreate(client, doctype, data, options)
+    }
+
+    const handleUpdated = data => {
+      dispatchUpdate(client, doctype, data, options)
+    }
+
+    const handleDeleted = data => {
+      dispatchDelete(client, doctype, data, options)
+    }
+
     const subscribe = async () => {
-      await realtime.subscribe('created', doctype, data =>
-        dispatchCreate(client, doctype, data)
-      )
-      await realtime.subscribe('updated', doctype, data =>
-        dispatchUpdate(client, doctype, data)
-      )
-      await realtime.subscribe('deleted', doctype, data =>
-        dispatchDelete(client, doctype, data)
-      )
+      await realtime.subscribe('created', doctype, handleCreated)
+      await realtime.subscribe('updated', doctype, handleUpdated)
+      await realtime.subscribe('deleted', doctype, handleDeleted)
     }
     subscribe()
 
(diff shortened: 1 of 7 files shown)


packages/cozy-client/src/RealTimeQueries.jsx
```diff
@@ -55,11 +57,12 @@ const dispatchChange = (
...
  *
- * @param  {object} options - Options
- * @param  {import("./types").Doctype} options.doctype - The doctype to watch
+ * @param {object} options - Options
+ * @param {import("./types").Doctype} options.doctype - The doctype to watch
+ * @param {Function} options.customNormalization - Function to call before dispatching the change
  * @returns {null} The component does not display anything.
  */
-const RealTimeQueries = ({ doctype }) => {
+const RealTimeQueries = ({ doctype, customNormalization }) => {
```
Crash--, line 65: Nice! Maybe we should have the ability to customize the "normalization" by event (create / update / delete). 

Also, I would have done that outside of the `normalization` step, because here it's not a normalization anymore, it's a "business logical" before dispatching the result. 

I didn't have a good naming yet, but : 
`customComputeDocumentFromRealtime` can be more appropriated? 

In fact, we already have this `computeDocumentFromRealtime` without knowing it... : 
```
{ ...couchDBDoc, _deleted: true },
```

So this one cas ben the default on deleteEvent.

packages/cozy-client/src/helpers/realtime.js
```diff
@@ -0,0 +1,39 @@
+import { Q } from '../queries/dsl'
+import CozyClient from '../CozyClient'
```
paultranvan, line 2: nit: Prefer absolute path rather than relative :wink:

packages/cozy-client/src/RealTimeQueries.jsx
```diff
@@ -26,16 +27,34 @@ const RealTimeQueries = ({ doctype }) => {
...
+      if (doctype === 'io.cozy.files') {
+        options.computeUpdatedDoc = ensureFilePath
+      }
+      dispatchUpdate(client, doctype, data, options)
+    }
+
+    const handleDeleted = data => {
+      let options = {}
+      if (doctype === 'io.cozy.files') {
+        options.computeDeletedDoc = ensureFilePath
```
paultranvan, line 49: It's a detail, but I'm not sure it's worth naming differently each method with "delete", "create", "update"

packages/cozy-client/src/store/realtime.js
```diff
@@ -54,9 +47,23 @@ const dispatchChange = (
  * @param {object} client - CozyClient instance
  * @param {import("../types").Doctype} doctype - Doctype of the document to create
  * @param {import("../types").CouchDBDocument} couchDBDoc - Document to create
+ * @param {object} [options] Options
+ * @param {function} [options.computeCreatedDoc] Optional function to transform the normalized document before dispatch
```
paultranvan, line 51: `computeCreatedDoc` sounds ok, but maybe a `applyAttributeComputation` would be clearer: it better suggests that it is a function, you can reuse it whether it is a create/update/delete, and it explicit mention it is used for attribute computation.

packages/cozy-client/src/RealTimeQueries.jsx
cballevre: There's no advantage in doing this in-house now 👍

packages/cozy-client/src/store/realtime.js
cballevre: Good point 👍 I used the term enhance to mean the improvement of an existing object. @paultranvan does it seems more clearer ?

paultranvan (approved):
It's good to me, thanks @cballevre !

PR #1418, opened 2023-11-08
feat: Add useInstanceInfo hook

Ldoppea:
This hook can be use to retrieve instance info like context, uuid, disk usage etc

This is the same as the one found in cozy-settings and cozy-client projects. Those files may be adapted in the future to use this new method from cozy-client instead

Related links:
- https://github.com/cozy/cozy-settings/blob/8c6684be41f6c8e404bee5a7844e102fc697cc26/src/hooks/useInstanceInfo.js
- https://github.com/cozy/cozy-drive/blob/855dedfdbe1b7ad82d5573e83d10c23f719be7d9/src/hooks/useInstanceInfo.jsx


cballevre:
Great sharing 🎉 , I'll be happy to migrate to Drive and Settings once I've merged. However, I think we should migrate the model instance at the same time so that we don't have to use the data anymore ([ex](https://github.com/cozy/cozy-client/blob/81e6523fab6fc927e6526d880910608124168059/packages/cozy-client/src/models/instance.js#L75))

zatteo:
I think the most important work is to add this hook in cozy-client (you did it) and use it in other apps.

I do not see the _necessity_ to remove the .data so I would not do it. But I do not see any counter-argument. So if you want, okkkk letssssgo

Ldoppea:
I reverted to the old structure so changes will be retro-compatibles as the new structure was a too small improvement compared to the actions needed to make it compatible

cballevre (approved):

PR #1419, opened 2023-11-09
feat: Supports clisk for getKonnector

Crash--:
Add an isKonnector method that check if its a clisk or a server side konnector.

Use this isKonnector in getKonnector since we want to get the konnectors, we don't want the implementation details (aka we don't care if its a clisk or not).


Merkur39 (approved):
👍

PR #1420, opened 2023-11-10
feat: Improve flag management into cozy client

cballevre:
This PR is an enhancement to the existing flag panel. The goal behind was to have a simpler way to edit flag especially json flag. For now, I reload the page so that the new value is taken into account. In the future I would like to remove this mechanism 🙂

home alice cozy localhost_8080_


cballevre:
@Crash-- the CI is green now can I merge this PR ?

Crash-- (approved):

PR #1426, opened 2023-11-22
feat: Improve uploadFileWithConflictStrategy

Merkur39:
Following the development of @zatteo on the improvement of the "generateNewFileNameOnConflict" function (#1425),
we can take advantage of it to also offer the option in the "uploadFileWithConflictStrategy" function which uses the latter.


Crash-- (approved):

zatteo (approved):

zatteo:
Nice! I did not think about this part!

PR #1428, opened 2023-11-27
docs: Add return type to models.file.splitFilename

zatteo:
I want to reduce [cozy-client.d.ts](https://github.com/cozy/cozy-flagship-app/blob/master/src/%40types/cozy-client.d.ts) size in cozy-flagship-app.

I am currently trying to understand how it works and how to do it correctly. Here is a first very simple but working type.


Crash-- (approved):
Nice! 

Note that you have tagged the commit as `docs` so it'll not trigger a new version.

PR #1430, opened 2023-12-05
fix(AppsRegistryCollection): Make sure apps always have an id

cballevre:
Applications from the registry have no id. For cozy-client needs, we add one based on the source in the manifest. However, some applications do not have a source, for example banking konnector. This commit ensures that there is an id based on the slug to avoid errors.

### DIFF
--- a/packages/cozy-stack-client/src/AppsRegistryCollection.js
+++ b/packages/cozy-stack-client/src/AppsRegistryCollection.js
@@ -1,13 +1,29 @@
-import { transformRegistryFormatToStackFormat } from 'cozy-client/dist/registry'
-
 import DocumentCollection, { normalizeDoc } from './DocumentCollection'
-import { FetchError } from './errors'
 
 export const APPS_REGISTRY_DOCTYPE = 'io.cozy.apps_registry'
 
-export const normalizeAppFromRegistry = (app, doctype) => {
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/AppsRegistryCollection.js
```diff
@@ -1,13 +1,22 @@
 import { transformRegistryFormatToStackFormat } from 'cozy-client/dist/registry'
 
 import DocumentCollection, { normalizeDoc } from './DocumentCollection'
-import { FetchError } from './errors'
 
 export const APPS_REGISTRY_DOCTYPE = 'io.cozy.apps_registry'
 
 export const normalizeAppFromRegistry = (app, doctype) => {
+  // The stack returns some app without an id, so we use the slug as id.
+  // Sometime we use latest_version.manifest.source as an id.
```
paultranvan, line 9: That means there is 2 levels of normalization for the registry? Shouldn't we only use `normalizeAppFromRegistry`?

paultranvan (approved):

packages/cozy-stack-client/src/AppsRegistryCollection.js
cballevre: After a discussion with @Crash--, we came to the same conclusion: this new version has only one normalization left

PR #1431, opened 2023-12-12
feat: Add createDemoClient to help with demo

cballevre:
This new function will create a client like createMockClient but without the jest mock. It will be useful for demo in documentation because isn't defined in there context. Two usecases are in storybook and in the cozy-ui documentation.

**Remaining work:**
- ~~make unit tests~~

### DIFF
--- a/packages/cozy-client/src/mock.js
+++ b/packages/cozy-client/src/mock.js
@@ -1,3 +1,5 @@
+import merge from 'lodash/merge'
+
 import CozyClient from './CozyClient'
 import { receiveQueryResult, receiveQueryError, initQuery } from './store'
 import { normalizeDoc } from 'cozy-stack-client'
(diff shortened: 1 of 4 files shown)


cballevre:
@Crash-- before carrying out the tests, I would have liked to have your opinion on this PR

cballevre:
@Crash-- can I have an review on this revised version? pls 🙏

packages/cozy-client/src/mock.js
```diff
@@ -19,7 +21,7 @@ const fillQueryInsideClient = (client, queryName, queryOptions) => {
   }
 }
 
-const mockedQueryFromMockedRemoteData = remoteData => qdef => {
+const mockedQueryFromMockedRemoteData = remoteData => async qdef => {
```
Crash--, line 24: why this async stuff?

Crash-- (approved):

packages/cozy-client/src/mock.js
cballevre: I go an type error on the 105 line when I'm using the function without jest

PR #1432, opened 2023-12-18
feat: Update warning message when Flagship certification fails

Ldoppea:
Failing Flagship certification is expected when the app is run in a dev environment and when the user uses a non-official phone (i.e. alternative Android OS)

This would help developers understanding when this message is problematic and where it is not


paultranvan (approved):
:kissing_closed_eyes:

PR #1440, opened 2024-02-02
[SCR-188] feat: Remove undefined properties in qualification object

LucsT:
Why this PR:
Konnectors compare existing qualification and newly generate one to determinate if file metadata update is needed.
Undefined attributes in qualification are removed when saving files metadata, so we need qualification objects to be clean of undefined attributes before getting deliver to konnector by Qualification.getByLabel()

This PR want to avoid qualification object with undefined properties like
```
Qualification {
  label: 'other_tax_document',
  purpose: undefined,
  sourceCategory: undefined,
  sourceSubCategory: undefined,
  subjects: [ 'tax' ]
}

```

This PR deliver a constructor with only existing properties

```
Qualification { label: 'other_tax_document', subjects: [ 'tax' ] }
```

### DIFF
--- a/packages/cozy-client/src/models/document/qualification.spec.js
+++ b/packages/cozy-client/src/models/document/qualification.spec.js
@@ -6,6 +6,42 @@ import {
 import * as qualificationModel from '../../assets/qualifications.json'
 import logger from '../../logger'
 
+jest.mock('../../assets/qualifications.json', () => ({
+  qualifications: [
+    {
+      label: 'isp_invoice',
+      purpose: 'invoice',
+      sourceCategory: 'telecom',
+      sourceSubCategory: 'internet',
+      subjects: ['subscription']
+    },
+    {
+      label: 'national_id_card',
+      purpose: 'attestation',
+      sourceCategory: 'gov',
+      sourceSubCategory: 'civil_registration',
+      subjects: ['identity']
+    },
+    {
+      label: 'health_invoice',
+      purpose: 'invoice',
+      sourceCategory: 'health'
+    },
+    {
+      label: 'other_identity_document',
+      purpose: 'attestation',
+      subjects: ['identity']
+    },
+    {
+      label: 'dummy_empty'
+    }
+  ],
+  purposeKnownValues: ['attestation', 'invoice'],
+  sourceCategoryKnownValues: ['telecom', 'health', 'gov'],
+  sourceSubCategoryKnownValues: ['civil_registration', 'internet'],
+  subjectsKnownValues: ['identity', 'subscription']
+}))
+
 describe('document qualification', () => {
   beforeEach(() => {
     jest.spyOn(logger, 'warn').mockImplementation(() => jest.fn())
@@ -130,6 +166,17 @@ describe('qualifications items', () => {
   it('should always define a label', () => {
     qualificationModel.qualifications.forEach(q => {
       expect(q).toHaveProperty('label')
+    })
+  })
+
+  it('should not have undefined attributes', () => {
+    const qualification = Qualification.getByLabel('dummy_empty')
+    expect(Object.hasOwn(qualification, 'purpose')).toBe(false)
+    expect(Object.hasOwn(qualification, 'sourceCategory')).toBe(false)
+    expect(Object.hasOwn(qualification, 'sourceSubCategory')).toBe(false)
+    expect(Object.hasOwn(qualification, 'subjects')).toBe(false)
+    Object.keys(qualification).forEach(key => {
+      expect(qualification[key]).toBeDefined()
     })
   })
 


packages/cozy-client/src/models/document/qualification.js
```diff
@@ -36,12 +36,16 @@ export class Qualification {
       /**
        * @type {string?}
        */
-      this.purpose = attributes.purpose || qualification.purpose
-      this.sourceCategory =
+      const purpose = attributes.purpose || qualification.purpose
+      if (purpose) this.purpose = purpose
```
paultranvan, line 40: That's a nit, but I kind of dislike this syntax, especially with no indent. I find it quite hard to read, even more when there are several following with no line break. Good ol' verbose seems more readable to me:
```js
if (purpose) {
  this.purpose = purpose
}
```

paultranvan (approved):

doubleface (approved):

Crash-- (changes requested):
cf le commentaire

Crash--:
Since Konnectors are relying on this behavior (aka not returning undefined value) you should definitively add a test @LucsT in order to prevent any regression on that.

Crash--:
Also, you should add the context to your PR: why this modification is needed.

packages/cozy-client/src/models/document/qualification.spec.js
```diff
@@ -133,6 +133,12 @@ describe('qualifications items', () => {
     })
   })
 
+  it('should never have undefined attribute', () => {
+    // This qualification has no sourceSubCategory attribute
+    const qualification = Qualification.getByLabel('other_tax_document')
+    expect(qualification.hasOwnProperty('sourceSubCategory')).toBe(false)
+  })
```
Crash--, line 140: We should not rely on that "side effect". Because if we change `other_tax_document` to have a sourceSubCaterogry, the test will fail. This is not what is attended. 

Also, you should not use `howOwnProperty` but `hasOwn` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

Crash--: No, it's not. Unit test should not depends on side effect. 🙏

Crash--: Not specially. I don't think we need to map on all the qualificationModel, but why not... 

I would have done a mock of the qualificationModel to write what I want and that's all. 

Also, I don't know if the test is right anymore. Before you tested that the key was not present anymore, now you test that the value is not undefined. This is not the same thing. Are you sure about the choice you made?

paultranvan: I have trouble to understand why it takes so much time and discussions :confused: 
To mock the qualification file, you can simply do: 
```
jest.mock('../../assets/qualifications.json', () => ({
  qualifications: [
    {
      label: 'isp_invoice',
      purpose: 'invoice',
      sourceCategory: 'telecom',
      sourceSubCategory: 'internet',
      subjects: ['subscription']
    },
    {
      label: 'national_id_card',
      purpose: 'attestation',
      sourceCategory: 'gov',
      sourceSubCategory: 'civil_registration',
      subjects: ['identity']
    },
    {
      label: 'health_invoice',
      purpose: 'invoice',
      sourceCategory: 'health'
    },
    {
      label: 'other_identity_document',
      purpose: 'attestation',
      subjects: ['identity']
    }
  ],
  purposeKnownValues: ['attestation', 'invoice'],
  sourceCategoryKnownValues: ['telecom', 'health', 'gov'],
  sourceSubCategoryKnownValues: ['civil_registration', 'internet'],
  subjectsKnownValues: ['identity', 'subscription']
}))
```

And if you just want to test there is no undefined attribute: 
```
  it('should not have undefined attributes', () => {
    const qualification = Qualification.getByLabel('other_identity_document')
    Object.keys(qualification).forEach(key => {
      expect(qualification[key]).toBeDefined()
    })
  })
  ```
  And it's done

Comment (Crash--) [reply packages/cozy-client/src/models/document/qualification.spec.js:164]: > And is the same as my first test about other_tax_document. No ?

Bah non, ici on mock le model. Donc si on change le vrai code du model, ça n'aura pas d'impact sur le test. En gros, si je rajoute un sourceSubCaterogry à `other_tax_document` ça va pas péter le test. Et c'est ce qu'on veut. Donc ce mock est très bien. 

Après le test, j'aurais plus vu un truc du genre : 
```
it('should not have undefined attributes', () => {
    const qualification = Qualification.getByLabel('other_identity_document')
    expect(qualification.hasOwn('sourceSubCategory')).toBe(false)
  })
```

paultranvan: > That do not answer to Crash first request about code really working

I don't know what "code really working" means, but the point was to not rely on side-effect, i.e. the actual qualifications file. Here I mock the qualification file, so it answers this point.

> And is the same as my first test about other_tax_document. No ?

No, since your test was relying on the real qualifications file. It the real `other_tax_document` qualification change, your test might fail. If you mock the file as suggested, it won't, since the mocked values won't change.

paultranvan: De ce que je vois, ça reprend ma proposition, mais pas celle de @Crash-- , donc pas sûr que ce soit une synthèse ? 
C'est ok de mon côté, mais il manque ce qui était suggéré avec `expect(qualification.hasOwn('sourceSubCategory')).toBe(false)`

PR #1441, opened 2024-02-02
fix: Use complete settings id in useInstanceInfo

zatteo:
Related to : in next versions of cozy-client, it will not be possible to query settings with an incomplete id


Crash-- (approved):
<3

paultranvan (approved):

PR #1443, opened 2024-02-07
[SCR-164] feat: New strategy for ensureKonnectorFolder

doubleface:
Now konnector destination folders will have the following references:

- Main konnector folder : with a reference to the associated konnector
  slug
- konnector account folder : with a reference to the associated
  konnector slug + another reference to the associated
  sourceAccountIdentifier

This way, even if the user moves/or rename the konnector folder or the
konnector account folder, the konnector will find it and reuse it after
konnector disconnect/reconnect.

Existing konnector main folders and konnector account folders will be
migrated to these references
- docs: Add missing dir_id to IOCozyFolder type


packages/cozy-client/src/models/konnectorFolder.js
```diff
@@ -244,3 +281,157 @@ export const buildFolderPermission = folder => {
...
+export const findKonnectorAccountFolderByReference = async ({
+  client,
+  slug,
+  sourceAccountIdentifier
+}) => {
+  const fileCollection = client.collection('io.cozy.files')
+  const resp = await fileCollection.findReferencedBy({
+    _type: 'io.cozy.konnectors',
+    _id: `io.cozy.konnectors/${slug}`
+  })
```
paultranvan, line 381: Why not using  cozy-client query ? e.g. `client.query(Q('io.cozy.files').referencedBy({...})`
This way, you will benefit from the store

packages/cozy-client/src/models/konnectorFolder.js
```diff
@@ -244,3 +281,157 @@ export const buildFolderPermission = folder => {
...
+  const files = resp.included
+  return files.find(
+    file =>
+      file.type === 'directory' &&
+      !isTrashedFolder(file) &&
+      hasSourceAccountIdentifierReference(file, sourceAccountIdentifier)
+  )
+}
```
paultranvan, line 389: You are bypassing all the cozy-client built-in query features, which is not good :wink: 
First, why using collection and not cozy-client query ? e.g. `client.query(Q('io.cozy.files').referencedBy({...})`
This way, you will benefit from the store

Second, you query the references, and after that filter the retrieved values, while it is precisely the client.query job.
So, that would make something like this: 

```js
const resp = await client.query(Q('io.cozy.files').partialIndex({type: "directory", trashed: false}.referencedBy({...})
```

For the `hasSourceAccountIdentifierReference`, I'm not sure I understand correctly, the file must have a `io.cozy.konnectors` reference + `io.cozy.accounts.sourceAccountIdentifier`? And I'm surprised about this last one, is it a new doctype?

packages/cozy-client/src/models/konnectorFolder.js
```diff
@@ -244,3 +281,157 @@ export const buildFolderPermission = folder => {
...
+ * @param {string} options.slug - konnector slug
+ * @returns {Promise<import('../types').IOCozyFolder>}
+ */
+const findKonnectorMainFolderByReference = async ({ client, slug }) => {
+  const fileCollection = client.collection('io.cozy.files')
+  const { included: files } = await fileCollection.findReferencedBy({
+    _type: 'io.cozy.konnectors',
+    _id: `io.cozy.konnectors/${slug}`
+  })
+  return files.find(file => file.type === 'directory' && !isTrashedFolder(file))
```
paultranvan, line 405: Same remark about query

packages/cozy-client/src/models/konnectorFolder.js
doubleface: I did not know about this `Q().referencedBy`, I just took the same method used to query magic folders in [folder.js](https://github.com/cozy/cozy-client/blob/0bc07eb5b637c00bd5e7e96e6acf68b57ad02b5c/packages/cozy-client/src/models/folder.js#L93). Ok, let's use cozy-client query (And I will update folder.js.

And second, I'm surprised the request you propose works, since the collection findReferencedBy uses a specific stack request but I see there are helpers for this in the QueryDefinition. let's do this then, there will be a lot less code :+1:.

paultranvan (approved):
I didn't have time to completely understand this PR scope, and my main concerns looks addressed, so nothing blocking on my side

doubleface:
@Crash-- Let's merge this on next monday morning if I have no news.

PR #1446, opened 2024-02-16
[SCR-114] feat: Allow CozyClient to get sourceAccountIdentifier in appMetadatas

doubleface:
And add sourceAccountIdentifier in cozyMetadatas.sourceAccountIdentifier

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -693,7 +693,12 @@ client.query(Q('io.cozy.bills'))`)
       doctypeVersion = get(schema, 'doctypeVersion')
     }
 
-    const { slug, sourceAccount, version } = this.appMetadata
+    const {
+      slug,
+      sourceAccount,
+      sourceAccountIdentifier,
+      version
+    } = this.appMetadata
 
     const now = new Date().toISOString()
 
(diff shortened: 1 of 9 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -693,7 +693,12 @@ client.query(Q('io.cozy.bills'))`)
       doctypeVersion = get(schema, 'doctypeVersion')
     }
 
-    const { slug, sourceAccount, version } = this.appMetadata
+    const {
+      slug,
+      sourceAccount,
+      sourceAccountIdentifier,
+      version
+    } = this.appMetadata
```
Crash--, line 701: Don't we need to update the appMetadata type? If it's an any, maybe we should type it.

Crash-- (approved):

packages/cozy-client/src/CozyClient.js
Crash--: Is it documented somewhere to pass the sourceAccount and sourceAccountIdentifier? If no, you know my point ;)

doubleface: Enhanced  appMetadata type documentation + updated cozy-client metadata documentation with sourceAccountIdentifier

PR #1450, opened 2024-03-01
Fix: Normalize behavior of query definition

Merkur39:
Cette PR fait suite à cette issue https://github.com/cozy/cozy-client/issues/1367

Le fix de la JSDoc a mis en évidence une incohérence dans une condition (pour un log) du `useQuery` (https://github.com/cozy/cozy-client/compare/fix/ver-329?expand=1#diff-f6c79185f6ddf5d8fefc83938515c50a323cc9ebe2b1d1724c15a226008a273dR54)

Sachant que la queryDefinition peut être une fonction, dans ce cas cette condition ne sera jamais vrai.

### DIFF
--- a/packages/cozy-client/src/CozyClient.js
+++ b/packages/cozy-client/src/CozyClient.js
@@ -90,6 +90,9 @@ const securiseUri = uri => {
 
   return uri
 }
+const resolveToValue = fnOrValue => {
+  return typeof fnOrValue === 'function' ? fnOrValue() : fnOrValue
+}
 
 const DOC_CREATION = 'creation'
 const DOC_UPDATE = 'update'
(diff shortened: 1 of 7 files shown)


cballevre:
@Merkur39 I have the feeling that this [commit](https://github.com/cozy/cozy-client/pull/1471/commits/d0cda89ccba153efc35741fff5e0a8f1833d96c1) that I merged recently corrects part of what you wanted to do in this PR

packages/cozy-client/src/CozyClient.js
```diff
@@ -90,6 +90,9 @@ const securiseUri = uri => {
 
   return uri
 }
+const resolveToValue = fnOrValue => {
+  return typeof fnOrValue === 'function' ? fnOrValue() : fnOrValue
+}
```
paultranvan, line 95: This function is already defined [here](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/hooks/useQuery.js#L11), some mutualization would be nice

PR #1454, opened 2024-03-13
[VO-435] feat(SharingCollection): Add revokeGroup

cballevre:


packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -214,6 +214,20 @@ class SharingCollection extends DocumentCollection {
       uri`/sharings/${sharing._id}/recipients/${recipientIndex}`
     )
   }
+
+  /**
+   * Revoke only one group of the sharing.
+   *
+   * @param {object} sharing Sharing Object
```
Crash--, line 221: is it this one ? https://github.com/cozy/cozy-client/blob/28480d5ee61e715e4f0e005813417e7b253f691d/packages/cozy-stack-client/src/SharingCollection.js#L73-L84 

If yes, maybe we can create the type and use it instead of `object`?

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -214,6 +214,20 @@ class SharingCollection extends DocumentCollection {
...
+   * @param {number} groupIndex Index of this group in the groups array of the sharing
+   */
+  revokeGroup(sharing, groupIndex) {
+    return this.stackClient.fetchJSON(
+      'DELETE',
+      uri`/sharings/${sharing._id}/groups/${groupIndex}`
+    )
+  }
```
Crash--, line 229: Can't we expose this method via the DSL? (aka destroy)

packages/cozy-stack-client/src/SharingCollection.js
cballevre: No, it didn't match that one, but you're right to point out the lack of type. I've added the description of the object which was an `io.cozy.sharing` document

packages/cozy-stack-client/src/SharingCollection.js
cballevre: I'm wondering if this makes sense because we're not deleting the io.cozy.sharing document, just modifying it. The other cases of revocation would have to be managed at the same time. That's how I'd imagine the api would look: 

**Revoke a group :** 
```js
client.collection('io.cozy.sharing').destroy(sharingDocument, {
  group: 1
})
```
**Revoke a recipient :** 
```js
client.collection('io.cozy.sharing').destroy(sharingDocument, {
  recipient: 0
})
```
Defining for revocation of all recipients bother me a little in terms of naming. Was it something like this that you saw ?

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -18,21 +18,45 @@ const normalizeSharing = sharing => normalizeDoc(sharing, SHARING_DOCTYPE)
...
 /**
  * @typedef {object} Recipient An io.cozy.contact
  */
+
+/**
+ * @typedef {object} Group An io.cozy.contacts.groups
+ */
+
 /**
  * @typedef {object} Sharing An io.cozy.sharings document
```
Crash--, line 31: The convention, if this is the io.cozy.sharings document is to call is IoCozySharings.

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -214,10 +238,25 @@ class SharingCollection extends DocumentCollection {
...
   }
+
+  /**
+   * Revoke only one group of the sharing.
+   *
+   * @param {Sharing} sharing Sharing Object
+   * @param {number} groupIndex Index of this group in the groups array of the sharing
+   */
+  revokeGroup(sharing, groupIndex) {
+    console.log('sharing', sharing)
```
Crash--, line 249: to remove

packages/cozy-stack-client/src/SharingCollection.js
cballevre: I had a doubt, thanks for the clarification 👍

packages/cozy-stack-client/src/SharingCollection.js
Crash--: I don't have any clue, I just know that since you are bypassing the store, you'll have to update the state manually. 

I'll try to move on being able to call any method from the query (client.query(Q('io.cozy.sharings').call('revokeGroup', {})), it should fix the issue.

Crash-- (approved):

PR #1458, opened 2024-04-02
Add an helper to delete a contact

zatteo:
Depending if a contact has been imported by a konnector or not, there is logic to delete a contact.

We need to delete a contact at least in :
- cozy-contacts
- cozy-keys-browser

So let's centralize this logic.

Original function : https://github.com/cozy/cozy-contacts/blob/5c5df4606b53133fd3b0b2bfffc9fdda48e5052c/src/connections/allContacts.js#L77

### DIFF
--- a/packages/cozy-stack-client/src/ContactsCollection.js
+++ b/packages/cozy-stack-client/src/ContactsCollection.js
@@ -31,6 +31,28 @@ class ContactsCollection extends DocumentCollection {
     }
     return col
   }
+
+  /**
+   * Destroys a contact
+   *
+   * If the contact is linked to accounts, it will be trashed instead of being
+   * destroyed.
+   *
+   * @param  {object} contact - Contact to destroy. IT MUST BE THE FULL CONTACT OBJECT
+   * @returns {Promise} - Resolves when contact has been destroyed
+   */
+  async destroy(contact) {
+    const syncData = contact?.cozyMetadata?.sync || {}
+    const isLinkedToAccounts = Object.keys(syncData).length > 0
+    if (isLinkedToAccounts) {
+      return super.update({
+        ...contact,
+        trashed: true
+      })
+    } else {
+      return super.destroy(contact)
+    }
+  }
 }
 
 export const CONTACTS_DOCTYPE = 'io.cozy.contacts'
--- a/packages/cozy-stack-client/src/ContactsCollection.js
+++ b/packages/cozy-stack-client/src/ContactsCollection.js
@@ -1,4 +1,6 @@
 import DocumentCollection, { normalizeDoc } from './DocumentCollection'
+// @ts-ignore Need to import it to be used in jsdoc
+import { IOCozyContact } from 'cozy-client/dist/types'
 
 const normalizeMyselfResp = resp => {
   return {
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/models/contact.js
```diff
@@ -247,3 +248,21 @@ export const getIndexByFamilyNameGivenNameEmailCozyUrl = contact => {
...
+  const isLinkedToAccounts = Object.keys(syncData).length > 0
+  if (isLinkedToAccounts) {
+    return client.save({
+      ...contact,
+      trashed: true
+    })
+  } else {
+    return client.destroy(contact)
+  }
+}
```
Crash--, line 268: There is already a ContactsCollection, why not adding the destroy method there?

packages/cozy-client/src/models/contact.js
```diff
@@ -247,3 +248,21 @@ export const getIndexByFamilyNameGivenNameEmailCozyUrl = contact => {
  * @returns {boolean}
  */
 export const isContact = doc => doc._type === CONTACTS_DOCTYPE
+
+/**
+ * @param {CozyClient} client - Instance of CozyClient
+ * @param {object} contact - A contact
```
paultranvan, line 254: It would be nice to add a IOCozyContact type :pray: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/types.js

packages/cozy-client/src/models/contact.js
```diff
@@ -247,3 +248,21 @@ export const getIndexByFamilyNameGivenNameEmailCozyUrl = contact => {
...
+ * @returns {Promise<object>} - Modified or deleted contact
+ */
+export const deleteContact = (client, contact) => {
```
paultranvan, line 257: The specific CRUD methods tend to be defined in the associated collection, so I feel like this belong more on ContactCollection. This way, the client don't have to know the `deleteContact` existence, it just has to call the high-level `client.destroy` method

packages/cozy-client/src/models/contact.js
paultranvan: Oh well... Looks like I was a bit late https://github.com/cozy/cozy-client/pull/1458/files#r1548113459

packages/cozy-stack-client/src/ContactsCollection.js
```diff
@@ -31,6 +33,28 @@ class ContactsCollection extends DocumentCollection {
...
+   * Destroys a contact
+   *
+   * If the contact is linked to accounts, it will be trashed instead of being
+   * destroyed.
+   *
+   * @param  {object} contact - Contact to destroy. IT MUST BE THE FULL CONTACT OBJECT
+   * @returns {Promise} - Resolves when contact has been destroyed
+   */
+  async destroy(contact) {
+    const syncData = get(contact, 'cozyMetadata.sync', {})
```
paultranvan, line 47: Using an external lib just for this seems overkill, why not just `contact?.cozyMetadata?.sync`?

packages/cozy-stack-client/src/ContactsCollection.js
```diff
@@ -31,6 +33,28 @@ class ContactsCollection extends DocumentCollection {
...
     return col
   }
+
+  /**
+   * Destroys a contact
+   *
+   * If the contact is linked to accounts, it will be trashed instead of being
+   * destroyed.
+   *
+   * @param  {object} contact - Contact to destroy. IT MUST BE THE FULL CONTACT OBJECT
```
paultranvan, line 43: It would still be nice to have a IOCozyContact :wink:

paultranvan (approved):

packages/cozy-stack-client/src/ContactsCollection.js
zatteo: Tried something what do you think about it ?

paultranvan: Wow, this is very complete, thank you! :heart: 
Now, I'm wondering how we could link the doctype documentation to those types, it would great to have some automatic integration, so we don't need to have 2 places to update when a doctype attribute changes.
Maybe some script `yarn auto-update-types` that would fetch the cozy-doctype repo and autofill the types.js ?

zatteo: Or maybe the source of trush could be the ts types ? And then we could generate an beautiful online documentation ?

PR #1460, opened 2024-04-04
feat: Add methods for cozy-app settings and instance settings manipulation

Ldoppea:
In our cozy-apps, we often need to handle specific settings for each cozy-app. This includes:
- io.cozy.settings/io.cozy.settings.instance
- io.cozy.settings/io.cozy.settings.bitwarden
- io.cozy.home.settings
- io.cozy.coachco2.settings
- io.cozy.mespapiers.settings

In the current code base, all those access are made with different code that can be heterogeneous

We want to mutualize those access into two methods:
- client.getSetting(slug, key)
- client.saveAfterFetchSetting(slug, key, value)

`getSetting` is responsible to retrieve the given `key` from a specific cozy-app setting

`saveAfterFetchSetting` is responsible to update the given `key` for a specific cozy-app setting

Note that `saveAfterFetchSetting` will first fetch the cozy-app settings before editing and saving them. This prevents from having to pass existing settings as parameter and also this ensure we always inject up to date data into the database

Also, in order to have Reactive methods, this PR also implements the `useSetting` hook that relies on `useQuery` and `useMutation` to keep data up-to-date and to mutate them

Hook usage:
```js
const { value, save, query, mutation} =
  useSetting('instance', 'default_redirection', '<optional default value>')

...

if (hasQueryBeenLoaded(query)) {
  console.log('current default app is', value)
  save('mespapiers/')
}

...

useEffect(() => {
  if (mutationStatus === 'loaded') {
    console.log('default app updated')
  }
}, [mutation])

```

Get method usage:
```js
const value = await client.getSetting('home', 'default_redirection_snackbar_disabled')
```

Save method usage:
```js

// with raw value
await client.saveAfterFetchSetting('home', 'default_redirection_snackbar_disabled', false)

// or with an update method
await client.saveAfterFetchSetting('home', 'default_redirection_view_count', currentValue => currentValue + 1)
```

___

### TODO:
- [x] Add documentation near https://docs.cozy.io/en/cozy-client/react-integration/#3a-mutating-data-with-usemutation
- [ ] Update cozy-apps to call new methods
  - [x] In cozy-home ([here](https://github.com/cozy/cozy-home/blob/master/src/components/DefaultRedirectionSnackbar/helpers.js#L8-L38) and [here](https://github.com/cozy/cozy-home/blob/master/src/components/DefaultRedirectionSnackbar/helpers.js#L40-L58))
  - [ ] In mespapiers ([here](https://github.com/cozy/cozy-libs/commit/2438da5b24a1f54309e753173a67f85aa9691bfd))
  - [ ] In cozy-pass-web ([here](https://github.com/cozy/cozy-pass-web/blob/master/src/cozy/react/components/InstallationPage/components/InstallationStepMobilePart.jsx#L18-L28))
- [x] Add default value for undefined settings 

### What can be improved:
- Allow `onSuccess` and `onError` callbacks that can be passed to `useMutation()`


packages/cozy-client/src/helpers/settings.js
```diff
@@ -0,0 +1,182 @@
...
+ */
+export const saveAfterFetchSetting = async (
+  client,
+  slug,
+  key,
+  valueOrSetter
+) => {
+  const query = getQuery(slug)
+
+  const currentSettingsResult = await client.query(
```
cballevre, line 49: Wouldn't it be relevant to use a`fetchQueryAndGetFromState` here? This would look more like a useQuery by retrieving the cache if it exists

Ldoppea: Sorry I made the edit but did not pushed it yet seems some discussion are still in progress

I pushed theme here, but CI will fail due to types error: https://github.com/cozy/cozy-client/compare/0af2607eabfdcc3fe08b06f560166123a320b91d..c44f83a86d7b55513e3bdffbbff1ca9edf62cbb6

packages/cozy-client/src/helpers/settings.js
```diff
@@ -0,0 +1,188 @@
...
+ * @returns {import('../types').Query} - the Query that can be used to fetch the cozy-app settings
+ */
+export const getQuery = slug => {
+  if (slug === 'instance') {
+    return getNestedSettings(slug)
+  }
+
+  if (slug === 'passwords') {
+    return getNestedSettings('bitwarden')
+  }
```
paultranvan, line 144: I feel like 'instance' and 'passwords' should be explicitly mentioned in the doc + jsdoc, as it is not obvious. `instance` is mentioned as an example, but not `passwords`, and the latter is particularly not intuitive

packages/cozy-client/src/helpers/settings.js
```diff
@@ -0,0 +1,182 @@
+import CozyClient from '../CozyClient'
+import fetchPolicies from '../policies'
+import { Q } from '../queries/dsl'
+
+const defaultFetchPolicy = fetchPolicies.olderThan(60 * 60 * 1000)
```
zatteo: I'm thinking about this default fetch policy. Maybe it is too long ? I do not have any use case now to explain (😅) but I have the feeling that with 1 hour we can get outdated setting value where we de not want outdated setting value.

Maybe it should be short but enough to avoid 2 requests when doing a getSetting + saveAfterFetchSetting? Like 10 seconds?

packages/cozy-client/src/helpers/settings.js
```diff
@@ -0,0 +1,188 @@
...
+const defaultFetchPolicy = fetchPolicies.olderThan(10 * 1000)
+
+/**
+ * Query the cozy-app settings corresponding to the given slug and
+ * extract the value corresponding to the given `key`
+ *
+ * @param {CozyClient} client - Cozy client instance
+ * @param {string} slug - the cozy-app's slug containing the setting (can be 'instance' for global settings)
+ * @param {string} key - The name of the setting to retrieve
+ * @param {any} [defaultValue] - The default value of the setting if it does not exist
```
paultranvan, line 14: I'm not sure about the value of having this `defaultValue`. What does it bring to pass it as an argument? The app could just deal with it itself, no? 
I fell like it brings more confusion/API complexity than value

packages/cozy-client/src/helpers/settings.js
```diff
@@ -0,0 +1,188 @@
...
+
+const defaultFetchPolicy = fetchPolicies.olderThan(10 * 1000)
+
+/**
+ * Query the cozy-app settings corresponding to the given slug and
+ * extract the value corresponding to the given `key`
+ *
+ * @param {CozyClient} client - Cozy client instance
+ * @param {string} slug - the cozy-app's slug containing the setting (can be 'instance' for global settings)
+ * @param {string} key - The name of the setting to retrieve
```
paultranvan, line 13: Maybe having an array of keys could have been useful

packages/cozy-client/src/hooks/useSetting.js
```diff
@@ -0,0 +1,49 @@
...
+      return mutate(newSettings)
+    },
+    [key, mutate, settingsData, slug]
+  )
+
+  const settings = normalizeSettings(settingsData)
+
+  return {
+    query: settingsQuery,
+    value: settings?.[key],
```
zatteo: nit: `?.` should not be necessary because it comes from `normalizeSettings`?

packages/cozy-client/src/CozyClient.js
```diff
@@ -905,6 +905,10 @@ client.query(Q('io.cozy.bills'))`)
    * `getQueryFromState` or directly using `<Query />`. `<Query />` automatically
    * executes its query when mounted if no fetch policy has been indicated.
    *
+   * If the query is called under the fetch policy's delay, then the query
+   * is not executed and nothing is returned. If you need a result anyway,
+   * please use `fetchQueryAndGetFromState` instead
```
zatteo, line 910: 👍👍

packages/cozy-client/src/helpers/settings.js
Ldoppea: This is because I already felt on that trap here: https://github.com/cozy/cozy-home/pull/2117#discussion_r1560734796

I suppose that with an API that simplify the settings access it is easier to forget to check those kind of details.

But I'm opened to remove it if you still think this is not worth

packages/cozy-client/src/helpers/settings.js
Ldoppea: Why not, I'll try to add this

packages/cozy-client/src/helpers/settings.js
paultranvan: In my opinion, it's better to keep the API simple and avoid having too many arguments, especially for this simple thing that can be easily handled afterwards.
But this is a matter of opinion, feel free to keep it if you feel it

packages/cozy-client/src/helpers/settings.js
Ldoppea: I have this in my mind since yesterday but I'm struggling on choosing the best API.

### Declaration

For the declaration I imagine the following possibilities:

Use same hook for both single string param or multiple array param
```js
const { value } = useSetting('home', 'mysetting')
// value = 'some_value'

const { value } = useSetting('home', ['mysetting1', 'mysetting2'])
// value = {
//   mysetting1: 'some_value_1',
//   mysetting2: 'some_value_2'
// }
```

Use a different hooks for single and multiple params
```js
const { value } = useSetting('home', 'mysetting')
const { values } = useSettings('home', ['mysetting1', 'mysetting2'])
```

Use an object instead of an array to declare default values (but I we may [remove the default value thing](https://github.com/cozy/cozy-client/pull/1460#discussion_r1562642759), I think I can forget this option)
```js
const { value } = useSetting('home', 'mysetting')
const { values } = useSettings('home', {
  mysetting1: 'defaultValue1',
  mysetting2': 'defaultValue2'
})
```

### Mutation

Mutation is getting harder to approach

Should we return multiple save methods for each key?
```js
const { value, savemysetting1,  savemysetting2} = useSetting('home', ['mysetting1', 'mysetting2'])

savemysetting1('value1')
savemysetting2('value2')
```
or
```js
const { value, save } = useSetting('home', ['mysetting1', 'mysetting2'])

save.mysetting1('value1')
save.mysetting2('value2')
```
This constrains the API and so prevents errors, but sadly this also prevents mutating multiple settings at once

Should we return a save that takes an objet?
```js
const { value, save} = useSetting('home', ['mysetting1', 'mysetting2', 'mysetting3'])

savemysetting1({
  mysetting1: 'value1',
  mysetting3: 'value3'
})
```
This allow mutating multiple settings at once, but this makes the API less contraint and I'm wondering if we should allow inserting new keys or make a code that throw a specific error is the method is called with keys that are not present in the initial array?

Ldoppea: I made a proposal here: https://github.com/cozy/cozy-client/pull/1460/commits/dcb23c50e9c51f2db0cfd67ac1e22b736eb76078

(I still have to update documentation)

Usage would be:

### Hook

```js
const { value, save } = useSetting('home', ['mysetting1', 'mysetting2'])
...

// ✅ Valid
save({
  mySetting1: 'newValue1',
  mySetting2: 'newValue2'
})

// ✅ Valid
save({
  mySetting1: 'newValue1'
})

// ❌ Not valid because new key inserted
save({
  mySetting1: 'newValue1',
  mySetting2: 'newValue2',
  mySetting3: 'newValue3'
})

```

### Get
```js
const result = await getSetting(client, 'home', ['some_key'])
```


### Save
```js
// With raw value
await client.saveAfterFetchSetting(
  'instance',
  {
    some_instance_key: 'some_new_instance_value',
    some_instance_key2: 'some_new_instance_value_2'
  }
)


// With a method
await client.saveAfterFetchSetting(
  'instance',
  currentValue => ({
    ...currentValue,
    some_instance_key: currentValue.some_instance_key + 1,
    some_instance_key2: 'some_new_instance_value_2'
  }),
  ['some_instance_key']
)
```

When using a method we must pass a `setterKeys` parameters so the save method can still extract correct values. I would prefer not to have this parameter but I did not find a better way 😕

Ldoppea: Note that I typed those methods to prevent mistakes when using them. Here is how the linter and the auto-complete features behave:
image

Ldoppea: I updated the cozy-home PR with the new `multiple-settings` API: https://github.com/cozy/cozy-home/pull/2117/commits/e20239d6dee132ba566ca3f0a3e6ca075acab56c

paultranvan (approved):
Thanks for the work! Hopefully the beginning of future great contributions :smile:

packages/cozy-client/src/helpers/settings.js
Ldoppea: Documented [here](https://github.com/cozy/cozy-client/pull/1460/commits/c53f32c5da6d49f3de2d1a644f34505710c27d85)

docs/react-integration.md
```diff
@@ -17,7 +17,8 @@ Once connected, your components will receive the requesting data and a fetch sta
   - [3. Mutating data](#3-mutating-data)
   - [3.a Mutating data with `useMutation`](#3a-mutating-data-with-usemutation)
   - [3.b Mutating data with `Query`](#3a-mutating-data-with-query)
-  - [4. Testing](#4-testing)
+  - [4. Accessing and Mutating cozy-app settings with `useSetting`](#4-accessing-and-mutating-cozy-app-settings-with-usesetting)
```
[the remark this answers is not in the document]
Ldoppea: Moved [here](https://github.com/cozy/cozy-client/pull/1460/commits/de7a8c6c18a0ee2bd4f2996f48ed507306c491b2)

PR #1470, opened 2024-04-30
feat: Do not hydrate document with empty relationship data

Merkur39:
Cette PR propose un changement important dans le comportement de l'hydratation des documents! (voir le message de commit)


Merkur39:
I'm closing this, the approach isn't good. It's probably better to act on the "dehydrate" methods instead.

PR #1471, opened 2024-05-07
[VO-711] feat: Add a collection to access files into a Nextcloud

cballevre:
This PR is the first one to implement the new type `io.cozy.remote.nextcloud.files`. It allow us to interact with Nextcloud server inside Drive 🙂 

**Related PRs :**
- https://github.com/cozy/cozy-stack/pull/4392

### DIFF
--- a/packages/cozy-client/src/hooks/useQuery.js
+++ b/packages/cozy-client/src/hooks/useQuery.js
@@ -21,7 +21,7 @@ const generateFetchMoreQueryDefinition = queryResult => {
 /**
  * Fetches a queryDefinition and returns the queryState
  *
- * @param {QueryDefinition} queryDefinition - Definition created with Q()
+ * @param {QueryDefinition|(() => QueryDefinition)} queryDefinition - Definition created with Q()
  * @param {import("../types").QueryOptions} options - Options created with Q()
  * @returns {import("../types").UseQueryReturnValue}
  */
@@ -51,7 +51,7 @@ const useQuery = (queryDefinition, options) => {
 
   const client = useClient()
   const queryState = useSelector(() => {
-    if (options.singleDocData === undefined && queryDefinition.id) {
+    if (options.singleDocData === undefined && definition?.id) {
       logger.warn(
         'useQuery options.singleDocData will pass to true in a next version of cozy-client, please add it now to prevent any problem in the future.'
       )
--- a/packages/cozy-stack-client/src/NextcloudFilesCollection.js
+++ b/packages/cozy-stack-client/src/NextcloudFilesCollection.js
@@ -1,4 +1,5 @@
 import DocumentCollection from './DocumentCollection'
+import { forceDownload } from './utils'
 
 const NEXTCLOUD_FILES_DOCTYPE = 'io.cozy.remote.nextcloud.files'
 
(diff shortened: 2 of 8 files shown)


packages/cozy-stack-client/src/NextcloudFilesCollection.js
```diff
@@ -0,0 +1,46 @@
...
+const normalizeDoc = DocumentCollection.normalizeDoctypeJsonApi(
+  NEXTCLOUD_FILES_DOCTYPE
+)
+const normalizeNextcloudFile = (sourceAccount, path) => file => {
+  const extendedAttributes = {
+    ...file.attributes,
+    path: `${path}${path.endsWith('/') ? '' : '/'}${file.attributes.name}`,
+    cozyMetadata: {
+      ...file.attributes.cozyMetadata,
+      sourceAccount
```
paultranvan, line 15: Why do we have to manually set the `sourceAccount`? I feel like this should be up to the client to set it, just like the other `cozyMetadata` attributes?

packages/cozy-client/src/hooks/useQuery.js
```diff
@@ -51,7 +51,7 @@ const useQuery = (queryDefinition, options) => {
 
   const client = useClient()
   const queryState = useSelector(() => {
-    if (options.singleDocData === undefined && queryDefinition.id) {
+    if (options.singleDocData === undefined && definition?.id) {
```
paultranvan, line 54: Good catch!

packages/cozy-stack-client/src/NextcloudFilesCollection.js
```diff
@@ -0,0 +1,46 @@
...
+
+const NEXTCLOUD_FILES_DOCTYPE = 'io.cozy.remote.nextcloud.files'
+
+const normalizeDoc = DocumentCollection.normalizeDoctypeJsonApi(
+  NEXTCLOUD_FILES_DOCTYPE
+)
+const normalizeNextcloudFile = (sourceAccount, path) => file => {
+  const extendedAttributes = {
+    ...file.attributes,
+    path: `${path}${path.endsWith('/') ? '' : '/'}${file.attributes.name}`,
```
paultranvan, line 12: The stack does not return the path like the other files?

paultranvan (approved):
Some questions about the returned data, but looks good!

packages/cozy-stack-client/src/NextcloudFilesCollection.js
cballevre: This is for convenience, because in order to use all the routes in the stack we need to have the Nextcloud path and sourceAccount pair. This is only present in the shortcut describing the Nextcloud. To reuse the same action as the io.cozy.files, I've introduced it into the file so that you only have to pass a file as a parameter as expected

packages/cozy-client/src/hooks/useQuery.js
cballevre: Thanks to typing, yet another example, even with JSDoc

packages/cozy-stack-client/src/NextcloudFilesCollection.js
cballevre: Nop, but maybe she could do it

paultranvan: I think it should, @nono what do you think?

nono: Well, it makes the response bigger, and they are quite easy to compute client-side, so I decided to not include them.

PR #1475, opened 2024-05-15
Feat(country): Add translations for country names & nationalities

Merkur39:
This PR rearranges the country models and adds the possibility of having country names in different languages.
It will be easy to add more helpers as needed.


packages/cozy-client/src/models/country/countries.js
```diff
@@ -433,3 +276,68 @@ export const checkCountryCode = code => {
...
+    if (type === 'string') {
+      // @ts-ignore - "code" is a string
+      const codeUpperCase = code.toUpperCase()
+      return item.code2 === codeUpperCase || item.code3 === codeUpperCase
+    } else if (type === 'number') {
+      return item.number === code
+    } else {
+      return false
+    }
+  }
```
zatteo, line 306: Is there a reason to have this function inside the getCountryNameByCodeISO ?

zatteo (approved):

packages/cozy-client/src/models/country/countries.js
```diff
@@ -433,3 +276,68 @@ export const checkCountryCode = code => {
...
   return !!foundCountry
 }
+
+/**
+ * Get country name by its ISO 3166-1 (alpha-2, alpha-3 or numeric code)
+ *
+ * @param {string|number} code - ISO 3166-1 alpha-2, alpha-3 or numeric code
+ * @param {object} options - Options
+ * @param {string} options.lang - Language of country names
+ * @returns {string|number} - Country name or code if not found
```
paultranvan, line 286: Looks like it actually returns null if not found, not code (which is good I think)

packages/cozy-client/src/models/country/countries.js
paultranvan: Agreed, looks like a mistake

packages/cozy-client/src/models/country/locales/fr.json
```diff
@@ -0,0 +1,251 @@
...
+  "GG": "Guernesey",
+  "GH": "Ghana",
+  "GI": "Gibraltar",
+  "GL": "Groenland",
+  "GM": "Gambie",
+  "GN": "Guinée",
+  "GP": "Guadeloupe",
+  "GQ": "Guinée équatoriale",
+  "GR": "Grèce",
+  "GS": "Géorgie du Sud-et-les Îles Sandwich du Sud",
```
paultranvan, line 91: Makes me hungry

paultranvan (approved):

packages/cozy-client/src/models/country/countries.js
Merkur39: Not really, as it's a `.find` function below I thought it would be a good idea to scope it to the parent function.
But during coffee we noticed a major problem with the condition on the "number" type, which will never be true...

I'll ping you once it's fixed

Merkur39: Now that's fixed, if there's no alert I'll merge after the CI 👍

Merkur39:
@paultranvan @zatteo 
I'm taking advantage of this PR to add nationalities (current need).
I reviewed the code on this commit: 93585331cf4bce88d45fd665b505b0c4235b06f2

Sorry for the inconvenience, I promise not to add anything more on this PR 😁

paultranvan (approved):

PR #1480, opened 2024-05-21
fix: Force 2.7.0 version of node-fetch (SCR-604)

doubleface:
To avoid "socket hang up" / ECONNRESET on consecutive requests with
Node.js 19 and Node.js 20
https://github.com/node-fetch/node-fetch/issues/1735

We did not have this problem before because projects using cozy-client
had 2.7.0 in their yarn.lock even if lower version was specified in
package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,9 @@
     "unist-util-visit": "3.1.0",
     "whatwg-fetch": "3.5.0"
   },
+  "resolutions": {
+    "**/node-fetch": "^2.7.0"
+  },
   "private": true,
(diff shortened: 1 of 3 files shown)


Crash--:
Since Cozy is running on node 20, I think we should delete the dep to node-fetch since `native fetch` can / should be used. If a service or a konnector needs to have something different than `native fetch` then it should polyfill it itself. 

WDYT?

paultranvan:
What is the status of this PR @doubleface ?

Issue #1132, opened 2022-02-22
CozyClient should issue a warning when using `$allMatch` in a Mango selector


taratatach:
Our CouchDB version supports the `$allMatch` operator in Mango selectors but `cozy-client` uses the [sift](https://www.npmjs.com/package/sift) library to find which local documents are matched by queries and this library does not support `$allMatch`.

We should at least issue a warning in the same way we do when using `$exists: false`.

taratatach:
We could also define `$allMatch` ourselves via `sift`'s custom operations : https://github.com/crcn/sift.js#custom-operations

PR #1485, opened 2024-05-30
docs: Improve store internals documentation

paultranvan:


docs/architecture.md
```diff
@@ -112,71 +112,199 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
 
-`documents`: If the `_id` of the mutated document is not present, then we add the document. 
-If the `_id` is already there, then we update the content of the document with the fresh data
-(for now the work is done in extractAndMergeDocument() method). 
+Since there is a link between queries and documents through the ids list in each query object, we need to be able to deal with changes occuring in the documents collection, i.e. when documents are added or deleted.
 
-So if your app is linked to the documents store via getDocumentFromStore() for instance
-your app will have the updated value. 
+For this, cozy-client has a mechanism called "query updater": each time a query is run, all the queries in the store are re-evaluated directly in the store.
+That means that for each query, we take its definition, convert it into a mango predicate, and run it against the store's document collection.
```
cballevre, line 190: Maybe we can talk about sift.js here to explain why its important to have compatible query

docs/architecture.md
```diff
@@ -112,71 +112,199 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
 
-As already mentioned, when a mutated document already exists in the store throught its `_id`, it is updated with the new content. However, this update never removes existing fields, it only merges or adds new fields. This is typically useful when two queries retrieve the same document, but not with all the fields. 
+The fetch policy is a simple mechanism useful to prevent excessive fetching. 
+It is optionnaly defined for each query and define how many time should be elapsed since the last query execution before running again the query.
+
+In this example we query the checked todos, only if the last execution occured more than 10 minutes ago:
+```js
+const queryDef = Q('io.cozy.todos').where(checked: true)
+const queryOptions = {
+  as: 'checkedTodos', // query unique name, to find it in the store
```
cballevre, line 205: We could follow the best practice which I proposed to introduce in a previous comment before. This would give `io.cozy.todos/checked/true`

docs/architecture.md
```diff
@@ -112,71 +112,199 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+  fetchPolicy: CozyClient.fetchPolicies.olderThan(10 * 60 * 1000) // fetch policy, for 10min 
+}
+await client.query(queryDef, queryOptions)
+```
+
+If the query had been run before the 10 minutes, an early return will happen and nothing will be updated in the store.
+
+💡 Thanks to the auto-query updater, the data will not be stale if there is a data change in the app: indeed, any change on the documents will trigger the in-memory re-evaluation of the query and a render on the app if something is different.
+
+ℹ️ The `as` option is used to name the query, which is always a good idea to ensure that it has a unique name. When it's not specified, a random id is generated, resulting in potential query duplication in the store. Likewise, one must be very careful to prevent queries with the same name, as it would mixup results in the store.
```
cballevre, line 215: I would add a like to the devtool page here: `You can easily view each query using [Devtools](https://github.com/cozy/cozy-client/blob/master/docs/devtools.md) to identify duplicates.`

docs/architecture.md
```diff
@@ -112,71 +112,199 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+    participant App as App
+    participant CC as cozy-client
+    participant Store as Store
+    participant Backend as Backend
+
+    App->>CC: Query documents
+    CC->>Store: Check query
+    alt Query does not exist
+    CC->>Store: Init query
+    Store->>Store: Add query to collecton
```
cballevre, line 234: misspelling of the word `collection`

docs/architecture.md
```diff
@@ -112,71 +112,199 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+Here, we detail what happens when a document is updated, i.e. a mutation happens, typically after a `client.save`.
+
+If the document `_id` does not exist in the store's `documents` collection, i.e. after the a creation, it is simply added.
+If it already exists, the document content is updated in this collection.
+
+Furthermore, just like with the queries, any mutation triggers the [auto query updater](#auto-query-update), that can result in adding or removing document ids.
+
+Thanks to it, the app is guaranteed to have the up-to-date documents, returned from the `documents` collection.
+
+⚠️ As already mentioned, when a mutated document already exists in the store throught its `_id`, it is updated with the new content. However, this update never removes existing fields, it only merges or adds new fields. This is typically useful when two queries retrieve the same document, but not with all the fields.
```
cballevre, line 307: I think this based on the attribute `_rev` and not `_id` ([code](https://github.com/cozy/cozy-client/blob/ea8b7909ac8241fcc443fc16536985b702966e40/packages/cozy-client/src/CozyClient.js#L782))

cballevre (approved):

docs/architecture.md
paultranvan: I intentionally avoid mentioning sift. In the previous doc, it was mentioned in a quite technical manner that I found a bit hard to read. Here, I feel like it brings more distraction than value in this particular paragraph, but: you are definitely right that it is worth mentioning sift somewhere. So I'll add a dedicated section.

docs/architecture.md
paultranvan: Ah, good point, I always forget its existence

docs/architecture.md
paultranvan: Fair enough, but the code you mention is used to determine rather it is a create or update mutation, that will be used afterwards by the Link to call a create or update on the collection: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/StackLink.js#L117-L119
Here, we are on the documents store perspective, which rely on the `_id` existence:  https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/documents.js#L178-L183

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+
+    CC->>Store:  Load query
+    alt Without background fetching
+      Store->>Store: update query status
+      Store->>App: Loading query status
+      App->>App: Spinner
+    else With background fetching
+      Store->>Store: no change on query status
+      Store->>App: isFetching status
+      App->>App: Do nothing
```
Ldoppea, line 289: With `Do nothing` my understanding is that the component won't even display anything. Maybe something like `Display persisted data` would be more explicit?

docs/architecture.md
```diff
@@ -190,3 +320,13 @@ await client.query(Q1) // returns { _id: 'my-todo', _type: 'io.cozy.todos', labe
 
 Q1 retrieves the full content of 'my-todo' while Q2 only retrieves its label. When Q2 is run, 'my-todo' already exists in the store, as Q1 was previsouly run. However, the fields retrieved by Q1 for 'my-todos' but absent from Q2 are not lost. This way, when Q1 is run again, the results fetched from the store (thanks to the fetch policy) will include all the fields.
 
+The problem is when some fields are removed from a document: the store will not reflect this change and result in stale data.
+
+
+### In-memory queries evaluation
```
Ldoppea, line 326: Care there is an unbreakable space after `###` that breaks the rendering of the title

(also there is 3 other unbreakable spaces in this commit that don't break anything but you can still convert them to classic space)

Ldoppea (approved):
Nice documentation.

Adding links to PRs that implemented the features is 👌

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
 What do we mean exactly by saying "This redux store
 brings observability to cozy-client, and allows for connection of
 UI components to the data." Let's take a full exemple: 
 
-We have a component that displays a todolist: 
+Here is a simple React example on how to display data with cozy-client:
 ```jsx
-// TodoListComponent.jsx
 const { data, fetchStatus } = useQuery(Q('io.cozy.todos'), {'as': 'todoslist'})
+if(fetchStatus === 'loading'){
```
Crash--, line 128: While you're at it... 
```
if(fetchStatus === "failed"){
return <Error />
}
```

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
 }
 ```
 
-But we also have a component that gives use the opportunity to add a 
-new todo: 
-```jsx
-client.save({_type: 'io.cozy.todos', 'label': 'New TODO'});
+This way, a spinner will be displayed during the time the query is actually run. Once the query is done, data is available and can be displayed.
+
+Any data change will be automatically handled by cozy-client.
```
Crash--, line 138: It can be hard to understand that sentence. We should talk about the data concerned by the queries, not all the data. 

If you're still on the exemple, then having `Any data change on the io.cozy.todos doctype, will be automatically handled by cozy-client`. 

But I'm wondering also, what does `automatically handled` mean.

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+It can take the following values:
+- `pending`: the query is about to be run for the first time.
+- `loading`: the query is currently running and might return results. From a UX perspective, it is often used by apps to display a spinner.
+- `loaded`: the query had been run and returned results.
+- `failed`: the last query execution failed.
+
+When a query is called for the first time, it is first initialized in the `queries` collection in the store, with the `pending` status.
 
-### How does it works?
+If the query already exists in the store, its status is checked to ensure that it is not already in a `loading` state.
```
Crash--, line 176: This check is here to make the de-duplication. Maybe we should add that?

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
 
-When a `query` is resolved, CozyClient dispatches a `receiveQueryResult`
-action for a simple `get` but `receiveMutationResult` when we mutate 
-something. 
+Then, the query is "loaded", making its status on `loading`.
```
Crash--, line 178: on loaded?

docs/architecture.md
Crash--: You should link to `Auto query update`

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+For each query result, we check if there is any change in the known ids list, and update it accordingly.
 
-### About updates
+Thanks to this, we ensure that all the queries are always up-to-date, and apps can be effectively rendered thanks to the [react integration](/docs/react-integration.md).
 
-As already mentioned, when a mutated document already exists in the store throught its `_id`, it is updated with the new content. However, this update never removes existing fields, it only merges or adds new fields. This is typically useful when two queries retrieve the same document, but not with all the fields. 
+
+#### Fetch policy and query naming
+
+The fetch policy is a simple mechanism useful to prevent excessive fetching.
```
Crash--, line 199: nit: No need to say its "simple".

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
 
-### About updates
+Thanks to this, we ensure that all the queries are always up-to-date, and apps can be effectively rendered thanks to the [react integration](/docs/react-integration.md).
 
-As already mentioned, when a mutated document already exists in the store throught its `_id`, it is updated with the new content. However, this update never removes existing fields, it only merges or adds new fields. This is typically useful when two queries retrieve the same document, but not with all the fields. 
+
+#### Fetch policy and query naming
+
+The fetch policy is a simple mechanism useful to prevent excessive fetching. 
+It is optionnaly defined for each query and define how many time should be elapsed since the last query execution before running again the query.
```
Crash--, line 200: This can give the impression that even if the request is not called again, it will be executed again after this time. 

Maybe we should talk about "debounce"? Or say something like "if the query is called again whitin this time, it'll not be executed".

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+```js
+const queryDef = Q('io.cozy.todos').where(checked: true)
+const queryOptions = {
+  as: 'io.cozy.todos/checked/true', // query unique name, to find it in the store
+  fetchPolicy: CozyClient.fetchPolicies.olderThan(10 * 60 * 1000) // fetch policy, for 10min 
+}
+await client.query(queryDef, queryOptions)
+```
+
+If the query had been run before the 10 minutes, an early return will happen and nothing will be updated in the store.
```
Crash--, line 212: this exemple is not self speaking. 

We do generally: 
```js
const queryDef = Q('io.cozy.todos').where(checked: true)
const queryOptions = {
  as: 'io.cozy.todos/checked/true', // query unique name, to find it in the store
  fetchPolicy: CozyClient.fetchPolicies.olderThan(10 * 60 * 1000) // fetch policy, for 10min 
}
const result = await client.query(queryDef, queryOptions)
```

You say that an early return will happen, ok, so what is the result of `result`?

docs/architecture.md
Crash--: We should also add a link to the cozy-guidelines here.

docs/architecture.md
```diff
@@ -85,7 +85,7 @@ UI components to the data.
 advanced techniques where you create selectors directly on the data
 of cozy-client.
 
-The redux store is composed of two `slices` : `documents` and `queries`: 
+The redux store is composed of two collections: `documents` and `queries`:
```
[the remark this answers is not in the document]
paultranvan: I was wondering that as well and took some time to think about it. My feeling-based conclusion is that `slice` is too "js-oriented" and "collection" is generic enough to be understandable by any developer. 
I didn't know that slice is used by the redux community, which might make me reconsider. But for now, I prefer to stick with collection

docs/architecture.md
paultranvan: The "Do nothing" was to make an opposition to the "spinner" sceario, in reaction to the changing status. Here, no status change, so no need to do anything on the app side.
But ok to stress out the data displaying, I'll make the change

docs/architecture.md
paultranvan: One day, I'll figure out how to avoid this. But this day is not today.

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
-- If the query is concerned, then it checks if it has to remove / add the id of the `mutated` 
-document
-(for now the work is done mainly in queries.js/updateData())
+Since there is a link between queries and documents through the ids list in each query object, we need to be able to deal with changes occuring in the documents collection, i.e. when documents are added or deleted.
 
-So in our previous example our `todoslist` is concerned by the addition of the new todo, then 
-the `id` is added to `todolist` data, then the component linked will be refreshed with this new 
-document.
+For this, cozy-client has a mechanism called "query updater": each time a query is run, all the queries in the store are re-evaluated directly in the store.
+That means that for each query, we take its definition, convert it into a mango predicate (see [here](#in-memory-queries-evaluation) for details), and run it against the store's document collection.
```
Crash--, line 191: I don't think we talk about `mango` earlier. If we do, maybe we should link to it.

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+For each query result, we check if there is any change in the known ids list, and update it accordingly.
```
Crash--, line 192: Saying like this, it seems that only the `ids` are updated. Maybe we should talk about the documents too?

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+    CC->>App: Results
+    Store->>Store: Auto-update queries
+```
+
+#### Persisted store and background fetching stategy 
+
+An app can decide to persist its store in localstorage. See [here](https://github.com/cozy/cozy-home/blob/c47858515d4f95c24fa88a6b96bb0b500b947424/src/store/configureStore.js#L46) for an example.
+When doing so, the app loads the store when it starts and benefit from the instant available cached data.
+
+But, the data could have significantly evolved between the last store save and the start, making the documents potentially stale.
```
Crash--, line 257: instead of `documents` maybe we should talk more globally about `data`?

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+    CC->>App: New data
+    App->>App: Refresh display
+```
+
+### Updates
+
+Here, we detail what happens when a document is updated, i.e. a mutation happens, typically after a `client.save`.
+
+If the document `_id` does not exist in the store's `documents` collection, i.e. after the a creation, it is simply added.
+If it already exists, the document content is updated in this collection.
```
Crash--, line 303: after the roundtrip with the server? or this is an optimistic update?

docs/architecture.md
```diff
@@ -190,3 +320,13 @@ await client.query(Q1) // returns { _id: 'my-todo', _type: 'io.cozy.todos', labe
 
 Q1 retrieves the full content of 'my-todo' while Q2 only retrieves its label. When Q2 is run, 'my-todo' already exists in the store, as Q1 was previsouly run. However, the fields retrieved by Q1 for 'my-todos' but absent from Q2 are not lost. This way, when Q1 is run again, the results fetched from the store (thanks to the fetch policy) will include all the fields.
 
+The problem is when some fields are removed from a document: the store will not reflect this change and result in stale data.
```
Crash--, line 323: Maybe we should add a warning emoji for that? And also say, that to avoid this, instead of removing a field, you can set it to undefined as a workaround?

docs/architecture.md
```diff
@@ -190,3 +320,13 @@ await client.query(Q1) // returns { _id: 'my-todo', _type: 'io.cozy.todos', labe
...
+
+
+### In-memory queries evaluation
+
+Each time a change occurs in the store's `documents` collection, either through a query or an update, all the stored queries are dynamically re-evaluated.
+
+As we do not want to request the actual database each time, we do the evaluation directly from the store, in memory. To achieve this, we take advantage of the [mongo-like nature](https://docs.couchdb.org/en/stable/api/database/find.html#find-selectors) of our queries definition and use the [sift.js](https://github.com/crcn/sift.js) library to evaluate the queries against the store's `documents`.
```
Crash--, line 330: Some mango operators are not supported, please report an issue if you find one. See https://github.com/cozy/cozy-client/issues/1132 for instance

docs/architecture.md
```diff
@@ -190,3 +320,13 @@ await client.query(Q1) // returns { _id: 'my-todo', _type: 'io.cozy.todos', labe
...
+
+This library is able to take as input a mongo query definition and transform it into a javascript predicate. Then, all the queries are evaluated on the stores's `documents`, and efficiently updated accordingly to the results.
```
Crash--, line 332: I think that if the user is not well aware of couch and mongo, we use `mango` and `mongo` and it's certainly chaotic to understand.

Crash-- (approved):
Great write up @paultranvan!

docs/architecture.md
paultranvan: At this point, we dispatch the `loadQuery` (what I mean by "loaded"), setting the status from "loaded" to "loading" here: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/queries.js#L158-L160
But I agree that the loaded/loading terminology is a bit confusing, I remember that we had some brain damage here :smile: 
Maybe it would be more understandable to remove the "loaded" notion here and say something like this instead? 

>  Then, the query status is set on `loading` and the query actually run [...]

docs/architecture.md
paultranvan: But it *is* simple! Especially compared to the queries mechanisms :)
But yeah, it's good practice to avoid those terms in documentation

docs/architecture.md
paultranvan: > But I'm wondering also, what does automatically handled mean.

right, it's a bit an easy word, I removed it

docs/architecture.md
paultranvan: Here, in the in-memory query updates, only the ids are indeed updated. The document update happens after a mutation or a "fetch" query.

docs/architecture.md
paultranvan: It is after the "server" (how do we call it to include pouch ? backend?) response: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L1057-L1068

docs/architecture.md
paultranvan: Good point, I rephrased it.

docs/architecture.md
paultranvan: > We do generally:

you seem to have copy/pasted the the same code?

I changed the example to better demonstrate what happens, is it clearer?

docs/architecture.md
```diff
@@ -112,76 +112,206 @@ The redux store is composed of two `slices` : `documents` and `queries`:
...
+const queryOptions = {
+  as: 'io.cozy.todos/checked/true', // query unique name, to find it in the store
+  fetchPolicy: CozyClient.fetchPolicies.olderThan(10 * 60 * 1000) // fetch policy, for 10min 
+}
+await client.query(queryDef, queryOptions)
+```
+
+If the query had been run before the 10 minutes, an early return will happen and nothing will be updated in the store.
+
+💡 Thanks to the auto-query updater, the data will not be stale if there is a data change in the app: indeed, any change on the documents will trigger the in-memory re-evaluation of the query and a render on the app if something is different.
```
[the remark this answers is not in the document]
paultranvan: Correct, I added the react mention

zatteo:
Very interesting 👍 👍

docs/architecture.md
Crash--: I just added the `result` assignation of the `client.query`. Because if you are not using the useQuery hooks, for instance you are in a node context or within a React Class Component, then if you run a `client.query` with a fetchPolicy and that fetchPolicy says "I don't need to fetch", then the result will be empty. 

https://github.com/cozy/cozy-client/issues/931

docs/architecture.md
Crash--: Yeah, I'll call that the backend. We don't care if this is the cozy-stack or the Pouch. But beware that the returned document can be different depending on the backend (aka if the stack adds a few information for instance)

docs/architecture.md
```diff
@@ -112,71 +112,199 @@ The redux store is composed of two `slices` : `documents` and `queries`:
 
 }
 ```
-> ℹ️ If queries are not named, a name is automatically generated but this means that the `queries`
-slice can grow indefinitely if there are a large number of queries. This is why you are 
-encouraged to name your queries : `client.query(qdef, { as: 'finishedTodos'})`. 
+> ⚠️ If queries are not named, a name is automatically generated, but this means that the `queries`
+collection can grow indefinitely. This is why you are encouraged to name your queries : `client.query(qdef, { as: 'finishedTodos'})`.
```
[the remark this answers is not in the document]
Crash--: So, in order to follow the bast practices, we should name that query `as: io.cozy.todos/finished/true` 😽

docs/architecture.md
Crash--: I like what you suggest.

docs/architecture.md
paultranvan: > But beware that the returned document can be different depending on the backend

Indeed, but ideally, there should be no difference from the app perspective. In the future, I assume we might have some FileCollection in the pouch link to handle files path, for example.

PR #1487, opened 2024-05-31
feat: Forward operation in `StackLink.request()` when offline

Ldoppea:
> [!NOTE]
> This PR follows #1483 and #1486

On the Flagship app we want to serve `.query()` request using the `StackLink` when the device is connected, but we want to fallback to the `PouchLink` when we detect a connection loss

To allow this, we allow the consuming app to provide an `isOnline()` method to the `StackLink`

When provided, the `StackLink` will check for connectivity before doing its request. When offline, instead of processing the request, it will instead forward the request to the next `Link`

The code is generic so the consuming app can configure a `PouchLink` or anything else as the next `Link`

We want to also catch network errors when the `isOnline()` methods fails to detect connection loss, then we also fallback to the next `Link`

___

### Related PRs:
- #1483
- #1486
- cozy/cozy-flagship-app#1209

### DIFF
--- a/packages/cozy-client/src/StackLink.js
+++ b/packages/cozy-client/src/StackLink.js
@@ -58,8 +58,9 @@ export default class StackLink extends CozyLink {
    * @param {object} [options] - Options
    * @param  {object} [options.stackClient] - A StackClient
    * @param  {object} [options.client] - A StackClient (deprecated)
+   * @param {import('cozy-pouch-link/dist/types').LinkPlatform} [options.platform] Platform specific adapters and methods
    */
-  constructor({ client, stackClient } = {}) {
+  constructor({ client, stackClient, platform } = {}) {
     super()
     if (client) {
       logger.warn(
@@ -67,6 +68,7 @@ export default class StackLink extends CozyLink {
       )
     }
     this.stackClient = stackClient || client
+    this.isOnline = platform?.isOnline
   }
 
   registerClient(client) {
@@ -77,7 +79,11 @@ export default class StackLink extends CozyLink {
     this.stackClient = null
   }
 
-  request(operation, result, forward) {
+  async request(operation, result, forward) {
+    if (this.isOnline && !(await this.isOnline())) {
+      return forward(operation)
+    }
+
     if (operation.mutationType) {
       return this.executeMutation(operation, result, forward)
     }
--- a/packages/cozy-client/src/StackLink.js
+++ b/packages/cozy-client/src/StackLink.js
@@ -5,6 +5,7 @@ import CozyLink from './CozyLink'
 import { DOCTYPE_FILES } from './const'
 import { BulkEditError } from './errors'
 import logger from './logger'
+import { isReactNativeOfflineError } from './utils'
 
 /**
  *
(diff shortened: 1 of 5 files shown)


packages/cozy-client/src/StackLink.js
```diff
@@ -77,11 +80,26 @@ export default class StackLink extends CozyLink {
...
-      return this.executeMutation(operation, result, forward)
+  async request(operation, result, forward) {
+    if (this.isOnline && !(await this.isOnline())) {
+      return forward(operation)
     }
-    return this.executeQuery(operation)
+
+    try {
+      if (operation.mutationType) {
+        return await this.executeMutation(operation, result, forward)
```
paultranvan, line 90: Why did you add the `await`?

packages/cozy-client/src/StackLink.js
```diff
@@ -77,11 +80,26 @@ export default class StackLink extends CozyLink {
...
+      }
+      return await this.executeQuery(operation)
+    } catch (err) {
+      if (isReactNativeOfflineError(err)) {
+        return forward(operation)
```
paultranvan, line 95: I assume this is a fallback in case the `this.isOnline()` was not defined or wrongly returned true? I see the value, but I'm wondering whether it is good to have this kind of logic here. 
What if someone has a Cozy react-native app with only the StackLink defined? Here, it will forward the operation and the app will never receive the error

packages/cozy-client/src/StackLink.js
Ldoppea: Without the await, the `try...catch` would not be able to catch anything executed inside the `executeMutation()` method.

This is because we directly return the Promise, so the `try...catch` will directly release the flow control without waiting for the `executeMutation` result.

If we add an `await` then the `try...catch` will continue its action until the Promise is fulfilled.

packages/cozy-client/src/StackLink.js
Ldoppea: > I assume this is a fallback in case the this.isOnline() was not defined or wrongly returned true?

Yes, this is mainly because when a device loose the Internet connection, it can take a few seconds before the offline status is triggered. This often happens when you are connected to an Internet box through Wifi. If you unplug the internet connection, then your computer/phones will still displayed the "online" icon for a few seconds.

> What if someone has a Cozy react-native app with only the StackLink defined? Here, it will forward the operation and the app will never receive the error

Good catch. But I'm not sure that we can check if any `next link` exist. I'll investigate how to do this correctly.

packages/cozy-client/src/StackLink.js
paultranvan: Very clear, thank you

packages/cozy-client/src/utils.js
```diff
@@ -68,3 +68,15 @@ export const hasQueriesBeenLoaded = queriesResults => {
...
+/**
+ * Check is the error is about ReactNative not having access to internet
+ *
+ * @param {Error} err - The error to check
+ * @returns {boolean} True if the error is a network error, otherwise false
+ */
+export const isReactNativeOfflineError = err => {
+  // This error message is specific to ReactNative
+  // Network errors on a browser would produce another error.message
+  return err.message === 'Network request failed'
```
paultranvan, line 81: I wonder how stable this message is, and if it could change in future RN upgrades...

paultranvan (approved):
Looks good, but I have a concern on error handling when only StackLink is used

PR #1490, opened 2024-06-07
[VO-824] feat(nextcloud): Delete files

cballevre:


packages/cozy-stack-client/src/NextcloudFilesCollection.js
```diff
@@ -50,9 +50,43 @@ class NextcloudFilesCollection extends DocumentCollection {
...
+  /**
+   *  Move a file to the trash
+   *
+   * @param {object} file - The `io.cozy.remote.nextcloud.files` file to move to the trash
+   * @returns {Promise<object>}
+   * @throws {FetchError}
+   */
+  async destroy(file) {
+    const resp = await this.stackClient.fetch(
+      'DELETE',
```
paultranvan, line 62: Why the use of `fetch` instead a `fetchJSON`, that typically handle refresh token? https://github.com/cozy/cozy-client/blob/f5639c70556c701eee1e810fb750d960ccb548f5/packages/cozy-stack-client/src/CozyStackClient.js#L245

paultranvan (approved):
I concern about fetch vs fetchJSON but looks good otherwise

packages/cozy-stack-client/src/NextcloudFilesCollection.js
cballevre: I hadn't thought of bypassing the token refresh but I used this method because the stack doesn't return anything apart from an http code

Crash--: I think this is not the only method that doesn't return data. And we still use fetchJSON. No?

cballevre: Good question

PR #1491, opened 2024-06-11
fix: Adjust color of devtools to be dark mode compliant

JF-Cozy:


Merkur39 (approved):

Crash--:
Does it works even with this old version of cozy-ui https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/package.json#L55C1-L55C25 ?

JF-Cozy:
> Does it works even with this old version of cozy-ui https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/package.json#L55C1-L55C25 ?

Yes it does : https://github.com/cozy/cozy-ui/blob/v93.1.1/stylus/settings/palette.styl#L200

PR #1494, opened 2024-06-20
feat(FileCollection): Create archive link by file ids with specific pages

Merkur39:
- [x] Merge/deploy https://github.com/cozy/cozy-stack/pull/4426 before

Added `createArchiveLinkByIds` method and depreciated `getArchiveLinkByIds` method.

The name is more accurate and it allows us to refine the selection by giving the pages we want for each file.

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -721,20 +721,70 @@ class FileCollection extends DocumentCollection {
     return formatBytes(parseInt(file.size), decimal)
   }
 
-  async downloadArchive(fileIds, notSecureFilename = 'files') {
-    const filename = slugify(notSecureFilename)
-    const href = await this.getArchiveLinkByIds(fileIds, filename)
+  /**
+   * Attributes used for create archive link by ids
+   *
+   * @typedef {object} ArchivePages
+   * @property {string} id - Id of the file
+   * @property {number} page - The page number. PDF files only (1 is the first page)
+   */
+
+  /**
+   * Download an archive of the files
+   *
+   * @param {string[]} fileIds - List of file ids
+   * @param {string} [notSecureFilename] - Name of the archive (default: 'files')
+   * @param {object} [options] - Additional options
+   * @param {ArchivePages[]} [options.pages] - Array of objects, with `id` the file identifier, and `page` the page number (1 is the first page)
+   */
+  async downloadArchive(fileIds, notSecureFilename, { pages } = {}) {
+    const filename = notSecureFilename ? slugify(notSecureFilename) : 'files'
+    const href = await this.createArchiveLinkByIds({
+      ids: fileIds,
+      name: filename,
+      pages
+    })
     const fullpath = this.stackClient.fullpath(href)
     this.forceFileDownload(fullpath, filename + '.zip')
   }
 
+  /**
+   * @deprecated Use createArchiveLinkByIds instead
+   */
   async getArchiveLinkByIds(ids, name = 'files') {
+    logger.warn(
+      'CozyClient FileCollection getArchiveLinkByIds method is deprecated. Use createArchiveLinkByIds instead'
+    )
     const resp = await this.stackClient.fetchJSON('POST', '/files/archive', {
       data: {
         type: 'io.cozy.archives',
         attributes: {
           name,
           ids
+        }
+      }
+    })
+    return resp.links.related
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


zatteo (approved):

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -723,18 +723,40 @@ class FileCollection extends DocumentCollection {
...
+   * Get the archive link for a list of files
+   * The generated archive is temporary and is not persisted
+   *
+   * @param {object} params - Parameters
+   * @param {string[]} params.ids - List of file ids
+   * @param {string} [params.name] - Name of the archive (default: 'files')
+   * @param {ArchivePages[]} [params.pages] - Array of objects, with `id` the file identifier, and `page` the page number (1 is the first page)
+   * @returns {Promise<string>} - The archive link
+   */
+  async getArchiveLinkByIds({ ids, name = 'files', pages }) {
```
nono, line 752: I would have avoided a breaking change. For example, we can keep the `getArchiveLinkByIds` as it was, marks it as deprecated in the documentation, and add a new function `createArchiveLinkByIds`.

Merkur39: I like that, and the name will be more accurate too 👍

zatteo (approved):

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -721,14 +721,39 @@ class FileCollection extends DocumentCollection {
...
+   */
+
+  /**
+   * Download an archive of the files
+   *
+   * @param {string[]} fileIds - List of file ids
+   * @param {string} notSecureFilename - Name of the archive
+   * @param {ArchivePages[]} pages - Array of objects, with `id` the file identifier, and `page` the page number (1 is the first page)
+   */
+  async downloadArchive(fileIds, notSecureFilename = 'files', pages) {
```
zatteo, line 739: nit: Function signature is maybe a little confusing ? Seems like `notSecureFilename` was not mandatory before but now : 
- it is in jsdoc
- we added a argument behind so it is (even if we can pass undefined)

Merkur39: Good point, corrected

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -721,14 +721,39 @@ class FileCollection extends DocumentCollection {
...
 
-  async downloadArchive(fileIds, notSecureFilename = 'files') {
-    const filename = slugify(notSecureFilename)
-    const href = await this.getArchiveLinkByIds(fileIds, filename)
+  /**
+   * Attributes used for create archive link by ids
+   *
+   * @typedef {object} ArchivePages
+   * @property {string} id - Id of the file
+   * @property {number} page - The page number
```
paultranvan, line 729: This is only for PDF, right? Could you add this precision?

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -721,14 +721,39 @@ class FileCollection extends DocumentCollection {
...
+   */
+
+  /**
+   * Download an archive of the files
+   *
+   * @param {string[]} fileIds - List of file ids
+   * @param {string} [notSecureFilename] - Name of the archive (default: 'files')
+   * @param {ArchivePages[]} [pages] - Array of objects, with `id` the file identifier, and `page` the page number (1 is the first page)
```
paultranvan, line 737: The naming of `ArchivePages` and pages does not sound right here, as the archive could be non-pdf files, and thus without any concept of pages.

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -721,14 +721,39 @@ class FileCollection extends DocumentCollection {
...
+   */
+  async downloadArchive(fileIds, notSecureFilename, pages) {
```
paultranvan, line 739: I feel like `pages` should be optionnal, like  `async downloadArchive(fileIds, notSecureFilename, { pages } = {})`

packages/cozy-stack-client/src/FileCollection.js
Merkur39: In fact, I'm adding it!

packages/cozy-stack-client/src/FileCollection.js
Merkur39: Ok, I don't have a strong opinion on that 👍

PR #1495, opened 2024-07-01
[VO-874] feat: Improve index naming

cballevre:
Index naming has two weaknesses: 
- it doesn't include partial index values

This can cause differences between the expected result and the result when modifying queries.

**An example of the new index name**: 

Query definition :
```js
Q('io.cozy.files')
  .where({
    type: 'file',
  })
  .partialIndex({
    class: {
      $or: ['image', 'pdf'] 
    }
    trashed: {
      $eq: false 
    }
  })
  .sortBy([
    { name: 'desc' }
  ])
```
Old index name : `_design/by_type_and_name_filter_class_and_trashed`
New index name : `_design/by_type_and_name_filter_(class_(image_$or_pdf))_and_(trashed_$eq_false)`

### DIFF
--- a/docs/api/cozy-stack-client.md
+++ b/docs/api/cozy-stack-client.md
@@ -83,10 +83,13 @@ Serves to dedupe equal queries requested at the same time</p>
 <dt><a href="#getIllegalCharacters">getIllegalCharacters</a> ⇒ <code>string</code></dt>
 <dd><p>Get the list of illegal characters in the file name</p>
 </dd>
+<dt><a href="#makeKeyFromPartialFilter">makeKeyFromPartialFilter</a> ⇒ <code>string</code></dt>
+<dd><p>Process a partial filter to generate a string key</p>
+</dd>
 <dt><a href="#getIndexNameFromFields">getIndexNameFromFields</a> ⇒ <code>string</code></dt>
 <dd><p>Name an index, based on its indexed fields and partial filter.</p>
 <p>It follows this naming convention:
-`by_{indexed_field1}<em>and</em>{indexed_field2}<em>filter</em>{partial_filter_field1}<em>and</em>{partial_filter_field2}</p>
+<code>by_{indexed_field1}_and_{indexed_field2}_filter_({partial_filter.key1}_{partial_filter.value1})_and_({partial_filter.key2}_{partial_filter.value2})</code></p>
 </dd>
 <dt><a href="#transformSort">transformSort</a> ⇒ <code><a href="#MangoSort">MangoSort</a></code></dt>
 <dd><p>Transform sort into Array</p>
@@ -2160,13 +2163,25 @@ Get the list of illegal characters in the file name
 | --- | --- | --- |
 | name | <code>string</code> | the file name |
 
+<a name="makeKeyFromPartialFilter"></a>
+
+## makeKeyFromPartialFilter ⇒ <code>string</code>
+Process a partial filter to generate a string key
+
+**Kind**: global constant  
+**Returns**: <code>string</code> - - The string key of the processed partial filter  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| condition | <code>object</code> | An object representing the partial filter or a sub-condition of the partial filter |
+
 <a name="getIndexNameFromFields"></a>
 
 ## getIndexNameFromFields ⇒ <code>string</code>
 Name an index, based on its indexed fields and partial filter.
 
 It follows this naming convention:
-`by_{indexed_field1}_and_{indexed_field2}_filter_{partial_filter_field1}_and_{partial_filter_field2}
+`by_{indexed_field1}_and_{indexed_field2}_filter_({partial_filter.key1}_{partial_filter.value1})_and_({partial_filter.key2}_{partial_filter.value2})`
 
 **Kind**: global constant  
 **Returns**: <code>string</code> - The index name, built from the fields  
@@ -2174,8 +2189,7 @@ It follows this naming convention:
 | Param | Type | Description |
 | --- | --- | --- |
 | fields | <code>Array.&lt;string&gt;</code> | The indexed fields |
-| params | <code>object</code> | The additional params |
-| [params.partialFilterFields] | <code>Array.&lt;string&gt;</code> | The partial filter fields |
+| [partialFilter] | <code>object</code> | The partial filter |
 
 <a name="transformSort"></a>
 
--- a/docs/api/cozy-stack-client.md
+++ b/docs/api/cozy-stack-client.md
@@ -101,6 +101,9 @@ query to work</p>
 <dt><a href="#isMatchingIndex">isMatchingIndex</a> ⇒ <code>boolean</code></dt>
 <dd><p>Check if an index is matching the given fields</p>
 </dd>
+<dt><a href="#makeOperatorsExplicit">makeOperatorsExplicit</a> ⇒ <code>object</code></dt>
+<dd><p>Transform a query to make all operators explicit</p>
+</dd>
 <dt><a href="#getPermissionsFor">getPermissionsFor</a> ⇒ <code>object</code></dt>
 <dd><p>Build a permission set</p>
 </dd>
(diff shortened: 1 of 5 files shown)


packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -493,7 +517,7 @@ The returned documents are paginated by the stack.
    * @returns {MangoQueryOptions} Mango options
    */
   toMangoOptions(selector, options = {}) {
-    let { sort, indexedFields, partialFilter } = options
+    let { sort, indexedFields, partialFilter, partialIndex } = options
```
paultranvan, line 520: It seems there is a confusion between partialFilter and partialIndex here, this is supposed to be the same thing (more precisely, a partialFilter is the actual definition of a partialIndex).

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -219,6 +220,19 @@ class DocumentCollection {
...
   }
 
+  async migrateOldNamedIndex(
+    indexedFields,
+    partialFilter,
+    existingIndex,
+    indexName
+  ) {
+    await this.destroyIndex(existingIndex)
+    await this.createIndex(indexedFields, {
```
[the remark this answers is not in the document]
cballevre: We're reusing the existing migration now. If the content of the partialIndex changes, this creates a new index without being able to migrate the old one.

packages/cozy-stack-client/src/DocumentCollection.js
```diff
@@ -193,17 +193,18 @@ class DocumentCollection {
   }
 
   /**
-   * Migrate an existing unamed index to a named one.
+   * Migrate an existing index to a new name one.
```
paultranvan, line 196: s/new name one/ new one ?

packages/cozy-stack-client/src/mangoIndex.js
```diff
@@ -46,26 +46,61 @@ export const normalizeDesignDoc = designDoc => {
...
- * @param {Array<string>=} params.partialFilterFields - The partial filter fields
+ * @param {object} [options] - The additional params
+ * @param {object} [options.partialFilter] - The partial filter
  * @returns {string} The index name, built from the fields
  */
-export const getIndexNameFromFields = (
-  fields,
-  { partialFilterFields } = {}
-) => {
+export const getIndexNameFromFields = (fields, options = {}) => {
```
paultranvan, line 93: Why the generic and less explicit `options`  while only `partialFilter` is expected?

packages/cozy-stack-client/src/mangoIndex.js
```diff
@@ -46,26 +46,61 @@ export const normalizeDesignDoc = designDoc => {
   return { id, _id: id, ...designDoc.doc }
 }
 
+/**
+ * Process a partial filter to generate a string key
+ *
+ * @param {object} condition - An object representing the partial filter or a sub-condition of the partial filter
+ * @returns {string} - The string key of the processed partial filter
+ */
+export const processPartialFilter = condition => {
```
paultranvan, line 56: nit: `processPartialFilter` sounds like it will actually execute the partial filter, maybe it would be clearer with something like `makeKeyFromPartialFilter`

packages/cozy-stack-client/src/mangoIndex.spec.js
```diff
@@ -162,16 +162,78 @@ describe('getIndexFields', () => {
...
+        {
+          trashed: false
+        }
+      ]
+    }
+
+    expect(getIndexNameFromFields(fields, { partialFilter })).toEqual(
+      'by_dir_id_and_type_and_name_filter_(type_file)_and_((trashed_$exists_false)_$or_(trashed_false))'
+    )
+  })
```
paultranvan, line 228: Could you add a test with an explicit $or including 3 predicates? :pray:

docs/api/cozy-stack-client.md
```diff
@@ -101,6 +101,9 @@ query to work</p>
 <dt><a href="#isMatchingIndex">isMatchingIndex</a> ⇒ <code>boolean</code></dt>
```
paultranvan, line 101: About the commit description : `it removes implicit operators`. It's more : `It adds explicit operators when they are implicit, typically the `$and` and `$eq` operators.

packages/cozy-stack-client/src/mangoIndex.spec.js
```diff
@@ -237,3 +238,126 @@ describe('getIndexNameFromFields', () => {
...
+      $and: [
+        {
+          _id: {
+            $nin: ['id123', 'id456']
+          }
+        },
+        { type: { $eq: 'file' } }
+      ]
+    }
+    expect(makeOperatorsExplicit(query)).toEqual(expected)
```
paultranvan, line 306: Well done to have thought about this test!

packages/cozy-stack-client/src/mangoIndex.spec.js
```diff
@@ -237,3 +238,126 @@ describe('getIndexNameFromFields', () => {
...
+      type: {
+        $or: ['konnector', 'worker']
+      }
+    }
+    const expected = {
+      type: {
+        $or: ['konnector', 'worker']
+      }
+    }
+    expect(makeOperatorsExplicit(query)).toEqual(expected)
```
paultranvan, line 320: Have you tested this one on couchdb? I don't think that would work like this, I would expect this from couchdb: 
```
  "$or": [
    {
      "type": {
        "$eq": "konnector"
      }
    },
    {
      "type": {
        "$eq": "worker"
      }
    }
  ]
  ```

**Review by paultranvan — COMMENTED** [review]
That's a really great and long-expected work, thank you @cballevre!
My main concern is about the explicit operators, I feel like at least one of your test is not working at it should on couchdb. We might need to push tests a bit further, directly on couchdb

File: packages/cozy-stack-client/src/DocumentCollection.js (hunk 3 above)
Comment (cballevre) [reply packages/cozy-stack-client/src/DocumentCollection.js]: It's true that this sentence was incomprehensible. I detailed it a little more

File: packages/cozy-stack-client/src/mangoIndex.js (hunk 4 above)
Comment (cballevre) [reply packages/cozy-stack-client/src/mangoIndex.js]: It was a remnant of the previous function but you're right, it's clearer without the option.

File: packages/cozy-stack-client/src/mangoIndex.spec.js (hunk 6 above)
Comment (cballevre) [reply packages/cozy-stack-client/src/mangoIndex.spec.js:231]: Yes I done it will make `by_dir_id_and_type_and_name_filter_(type_file)_and_((trashed_$exists_false)_$or_(trashed_false)_$or_(class_image))`

File: packages/cozy-stack-client/src/mangoIndex.spec.js (hunk 9 above)
Comment (cballevre) [reply packages/cozy-stack-client/src/mangoIndex.spec.js]: Well found it was a failure. I've corrected the function to take this case into account

**Comment by paultranvan** [comment]
Related to https://github.com/cozy/cozy-client/issues/1374

**Comment by Crash--** [comment]
So much cleaner than what I did before 👏 👏 

Since you have changed the index name, you're going to have a pretty heavy DB server load when you release an app with this (mainly app with services ;))

File: packages/cozy-stack-client/src/mangoIndex.js [hunk 10]
```diff
@@ -46,26 +47,57 @@ export const normalizeDesignDoc = designDoc => {
...
   const indexName = `by_${fields.join('_and_')}`
-  return partialFilterFields
-    ? `${indexName}_filter_${partialFilterFields.join('_and_')}`
-    : indexName
+
+  if (partialFilter) {
+    return `${indexName}_filter_(${makeKeyFromPartialFilter(partialFilter)})`
+  }
+
+  return indexName
```
Crash--, line 98: You didn't change the method for the Pouch Link. Is it wanted? Is it our of scope ?

paultranvan:
> Since you have changed the index name, you're going to have a pretty heavy DB server load when you release an app with this (mainly app with services ;))

Since we migrate the existing index through a copy, we should avoid a too much consequent load (i.e. we don't recompute index)

packages/cozy-stack-client/src/mangoIndex.js
paultranvan: Since @Ldoppea is currently heavily working on pouch link and notably on this part, we didn't feel the need to put it in this PR scope. But this is indeed identified :wink:

Crash--:
Yep, I updated my comment just before yours ;). 

Just a last thing: I know that we had conflicts between the sack and cozy-client about index name. A few users had blank page because of this. I think that stack use partialFilters on some place, I'll double check that ;).

paultranvan:
> A few users had blank page because of this. I think that stack use partialFilters on some place, I'll double check that ;).

If you find any info about this, that would be precious!

PR #1496, opened 2024-07-10
[VO-847] feat: Add missing types for IOCozyFile

cballevre:
Since I've been trying to use Typescript in cozy-drive, I've found some missing types. I've proposed to add it to cozy-client so that it can be used in all your apps 🙂


Merkur39 (approved):
👍

PR #1497, opened 2024-07-12
feat(models/paper): Add isForeignPaper helper

Merkur39:


zatteo (approved):

packages/cozy-client/src/models/paper.js
```diff
@@ -493,3 +493,14 @@ export const makeExpirationDescription = (expirationDate, { lang }) => {
 
   return t('Scan.expiration.description', { duration: distance })
 }
+
+/**
+ * @param {IOCozyFile} file - io.cozy.files document
+ * @returns {boolean}
+ */
+export const isForeignPaper = file => {
```
paultranvan, line 501: I'm coming too late, but the "foreign" naming is a bit too subjective for me.  I would have preferred something more explicit and not implying that this lib is french-localized,  like "isNotFrenchPaper"

PR #1500, opened 2024-07-16
feat: Enable docs and types in cozy-pouch-link using Typedoc and TSC

Ldoppea:
We will start to use PouchLink in cozy-flagship-app (that uses typescript) and so we want to enable typing for this project


paultranvan (approved):
:+1:

PR #1504, opened 2024-07-26
feat: Make Pouch replication work with react-native

Ldoppea:
In previous PRs we implemented react-native support for CozyPouchLink

Some of the doctypes we want to replicate may not always exist in remote database. This is the case for `io.cozy.accounts` that do not exist until the first konnector is configured. When this happens we want the replication process to be resilient and the react-native app to still serve data when offline

Some other doctypes are protected by the remote cozy-stack. When this is the case we still want to have a local Pouch to store cozy-stack answers and make them available offline. This is not a real replication but is considered as is in the code.

This PR adapts the Pouch replication process to those new needs and also make it more robust by fixing some historic bugs

### DIFF
--- a/packages/cozy-pouch-link/src/PouchManager.js
+++ b/packages/cozy-pouch-link/src/PouchManager.js
@@ -8,7 +8,7 @@ import Loop from './loop'
 import logger from './logger'
 import { platformWeb } from './platformWeb'
 import { replicateOnce } from './replicateOnce'
-import { getDatabaseName } from './utils'
+import { formatAggregatedError, getDatabaseName } from './utils'
 
 const DEFAULT_DELAY = 30 * 1000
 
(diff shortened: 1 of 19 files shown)


packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -179,7 +179,15 @@ class PouchManager {
   }
 
   handleReplicationError(err) {
-    logger.warn('PouchManager: Error during replication', err)
+    let aggregatedMessage = ''
+    // @ts-ignore
+    // eslint-disable-next-line no-undef
+    if (err instanceof AggregateError) {
```
Crash--, line 186: TIL, thanks

packages/cozy-pouch-link/src/replicateOnce.js
```diff
@@ -36,7 +36,8 @@ export const replicateOnce = async pouchManager => {
 
       const getReplicationURL = () => pouchManager.getReplicationURL(doctype)
 
-      const initialReplication = !pouchManager.isSynced(doctype)
+      const initialReplication =
+        pouchManager.getSyncStatus(doctype) !== 'synced'
```
paultranvan, line 40: it might be worth to use constants for the different sync status

paultranvan (approved):
Very nice work and test coverage :clap:

packages/cozy-pouch-link/src/replicateOnce.js
Ldoppea: This is not needed as the project's compiler will prevent to use the wrong values

Autompletion:
image

IDE linter:
image

CI error:
image

PR #1505, opened 2024-07-26
feat: Implement FlagshipLink

Ldoppea:
When hosted in the Flagship app, we want the cozy-apps to be able to forward their queries to the native code

This will allow the native code to have control on the data lifecycle and chose between the cozy-stack or a local Pouch database as the source for executing the query

For now we use cozy-intent in order to communicate through post messages as it is the fastest implementation possible but this may change in the future if we encounter performances issues (i.e. using a local HTTP server)

___

### TODO:

- Upgrade cozy-intent after cozy/cozy-libs#2562 being merged (current commit has incorrect version)
___

### Related PRs:
- cozy/cozy-libs#2562

### DIFF
--- a/packages/cozy-client/src/FlagshipLink.js
+++ b/packages/cozy-client/src/FlagshipLink.js
@@ -1,0 +1,38 @@
+import CozyLink from './CozyLink'
+import logger from './logger'
+
+export default class FlagshipLink extends CozyLink {
+  /**
+   * @param {object} [options] - Options
+   * @param  {object} [options.stackClient] - A StackClient
+   * @param  {object} [options.client] - A StackClient (deprecated)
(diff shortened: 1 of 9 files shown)


packages/cozy-client/src/FlagshipLink.js
```diff
@@ -0,0 +1,38 @@
...
+   * @param  {import('cozy-intent').WebviewService} [options.webviewIntent] - The webview's intent reference
+   */
+  constructor({ client, stackClient, webviewIntent } = {}) {
+    super()
+    if (client) {
+      logger.warn(
+        'Using options.client is deprecated, prefer options.stackClient'
+      )
+    }
+    this.stackClient = stackClient || client
```
paultranvan, line 18: Why do we have a stackClient here?

Ldoppea: Removed in #1524

PR #1506, opened 2024-07-26
feat: Make PouchLink querying system compatible with new PouchDB version

Ldoppea:
In previous PRs we modified cozy-client and cozy-pouch-link to make them compatible with the react-native Flagship app

Since we didn't use this link for a long time but also because we use a newer PouchDB version, some queries do not work anymore (i.e. incompatible indexes, bad selectors etc)

This PR improve PouchLink querying system to make our app ecosystem compatible with new PouchDB version

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -54,8 +54,8 @@ export const isExpiredTokenError = pouchError => {
   return expiredTokenError.test(pouchError.error)
 }
 
-const normalizeAll = (docs, doctype) => {
-  return docs.map(doc => jsonapi.normalizeDoc(doc, doctype))
+const normalizeAll = client => (docs, doctype) => {
+  return docs.map(doc => jsonapi.normalizeDoc(doc, doctype, client))
 }
 
 /**
@@ -239,7 +239,7 @@ class PouchLink extends CozyLink {
    * Emits an event (pouchlink:sync:end) when the sync (all doctypes) is done
    */
   handleOnSync(doctypeUpdates) {
-    const normalizedData = mapValues(doctypeUpdates, normalizeAll)
+    const normalizedData = mapValues(doctypeUpdates, normalizeAll(this.client))
     if (this.client) {
       this.client.setData(normalizedData)
     }
@@ -590,7 +590,7 @@ class PouchLink extends CozyLink {
       res.limit = limit
       withRows = true
     }
-    return jsonapi.fromPouchResult(res, withRows, doctype)
+    return jsonapi.fromPouchResult(res, withRows, doctype, this.client)
   }
 
   async executeMutation(mutation, result, forward) {
@@ -618,7 +618,8 @@ class PouchLink extends CozyLink {
     return jsonapi.fromPouchResult(
       pouchRes,
       false,
-      getDoctypeFromOperation(mutation)
+      getDoctypeFromOperation(mutation),
+      this.client
     )
   }
 
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -556,6 +556,15 @@ class PouchLink extends CozyLink {
       res = withoutDesignDocuments(res)
       withRows = true
     } else {
+      if (indexedFields) {
+        for (const indexedField of indexedFields) {
+          if (!Object.keys(selector).includes(indexedField)) {
+            selector[indexedField] = {
+              $gt: null
+            }
+          }
+        }
+      }
       const findOpts = {
         sort,
         selector,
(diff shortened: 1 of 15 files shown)


packages/cozy-pouch-link/src/CozyPouchLink.spec.js
```diff
@@ -589,16 +567,31 @@ describe('CozyPouchLink', () => {
...
-      expect(spy).toHaveBeenCalledWith({ index: { fields: ['myIndex'] } })
+      expect(spy).toHaveBeenCalled()
+      expect(spy).toHaveBeenCalledWith({
+        index: {
+          ddoc: 'by_myIndex',
+          fields: ['myIndex'],
+          indexName: 'by_myIndex',
+          partial_filter_selector: undefined
+        }
+      })
```
paultranvan, line 597: it might be worth adding a simple test with an actual partialFilter

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -556,6 +556,15 @@ class PouchLink extends CozyLink {
       res = withoutDesignDocuments(res)
       withRows = true
     } else {
+      if (indexedFields) {
+        for (const indexedField of indexedFields) {
+          if (!Object.keys(selector).includes(indexedField)) {
+            selector[indexedField] = {
```
paultranvan: I feel like we should add a warning here, that we do some magic to please pouchdb but that ideally the query should be fixed

paultranvan: And we probably should extract this logic into a function with a test :wink:

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -556,18 +556,28 @@ class PouchLink extends CozyLink {
...
               $gt: null
             }
           }
         }
       }
+      if (shouldAddId) {
+        findSelector['_id'] = {
+          $gt: null
+        }
+      }
```
paultranvan, line 573: You don't need to force the `_id` selector if the mechanism on line 566-567 added a predicate.
Thus, we might refacto like this:
```js
let findSelector = selector || {}
// ...do the stuff on lines 566...
findSelector = Object.keys(selector).length > 0 ? findSelector : { _id: { $gt: null } } // PouchDB does not accept empty selector 
```

And this whole behavior should definitely be tested, as suggested in my previous comment :wink:

packages/cozy-pouch-link/src/jsonapi.js
```diff
@@ -22,12 +24,36 @@ export const normalizeDoc = (doc, doctype) => {
...
     delete normalizedDoc.rev
   }
+
+  normalizeLinks(normalizedDoc, doctype, client)
+
   return normalizedDoc
 }
 
+const normalizeLinks = (docRef, doctype, client) => {
+  if (doctype !== 'io.cozy.apps') {
```
paultranvan, line 34: I guess the function is more `normalizeAppsLinks` , then?

packages/cozy-pouch-link/src/jsonapi.js
```diff
@@ -22,12 +24,36 @@ export const normalizeDoc = (doc, doctype) => {
...
+    self: `/apps/${docRef.slug}`,
+    related: webLink,
+    icon: `/apps/${docRef.slug}/icon/${docRef.version}`
+  }
+}
+
 const filterDeletedDocumentsFromRows = doc => !!doc
 
-export const fromPouchResult = (res, withRows, doctype) => {
+export const fromPouchResult = (res, withRows, doctype, client) => {
```
paultranvan, line 56: nit: When dealing with many params, I prefer a `({param1, param2,...})` API

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -96,6 +96,7 @@ class PouchLink extends CozyLink {
     this.storage = new PouchLocalStorage(
       options.platform?.storage || platformWeb.storage
     )
+    this.ignoreWarmup = ignoreWarmup
```
paultranvan, line 99: If we don't specify any warmup query, this is useless, right?

packages/cozy-pouch-link/src/CozyPouchLink.js
Ldoppea: Done in #1524

packages/cozy-pouch-link/src/CozyPouchLink.js
Ldoppea: Done in #1524

packages/cozy-pouch-link/src/jsonapi.js
Ldoppea: Done in #1524

packages/cozy-pouch-link/src/CozyPouchLink.js
Ldoppea: Done in #1524

PR #1509, opened 2024-07-29
[VO-957] fix(move): Trash the conflicting file instead of the destination folder

cballevre:
The issue comes from a previous [PR](https://github.com/cozy/cozy-client/pull/1488). Where, I changed the function header to pass an object like `io.cozy.files` instead of his attributes directly.

**Old:** 
```
move(client, fileId, destination, force = false)
``` 
**New:** 
```
move(client, file, destination, { force } = { force: false })
```

The old object `destination` could have a path property which was corresponding to the file path into destination (destination folder path + filename of the moved file). In the new one, as the object is an `io.cozy.files` folder always contains the path of himself only.

When there is a conflit and the destination is not a shared folder (`force`), we want to move into the trash the file that already exists inside destination folder so we could try to move again the file. 

**Problem:**
This [condition](https://github.com/cozy/cozy-client/blob/73a573aa72e2b1d5d2589ae5c0afeec349092c1e/packages/cozy-client/src/models/file.js#L381) was always true so it would not add the filename anymore to the path. So the path will correspond to the destination folder and it will be moved to the trash

**Answer:** 
We compute the destination path each time to retrieve the file already in the folder and move it to the trash

### DIFF
--- a/docs/api/cozy-client/interfaces/models.file.FileUploadOptions.md
+++ b/docs/api/cozy-client/interfaces/models.file.FileUploadOptions.md
@@ -14,7 +14,7 @@ Conflict options
 
 *Defined in*
 
-[packages/cozy-client/src/models/file.js:499](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/file.js#L499)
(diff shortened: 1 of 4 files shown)


paultranvan (approved):
Thanks for the great explanation!

PR #1511, opened 2024-07-30
feat: Support executing queries from store

paultranvan:
We add a new option in the query API called `executeFromStore`. This is useful to allow the execution of a query directly on the store, i.e. without running on the link (StackLink or PouchLink), hence the database.

The use-case for this is to be able to have a very fast query when we have the relevant data on the store, typically after a sync. In such situation, we do not need to make a network call on the database as we can simply benefit from the local store.


Crash-- (approved):
nice!

PR #1514, opened 2024-08-01
fix: Support executing queries from store

Merkur39:
Small correction to https://github.com/cozy/cozy-client/pull/1511

### DIFF
--- a/packages/cozy-client/src/store/queries.js
+++ b/packages/cozy-client/src/store/queries.js
@@ -12,11 +12,12 @@ import sift from 'sift'
 
 import flag from 'cozy-flags'
 
-import { getCollectionFromSlice, getDocumentFromSlice } from './documents'
+import { getDocumentFromSlice } from './documents'
 import { isReceivingMutationResult } from './mutations'
 import { properId } from './helpers'
 import { isAGetByIdQuery, QueryDefinition } from '../queries/dsl'
 import logger from '../logger'
+import { getCollectionFromState } from './stateHelpers'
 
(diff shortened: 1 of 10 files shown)


packages/cozy-client/src/store/queries.js
```diff
@@ -310,12 +311,12 @@ const getSelectorFilterFn = queryDefinition => {
...
  * Execute the given query against the document state.
  *
- * @param {import('../types').DocumentsStateSlice} state - The documents state
+ * @param {import('../types').CozyStore} store - The cozy store
  * @param {QueryDefinition} queryDefinition - The query definition to execute
  * @returns {import("../types").QueryStateData} - The returned documents from the query
  */
-export const executeQueryFromState = (state, queryDefinition) => {
-  const documents = getCollectionFromSlice(state, queryDefinition.doctype)
+export const executeQueryFromState = (store, queryDefinition) => {
```
Ldoppea, line 318: nit: Here, naming makes it very difficult to understand what's happening:
- In CozyClient we call `executeQueryFromState(this.store.getState(), ...`
  - Due to the `getState()` method we expect to manipulate a `state`
- But in the method definition, we expect a `store` parameter, so we are not talking about `state` anymore
- But then we call `getCollectionFromState` with `state` in its name and use `state` everywhere in its declaration

Ldoppea (approved):

packages/cozy-client/src/store/queries.js
Merkur39: I understand, I went back to this naming.

PR #1516, opened 2024-08-09
[VO-1032] feat: Update iocozyfile typings 

acezard:


packages/cozy-client/src/types.js
```diff
@@ -441,6 +441,7 @@ import { QueryDefinition } from './queries/dsl'
  * @property {string} [noticePeriod] - Notice period of the paper, in days
  * @property {string} [datetime] - Image EXIF date, if relevant
  * @property {string} [instanceName] - Name of the instance
+ * @property {string} [type] - Type of the file (mainly used for shortcuts categories)
```
cballevre, line 444: I find that the description can be misleading since the file already has a type attribute but I don't know who to name it better 🤔

paultranvan: Misleading indeed, and also not documented, I don't see any mention here: https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.files.md nor here: https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.files_metadata.md
Where does it come from @acezard ?

paultranvan: > they don't follow the files doctype to the letter

There is absolutely no reason to not follow the rules, especially in this central doctype that is used by many apps.   Either it's documented in cozy-doctypes and typed in cozy-client, either it should not exist, there should be no in-between situation.  
@acezard  please, ask the guys in charge of the toutatice connector to make a PR on cozy-doctypes

cballevre: I'm scribing for the team here's the [PR](https://github.com/cozy/cozy-doctypes/pull/268) with the new documentation maybe which can help refine the description here

cballevre:
After discussions in this PR we have converged on a new data format. This PR implements these changes in the cozy-client types 🙂 

@paultranvan if you want to take a look 🙏 I would appreciate

paultranvan (approved):

PR #1523, opened 2024-09-02
feat: Add locales for new announcements remote-doctype

cballevre:


paultranvan (approved):
I assume locales for non-dev announcements will come later?

cballevre:
Yes you are right we don't have the server right now

PR #1525, opened 2024-09-04
fix: Remove attributes and meta attributes before persisting docs

Ldoppea:
Those doc's attributes are specific to the JSON API and should not be inserted into the Pouch database

This implies that we will have a difference on documents regarding if they are served through the cozy-stack or through a local PouchDB, the first one may include those fields in their result, but not the second one

So from now we should avoid, as much as possible, to relies on the `attributes` member to prevent bugs on Offline mode. Multiple commits to fix usages of `attributes` in cozy-client will be done after this one

Usage of `attributes` and `meta` members on cozy-app will also have to be fixed. This will be a requirement to implement Offline mode on cozy-apps

___

Related PRs:
- #1507

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -391,17 +391,44 @@ class PouchLink extends CozyLink {
     }
   }
 
-  async persistData(data, forward = doNothing) {
+  sanitizeJsonApi(data) {
     const docWithoutType = sanitized(data)
-    docWithoutType.cozyLocalOnly = true
+
+    /*
+    We persist in the local Pouch database all the documents that do not
+    exist on the remote Couch database
+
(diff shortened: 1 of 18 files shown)


packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -391,17 +391,44 @@ class PouchLink extends CozyLink {
...
+
+    /*
+    We persist in the local Pouch database all the documents that do not
+    exist on the remote Couch database
+
+    Those documents are computed by the cozy-stack then are sent to the
+    client using JSON-API format containing `attributes` and `meta`
+    attributes
+
+    Then the cozy-client would normalize those documents by spreading
```
paultranvan, line 405: s/cozy-client/cozy-stack-client

paultranvan (approved):
Thanks!

PR #1528, opened 2024-09-11
fix: Fix some problems related to Offline feature

Ldoppea:
All commits have their respective description

### DIFF
--- a/packages/cozy-client/src/models/file.js
+++ b/packages/cozy-client/src/models/file.js
@@ -719,7 +719,12 @@ export const downloadFile = async ({ client, file, url, webviewIntent }) => {
   const filesCollection = client.collection(DOCTYPE_FILES)
 
   if (isFlagshipApp() && webviewIntent && !isEncrypted(file)) {
-    return await webviewIntent.call('downloadFile', file)
+    const isFlagshipDownloadAvailable =
+      (await webviewIntent?.call('isAvailable', 'downloadFile')) ?? false
+
+    if (isFlagshipDownloadAvailable) {
+      return await webviewIntent.call('downloadFile', file)
+    }
   }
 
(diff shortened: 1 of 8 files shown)


packages/cozy-client/src/models/file.js
```diff
@@ -719,7 +719,12 @@ export const downloadFile = async ({ client, file, url, webviewIntent }) => {
   const filesCollection = client.collection(DOCTYPE_FILES)
 
   if (isFlagshipApp() && webviewIntent && !isEncrypted(file)) {
-    return await webviewIntent.call('downloadFile', file)
+    const isFlagshipDownloadAvailable =
+      (await webviewIntent?.call('isAvailable', 'downloadFile')) ?? false
+
+    if (isFlagshipDownloadAvailable) {
+      return await webviewIntent.call('downloadFile', file)
+    }
```
paultranvan, line 727: This is a "pessimistic" approach, as we check everytime if the method is available. Would it be possible to have an "optimistic" strategy, where we call the method anyway, and handle an error returned by the flagship?
Note I'm asking this to avoid this extra check call, but maybe the performances are good enough to not worry about it

paultranvan (approved):

packages/cozy-client/src/models/file.js
Ldoppea: For small payloads, the order of magnitude is 1ms on iOS and 20-50ms on Android (those are numbers from RN66 but I did not benchmark them again since latest updates to RN73).

As the download action is rare enough and because the actual downloading of the file will take hundreds of ms, I consider this to be acceptable.

paultranvan: All right then

PR #1533, opened 2024-09-19
feat: Improve CozyLink typing and add `reset()` method

Ldoppea:
In the Flagship app, we want to be able to reset the local PouchDB files in order to prevent beta testers to be blocked by an erroneous Pouch replication

To make this possible we want to expose a public method that resets all the cozy-client's links

### DIFF
--- a/packages/cozy-client/src/FlagshipLink.js
+++ b/packages/cozy-client/src/FlagshipLink.js
@@ -14,7 +14,7 @@ export default class FlagshipLink extends CozyLink {
     // does nothing, we don't need any client for this kind of link
   }
 
-  reset() {
+  async reset() {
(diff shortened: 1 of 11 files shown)


packages/cozy-client/src/FlagshipLink.js
```diff
@@ -14,7 +14,7 @@ export default class FlagshipLink extends CozyLink {
     // does nothing, we don't need any client for this kind of link
   }
 
-  reset() {
+  async reset() {
     // does nothing, we don't need any client for this kind of link
```
paultranvan, line 18: Eventually it might be useful to call the AA to allow such reset from a web app

paultranvan (approved):

PR #1534, opened 2024-09-19
feat: Handle queries that fetch Konnectors by channel

Ldoppea:
`fetchJSON()` calls are not compatible with offline mode unless we wrap them in the `.query()` dsl

Recently we added a `fetchJSON()` call inside of cozy-home in order to display the new grouped-by-category folders in the home's konnectors section

So we want to move this call inside of the AppsRegistryCollection in order to make it compatible with offline mode

Related PR: cozy/cozy-home#2186


packages/cozy-stack-client/src/AppsRegistryCollection.js
```diff
@@ -27,6 +27,16 @@ export const normalizeAppFromRegistry = (data, doctype) => {
   }
 }
 
+const fetchKonnectorsByChannel = async (channel, doctype, stackClient) => {
+  const resp = await stackClient.fetchJSON(
+    'GET',
+    `/registry?versionsChannel=${channel}&filter[type]=konnector&limit=300`
```
paultranvan, line 33: Since we cannot increase the limit from the app anymore (we might need eventually to extend the `getById` to allow options), I would increase it from 300 to 500

paultranvan (approved):

packages/cozy-stack-client/src/AppsRegistryCollection.js
Ldoppea: Done here: https://github.com/cozy/cozy-client/compare/2fcf290a895419d01418d30bbc828d5d2c631c3d..ecb276d3a2c4d74390b69572ea5e16387ec5be9d

I would suggest to skip the "options" thing and to implement it if we need to change the value someday.

PR #1536, opened 2024-09-23
feat: Rename FlagshipLink to WebFlagshipLink

Ldoppea:
This should clarify that this link should be run on the web side and not on the Flagship app side


paultranvan (approved):
Thanks!

PR #1537, opened 2024-10-03
Improve pouchDB performances

paultranvan:
We noticed severe performances issues when using pouchdb in a  react-native environment.
See commit descriptions for more insights

**BEFORE**
```
Indexing time : 53 868 ms
Query time : 51 499 ms

```
**AFTER**
```
Indexing time : 31 833 ms
Query time : 1 854 ms

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -539,6 +539,19 @@ class PouchLink extends CozyLink {
 
     if (!indexedFields) {
       indexedFields = getIndexFields(options)
+    } else if (partialFilter) {
+      // Some pouch adapters does not support partialIndex, e.g. with websql in react-native
+      // Therefore, we need to force the indexing the partialIndex fields to ensure they will be
+      // included in the actual index. Thanks to this, docs with missing fields will be excluded
+      // from the index.
+      // Note the $exists: false case should be handled in-memory.
+      indexedFields = Array.from(
+        new Set([...indexedFields, ...Object.keys(partialFilter)])
+      )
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/cozy-pouch-link/src/CozyPouchLink.js [hunk 1]
```diff
@@ -539,6 +539,19 @@ class PouchLink extends CozyLink {
...
+    } else if (partialFilter) {
+      // Some pouch adapters does not support partialIndex, e.g. with websql in react-native
+      // Therefore, we need to force the indexing the partialIndex fields to ensure they will be
+      // included in the actual index. Thanks to this, docs with missing fields will be excluded
+      // from the index.
+      // Note the $exists: false case should be handled in-memory.
+      indexedFields = Array.from(
+        new Set([...indexedFields, ...Object.keys(partialFilter)])
+      )
+      // FIXME: should properly handle n-level attributes
```
Ldoppea, line 551: What's the impact for now? And should we handle this soon? Or it is for an hypothetical future?

paultranvan: For now it means that if we have queries with partialIndex having attributes not on the root level, i.e. with `$and: [{...}]`, we won't benefit from this fix. Thus slow queries. 
So *hopefully* there is no regression  regarding the current offline, **but** we should test it carefully

paultranvan:
> 30s to index 40k documents is still huge :(

It is indeed. While this time would be dramatically reduced on a more powerful device, this is disappointing, but not so surprising regarding previous performances measures, that tend to show massive performances differences depending on the adapter and document size : 
image

PR #1546, opened 2024-10-17
[DAT-61] feat: Handle timeout errors for mango queries

paultranvan:
When making a mango query, we enforce the use of an index to avoid very costly full-scans on the database. We follow an "optimistic" approach, meaning we assume the existence of the index associated with the query and execute it without waiting. In some cases, CouchDB will immediately return an error if it doesn't find the index (typically when we request sorting). However, sometimes it will still attempt to execute the query by foolishly forcing a costly full-scan on the database (we did not find a way to prevent this...).
CouchDB returns a warning when it does this, which allows cozy-client to detect it and create the index on the fly.

Here the cozy-stack timeout prevents us to catch the warning. So, we catch the timeout error itself to apply the same missing index strategy in this situation.


cballevre (approved):
LGTM 👍

PR #1547, opened 2024-10-17
fix: Allow remote request on web environment

paultranvan:
When using PouchLink in a web environment, we need to pass the cookies through the fetch method in order to authenticate to the stack. Note we didn't have the issue before we were using PouchLink in a mobile environment, with OAuth authentication.

For more insights, see
```
https://reqbin.com/code/javascript/lcpj87js/
javascript-fetch-with-credentials
```


cballevre (approved):
LGTM 👍

PR #1548, opened 2024-10-17
feat: Allow one-shot replication 

paultranvan:
When a replication was enabled, we used to automatically start a loop
replication process, on a fixed interval.
However, we might want a one-shot replication that could be manually
triggered through a `pouchLink.startReplication`


cballevre (approved):
LGTM 👍

PR #1549, opened 2024-10-17
[VO-1153] fix(mango): Order alphabetically object inside arrays based on 1st key

cballevre:
When CouchDB creates an index, it orders the partial filters alphabetically. This leads to errors in the comparison to find the existing partial filters. To correct the problem, we have improved the function to make our partial filters explicit so that it sorts in the same way as CouchDB.


paultranvan (approved):
Looks good! :muscle:

PR #1552, opened 2024-10-29
[PRD-573] Destroy local storage

paultranvan:
We introduce a `destroy` method in the storage interface, that will
be called after a `client.reset()`.
We implement this `destroy` in the localStorage platform, but it should
be implemented accordingly to deal with other storage context, such as
with indexedDB for workers or async-storage for react-native


packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -111,12 +111,9 @@ class PouchManager {
...
     await this.clearSyncedDoctypes()
     await this.clearWarmedUpQueries()
-    await this.storage.destroyAllDoctypeLastSequence()
-    await this.storage.destroyAllLastReplicatedDocID()
+    await this.storage.destroy()
 
-    return Promise.all(
-      Object.values(this.pouches).map(pouch => pouch.destroy())
-    )
+    await allSettled(Object.values(this.pouches).map(pouch => pouch.destroy()))
```
Merkur39, line 116: There's nothing left to return. Is this intentional ?

Merkur39 (approved):
👍

packages/cozy-pouch-link/src/utils.js
```diff
@@ -27,3 +27,47 @@ export const formatAggregatedError = aggregatedError => {
...
+ * @property {undefined} value - The resolved value of the promise (undefined when rejected)
+ */
+
+/**
+ * Takes an iterable of promises as input and returns a single Promise.
+ * This returned promise fulfills when all of the input's promises settle (including
+ * when an empty iterable is passed), with an array of objects that describe the
+ * outcome of each promise.
+ * This implementation is useful for env with no support of the native Promise.allSettled,
+ * typically react-native 0.66
```
zatteo, line 51: According to https://github.com/facebook/react-native/releases/tag/v0.70.6, we can now safely use Promise.allSettled on the flagship app.

packages/cozy-pouch-link/src/platformWeb.js
```diff
@@ -18,6 +18,9 @@ const storage = {
   },
   removeItem: async key => {
     return window.localStorage.removeItem(key)
+  },
+  destroy: async () => {
+    return window.localStorage.clear()
```
Ldoppea, line 23: The side effect of clearing the entire localStorage may be problematic as this is not the only mechanism relying on the local storage.

For example, on the Flagship app we expect to call the `link.reset()` method if we need to purge the offline data. For now we implemented it through an UniversalLink in this PR: https://github.com/cozy/cozy-flagship-app/pull/1247

With your implementation, the UL would also clear the persisted cozy-client.

I would prefer to call `removeItem()` for each [pouch related key](https://github.com/cozy/cozy-client/blob/800005e88f1b2e63ecaa7deeb81e16d7d8350f11/packages/cozy-pouch-link/src/localStorage.js#L1-L8)

Ldoppea (changes requested):

packages/cozy-pouch-link/src/platformWeb.js
paultranvan: > I would prefer to call removeItem() for each [pouch related key](https://github.com/cozy/cozy-client/blob/800005e88f1b2e63ecaa7deeb81e16d7d8350f11/packages/cozy-pouch-link/src/localStorage.js#L1-L8)

That's  what I wanted to avoid, but fair enough, this is less risky

packages/cozy-pouch-link/src/utils.js
paultranvan: But I assume we might have users using out-of-date flagship in RN 0.66?

zatteo: Maybe, not prioritary to do anyway, but I think it starts to be safe to change

packages/cozy-pouch-link/src/platformWeb.js
paultranvan: For you @Ldoppea [d45e99f](https://github.com/cozy/cozy-client/pull/1552/commits/d45e99ff7cb458896e28d089a2818fcd7ab4447f)

packages/cozy-pouch-link/src/PouchManager.js
paultranvan: Not really, I changed it

Ldoppea (approved):

packages/cozy-pouch-link/src/utils.js
Ldoppea: This won't be problematic as this is packaged in the build, so old flagship version will use the old code with custom method and new versions will use the new one. No retro-compatibility needed.

Ldoppea: But there is no priority at all as those methods are equivalent

PR #1557, opened 2024-11-06
feat: Remove devtools

JF-Cozy:
BREAKING CHANGE:
- you must change the import of devtools from `import CozyDevtools from 'cozy-client/dist/devtools'` to `import CozyDevtools from 'cozy-devtools'`
- you no longer need cozy-ui as dep for cozy-client usage


cballevre (approved):
You have removed the devtools documentation, but it should add a short text presenting this tool and redirecting to its documentation inside Cozy Client. Also it would be nice to add a line `yarn add  cozy-devtools` about installation inside the breaking change comment. Otherwise, the rest looks good to me.

JF-Cozy:
> You have removed the devtools documentation, but it should add a short text presenting this tool and redirecting to its documentation inside Cozy Client. Also it would be nice to add a line `yarn add cozy-devtools` about installation inside the breaking change comment. Otherwise, the rest looks good to me.

Pour la doc, c'est bien le cas, la correction du linter pollue un peu mais on a bien 
image
et 
image

PR #1558, opened 2024-11-14
fix: Add missing fields on CozyMetadata for io.cozy.files

cballevre:
By typing inside cozy-drive, I found out that `io.cozy.files` have extra fields inside their `cozyMetadata`. Their are documented inside cozy-doctype but not inside cozy-client ([doc](https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.files.md#cozymetadata)). This commit aims to correct this for future use


paultranvan (approved):
Nice, thank you!

PR #1562, opened 2024-11-20
feat: Remove react-inspector from dependencies

cballevre:
This package was only used for cozy-devtools. Since it was extracted in its own package, I'm removing the dependency

Note: This also allows me to test the new CI process


paultranvan (approved):
Let's see this new CI :)

PR #1564, opened 2024-11-21
chore: Avoid running the CI twice after publish

cballevre:
Lerna has a two-step process. After checking the master branch and publishing the packages. It publishes a commit with the tag of the new version and the modification of the packages.json for use in the next merge. This new commit would trigger the CI again, but this modification avoids this

**Note:** The second commit is also to test the CI working properly


paultranvan (approved):
Nice! :+1:

PR #1569, opened 2024-12-05
Avoid forcing useless renders

paultranvan:
## Issue 

We noticed that any query or mutation resulted in many unexpected render
by the apps, even when there is no new data.  This is particularly
problemetic in environement with low ressources such a the mobile: we
measured strong bottlenecks because of this, making the main thread
overloaded and resulting in not smooth experience for the user.

Each time a i`useQuery` is made, it is saved as a listener in redux,
through the useSelector hook. We provide the result of
`getQueryFromState` to this hook to evaluate whether or not the state
changed. By default, this evaluation is a referential equality check.
See https://react-redux.js.org/api/hooks#useselector

> When an action is dispatched, useSelector() will do a reference
comparison of the previous selector result value and the current result
value. If they are different, the component will be forced to re-render.

The issue here is that the `getQueryFromState` function always return a
new object. Therefore, the reference comparison will never be true.

And because the dispatch will make the evaluation of all listeners, it
means any query will force the re-render to all components having a
useQuery, leading to performance issues.

## Solution

We now have our own comparison function, taking advantage of our store
specifities. This is passed to the `useSelector` hook to determine
whether or not two queries have changed and if a render should be made.

## Impact

With this improvement, we made some measures:

- Display home page of Drive
  - [BEFORE] 6899 ms
  - [AFTER] 3590 ms
  - => x1.9 boost
- Upload 100 files of 1K
  - [BEFORE] 107446 ms
  - [AFTER] 18395 ms
  - => x5.8 boost

Capture d’écran du 2024-12-03 15-49-34

Capture d’écran du 2024-12-04 11-09-34

### DIFF
--- a/docs/api/cozy-client/README.md
+++ b/docs/api/cozy-client/README.md
@@ -1045,7 +1045,7 @@ Retrieve intance info like context, uuid, disk usage etc
 
 *Defined in*
 
-[packages/cozy-client/src/hooks/useQuery.js:94](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/hooks/useQuery.js#L94)
+[packages/cozy-client/src/hooks/useQuery.js:95](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/hooks/useQuery.js#L95)
 
 ***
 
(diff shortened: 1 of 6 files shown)


packages/cozy-client/src/CozyClient.js
```diff
@@ -1193,7 +1193,7 @@ client.query(Q('io.cozy.bills'))`)
           if (queryDef instanceof QueryDefinition) {
             definitions.push(queryDef)
           } else {
-            documents.push(queryDef)
+            documents.push(doc)
```
Ldoppea, line 1196: Is this part of the fix? I don't understand the impact compared to the previous implementation

PR #1574, opened 2024-12-17
feat: Add PerformanceApi to CozyClient

Ldoppea:
We want to be able to measure CozyClient's performances in web apps but also in mobile apps

To make this possible we introduce the PerformancesAPI (**full documentation available in the PR**)

This API should be injected in CozyClient's constructor, then CozyClient will be able to use it to measure performances

This API is meant to wrap platforms' native APIs. On web browsers we wrap the well known Performance API, but on mobile app we may want to use the implementation from react-native-performances

The CozyClient provides the web implementation, but other implementation should be implemented by consuming apps and then injected in the CozyClient's constructor

If no implementation is injected in the constructor, then the `defaultPerformance` implementation is used. This implementation does nothing, so there is no impact on web apps by default

Here is the result when using the browser's Performance devtools :
performances_devtools

Related link:
https://developer.mozilla.org/en-US/docs/Web/API/Performance

____

### TODO:

- [x] Add documentation for the new API
- [x] Add link to the Flagship implementation to the `Other platforms` section of the documentation


docs/performances.md
```diff
@@ -0,0 +1,169 @@
...
+async query(queryDefinition, { update, executeFromStore, ...options } = {}) {
+  const markName = client.mark(`Some mark name`)
+
+  console.log(markName) // "42 Some mark name"
+}
+```
+
+### Measure method
+
+The previously described `mark(markName)` method can be used to place some temporal marks in the code. The `measure(options)` method can be used to measure the elasped time since a specified temporal mark.
```
paultranvan, line 71: s/elasped/elapsed

docs/performances.md
```diff
@@ -0,0 +1,169 @@
...
+    })
+  }
+}
+```
+
+## Adding measurements
+
+On initial implementation, we chose to measure some parts of the code that we estimated useful and with potential high impact on performances (i.e. queries, store management, pouch management etc.)
+
+We excpect to need more measurements in the future but we need to care to keep the balance beetween having useful measurements and to not visualy overload the measurement graphs
```
paultranvan, line 121: s/excpect/expect

docs/performances.md
```diff
@@ -0,0 +1,169 @@
...
+
+The API is meant to be simple to use so measurements can be quickly added temporarily in development environements
```
paultranvan, line 123: s/added temporarily/temporarily added

paultranvan: s/environements/environments

docs/performances.md
```diff
@@ -0,0 +1,169 @@
...
+![Performance devtools screenshot](./performances_devtools.png)
+
+## Other platforms
+
+No implementation is provided by CozyClient for other platforms. So each consuming app needs to provide their own implementation of the API
+
+The API interface is described in the JSDoc
+
+Example of custom implementations:
+- Flagship App: [](TODO ADD LINK)
```
paultranvan, line 161: Will we think of this? :smile:

docs/performances.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Production environment
+
+By default, if no API is injected, then CozyClient will use the `defaultPerformanceAPI` implementation that does nothing
+
+The consuming app is then responsible to inject an implementation or keep the default one that doesn't do any profiling, regarding if the app is built for development environments or for production
+
+PerformancesAPI seems to have an impact on performances light enough to be kept on production. However, we suggest to disable them in that environement if possible
```
paultranvan, line 169: s/environement/environment

paultranvan: > light enough to be kept on production

Are we sure of this? I guess it's ok for a single measure, but I would assume a significative impact overall, since there can be a lot of measures during an app data lifecycle

docs/performances.md
```diff
@@ -0,0 +1,169 @@
+# Performances analysis
```
paultranvan, line 1: This doc is really great!
One small remark is that you seem to really dislike ending dots at the end of sentences :laughing:

paultranvan:
Some remarks, but excellent work overall  :clap:

packages/cozy-client/src/store/index.js
```diff
@@ -49,33 +49,65 @@ export class StoreProxy {
...
-const combinedReducer = (state = initialState, action) => {
+const combinedReducer = performanceApi => (state = initialState, action) => {
+  const markName = performanceApi.mark('combinedReducer')
+
   if (action.type == RESET_ACTION_TYPE) {
+    performanceApi.measure({
+      markName: markName,
+      measureName: `${markName} reset`,
+      category: 'CozyClientStore'
+    })
```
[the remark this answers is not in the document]
Ldoppea: The idea was to have all execution path measured even if some are instant so the output is more consistent.

For example, if there 4 methods calls that should call the reducer, then I prefer to have 4 performance entry even if some of them are very small, instead of having only some of them and wondering why some calls are not logged (is that due to an exception? is the reducer called? etc).

But for this particular example, the `reset` is rare enough so the example I explained is not relevant. So have no objection to remove it.

docs/performances.md
Ldoppea: For sure, the task is on the PR description's todo list, so I should not forget to do it (I won't merge this without adding the link).

docs/performances.md
Ldoppea: My guess is that, because the API can measure durations as small as a few μs, on high level methods this should not be perceptible. But this may be different on some API for other platforms. Also I did not find any info about it so this is just a guess.

But that's why I suggested to disable it.

docs/performances.md
Ldoppea: Fixed [here](https://github.com/cozy/cozy-client/compare/13bb965aca7d1ced675c7e4fa5bc7b0fc6e3bd4c..ee1b801325c7d7db3ca12d34712541c1b9b95165)

docs/performances.md
Ldoppea: Fixed [here](https://github.com/cozy/cozy-client/compare/13bb965aca7d1ced675c7e4fa5bc7b0fc6e3bd4c..ee1b801325c7d7db3ca12d34712541c1b9b95165)

docs/performances.md
Ldoppea: Fixed [here](https://github.com/cozy/cozy-client/compare/13bb965aca7d1ced675c7e4fa5bc7b0fc6e3bd4c..ee1b801325c7d7db3ca12d34712541c1b9b95165)

packages/cozy-client/src/store/index.js
```diff
@@ -49,33 +49,65 @@ export class StoreProxy {
...
     return initialState
   }
   if (!isQueryAction(action) && !isMutationAction(action)) {
+    performanceApi.measure({
+      markName: markName,
+      measureName: `${markName} query or mutate`,
+      category: 'CozyClientStore'
+    })
```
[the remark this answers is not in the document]
Ldoppea: Fixed [here](https://github.com/cozy/cozy-client/compare/ee1b801325c7d7db3ca12d34712541c1b9b95165..fabe643260d358b4cd558200d45dec15387e476b)

I renamed it `${markName} no query nor mutate action`

packages/cozy-client/src/store/index.js
```diff
@@ -49,33 +49,65 @@ export class StoreProxy {
...
+      nextDocuments,
+      haveDocumentsChanged
+    )
   }
+
+  performanceApi.measure({
+    markName: markName,
+    measureName: `${markName} default`,
+    category: 'CozyClientStore'
+  })
```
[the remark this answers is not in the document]
Ldoppea: Fixed [here](https://github.com/cozy/cozy-client/compare/ee1b801325c7d7db3ca12d34712541c1b9b95165..fabe643260d358b4cd558200d45dec15387e476b)

I renamed it `${markName} queries` but I'm wondering if this is correct too? Those can be queries or mutation? Does "queries" implies this can be also mutation?

packages/cozy-client/src/store/queries.js
```diff
@@ -166,11 +172,17 @@ const query = (state = queryInitialState, action, documents) => {
...
     case RECEIVE_QUERY_RESULT: {
+      const markName = performanceApi.mark('RECEIVE_QUERY_RESULT')
       const response = action.response
       // Data can be null when we get a 404 not found
       // see Collection.get()
       // but we still need to update the fetchStatus.
       if (!response.data) {
+        performanceApi.measure({
+          markName: markName,
+          measureName: `${markName} with data`,
```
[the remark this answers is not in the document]
Ldoppea: Fixed [here](https://github.com/cozy/cozy-client/compare/ee1b801325c7d7db3ca12d34712541c1b9b95165..fabe643260d358b4cd558200d45dec15387e476b)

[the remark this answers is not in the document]
Ldoppea: I also renamed the next one (line 209) `with object` as there was also a `!` operator I missed

packages/cozy-client/src/store/queries.js
```diff
@@ -220,6 +242,12 @@ const query = (state = queryInitialState, action, documents) => {
         count,
         fetchedPagesCount
       })
+
+      performanceApi.measure({
+        markName: markName,
+        measureName: `${markName} default`,
+        category: 'CozyClientStore'
+      })
```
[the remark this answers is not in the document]
Ldoppea: On my initial tests, this was to have an idea of the frequency of each return path. Not sure if this is still relevant.

packages/cozy-client/src/store/index.js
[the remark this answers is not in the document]
paultranvan: Ok, fair enough for the consistency, ok to keep it like this

docs/performances.md
Ldoppea: Done: https://github.com/cozy/cozy-client/compare/fabe643260d358b4cd558200d45dec15387e476b..2d7973e259ca4c6eb6e1c134e0dad87875518288

PR #1575, opened 2024-12-17
feat: Allow to enforce Stack link on request chain

Ldoppea:
In previous commits we implemented supports for offline mode through CozyPouchLink and FlagshipLink

Since this feature, the StackLink is able to check for internet reachability by using the `isOnline()` method. When no internet is detected, then the request is forwarded to the next link (i.e. the CozyPouchLink)

In some scenario we may want to prevent that behavior. This is the case when we are working on a feature that cannot work without internet access, or when the query's result is expected to contain data that is injected by the cozy-stack (and so that is not available in the local Pouch database)

To make this possible we introduce the `forceStack` parameter that can be set in the query options

When set to `true` then the StackLink will not attempt to forward the request when offline, instead it will still attempt to do the request and throw

We chose to attempt the request instead of directly throwing an error in order to prevent hypothetical scenario where `isOnline()` method returns false even if internet is reachable

BREAKING CHANGE: CozyLink's request methods now takes an additional `options` argument that can be used to enforce usage of Stack link. Although `.request()` is meant to be an internal API, if your code calls it, you'll want to introduce a new argument for `options`. This is also the case if you instanciate a new CozyLink with a handler argument for `request`

Before:
```js
// Initialization
new CozyLink((operation, result = '', forward) => {
  return forward(operation, result + 'foo')
})

// Call
link.request(operation)

// Call with result and forward
link.request(operation, null, () => { /* do stuff */ })
```

After:
```js
// Initialization
new CozyLink((operation, options, result = '', forward) => {
  return forward(operation, options, result + 'foo')
})

// Call
link.request(operation, options)

// Call with result and forward
link.request(operation, options, null, () => { /* do stuff */ })
```

Related PR: cozy/cozy-flagship-app#1269

### DIFF
--- a/packages/cozy-client/src/StackLink.js
+++ b/packages/cozy-client/src/StackLink.js
@@ -89,19 +89,19 @@ export default class StackLink extends CozyLink {
     this.stackClient = null
   }
 
-  async request(operation, result, forward) {
-    if (this.isOnline && !(await this.isOnline())) {
-      return forward(operation)
+  async request(operation, options, result, forward) {
+    if (!options?.forceStack && this.isOnline && !(await this.isOnline())) {
+      return forward(operation, options)
     }
 
     try {
       if (operation.mutationType) {
-        return await this.executeMutation(operation, result, forward)
+        return await this.executeMutation(operation, options, result, forward)
       }
       return await this.executeQuery(operation)
     } catch (err) {
-      if (isReactNativeOfflineError(err)) {
-        return forward(operation)
+      if (!options?.forceStack && isReactNativeOfflineError(err)) {
+        return forward(operation, options)
       }
       throw err
     }
@@ -138,7 +138,7 @@ export default class StackLink extends CozyLink {
     }
   }
 
-  async executeMutation(mutation, result, forward) {
+  async executeMutation(mutation, options, result, forward) {
     const { mutationType, document: doc, documents: docs, ...props } = mutation
     switch (mutationType) {
       case MutationTypes.CREATE_DOCUMENT:
@@ -182,7 +182,7 @@ export default class StackLink extends CozyLink {
           .collection(DOCTYPE_FILES)
           .upload(props.file, props.dirPath)
       default:
-        return forward(mutation, result)
+        return forward(mutation, options, result)
     }
   }
 }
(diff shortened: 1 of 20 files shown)


packages/cozy-client/src/StackLink.js
```diff
@@ -89,19 +89,19 @@ export default class StackLink extends CozyLink {
     this.stackClient = null
   }
 
-  async request(operation, result, forward) {
-    if (this.isOnline && !(await this.isOnline())) {
-      return forward(operation)
+  async request(operation, options, result, forward) {
+    if (this.isOnline && !(await this.isOnline()) && !options?.forceStack) {
```
paultranvan, line 93: It might be worth it to reverse the conditions here, by evaluation `forceStack` first:  if the `options?.forceStack` is true, there is no need to call the `this.isOnline()`, and thus we save a network call

packages/cozy-client/src/StackLink.js
```diff
@@ -89,19 +89,19 @@ export default class StackLink extends CozyLink {
     this.stackClient = null
   }
 
-  async request(operation, result, forward) {
-    if (this.isOnline && !(await this.isOnline())) {
-      return forward(operation)
+  async request(operation, options, result, forward) {
```
paultranvan, line 92: No jsdoc for the `options` ?

paultranvan (approved):

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -410,7 +410,7 @@ class PouchLink extends CozyLink {
     return !!this.getPouch(impactedDoctype)
   }
 
-  async request(operation, result = null, forward = doNothing) {
+  async request(operation, options, result = null, forward = doNothing) {
```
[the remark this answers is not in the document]
Ldoppea: Fixed [here](https://github.com/cozy/cozy-client/compare/99d56c83d59f30fcbeaf9ff71f6edd4e08d18122..8007276a33cdecf58aeb6e10145fcbdd09c604e0)

packages/cozy-client/src/StackLink.js
Ldoppea: Fixed [here](https://github.com/cozy/cozy-client/compare/8007276a33cdecf58aeb6e10145fcbdd09c604e0..d1b66bb18c0bdf0214f18fc8a3d3f3e7439fce4d)

PR #1585, opened 2025-01-31
fix(deps): Upgrade cozy-pouch-link's parcel

taratatach:
Install latest version of `parcel` to get a more recent version of
  `deasync` (one of its dependencies) as the current version prevents
  installing dependencies in some cases.


paultranvan:
> prevents installing dependencies in some cases.

what are those cases?

paultranvan (approved):

PR #1587, opened 2025-02-07
[DAT-75] Correctly handle refreshToken for dataproxy replication

paultranvan:
This includes:
- Proper refreshToken detection in PouchLink
- Proper handling of the refreshToken in worker env

### DIFF
--- a/packages/cozy-stack-client/src/CozyStackClient.js
+++ b/packages/cozy-stack-client/src/CozyStackClient.js
@@ -191,44 +191,32 @@ class CozyStackClient {
       credentials: 'include'
     }
 
-    if (!global.document) {
-      throw new Error('Not in a web context, cannot refresh token')
-    }
-
     const response = await fetch('/?refreshToken', options)
 
     if (!response.ok) {
       throw new Error(
-        "couldn't fetch a new token - response " + response.statusCode
+        "Couldn't fetch a new token - response " + response.statusCode
       )
     }
     const html = await response.text()
-    const parser = new DOMParser()
-    const doc = parser.parseFromString(html, 'text/html')
-    if (!doc) {
-      throw Error("couldn't fetch a new token - doc is not html")
-    }
-    const appNode = doc.querySelector('div[role="application"]')
-    if (!appNode) {
-      throw Error("couldn't fetch a new token - no div[role=application]")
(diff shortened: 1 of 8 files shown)


packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -191,7 +193,15 @@ class CozyStackClient {
...
     }
 
-    if (!global.document) {
+    const isWorkerEnv =
+      // @ts-ignore
+      typeof WorkerGlobalScope !== 'undefined' &&
+      // @ts-ignore
+      // eslint-disable-next-line no-undef
+      self instanceof WorkerGlobalScope
+    console.log('is worker env : ', isWorkerEnv);
```
Ldoppea, line 202: Do you want to keep that console.log in the commit code?

Ldoppea (approved):

Ldoppea:
In order to ease history tracking, the failing test on the expired token message is a side effect of this previous commit where we introduced errors aggregation 53738770f1d3f3abca34f8b952b2b718e379d326

packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -191,44 +191,36 @@ class CozyStackClient {
...
-      : { ...appNode.dataset }
+    const htmlCozyData = match[1].replace(/&#34;/g, '"')
+    const cozyData = JSON.parse(htmlCozyData)
 
-    let { token } = data
+    let { token } = cozyData
     if (token) {
-      token = token || data.cozyToken
+      token = token || cozyData.cozyToken
     }
```
Ldoppea, line 214: This block can be removed, `data.cozyToken` is from the `appNode.dataset` you removed.

Ldoppea (approved):

packages/cozy-stack-client/src/CozyStackClient.js
Ldoppea: Here is the result when calling `document.querySelector('div[role="application"]').dataset`:
image

The `dataset.cozy` value comes from the `data-cozy` attribute

All the `dataset.cozyXxx` values come from the deprecated `data-cozy-xxx`

PR #1588, opened 2025-02-12
fix: Always use response body as FetchError reason

taratatach:
BREAKING CHANGE: The `reason` attribute of non-JSON fetch errors is no
  longer a string composed of the server response status and status
  text.

  We used to have 2 different kinds of values for the `reason` attribute
  of `FetchError` built in `CozyStackClient`:
  - a string representation of the response body sent by the server (which may contain details about what when wrong, especially for 400 status responses) for JSON requests
  - a string composed of the server response status and status text for non-JSON requests

  In the non-JSON request situation, we lose information by not using
  the server response body as the reason of the `FetchError` which
  prevents users of `cozy-client` from reacting appropriately when
  receiving such an error to their request.
  Therefore we homogenize our `FetchError` instances by always using the
  server response body to build the error `reason`.

  We need to add a new `throwFetchErrors` option to
  `CozyStackClient.fetch` to determine when to throw the error rather
  than emitting it as JSON requests expect the error to be thrown but we
  can't keep reading the response body from the JSON request method to
  build a new error to throw as it's already been read to build the
  emitted error.
  Besides, by doing so, we build the error only once, making sure we get
  the same kind no matter what type of request we make.

### DIFF
--- a/packages/cozy-stack-client/src/CozyStackClient.js
+++ b/packages/cozy-stack-client/src/CozyStackClient.js
@@ -40,6 +40,13 @@ const normalizeUri = uriArg => {
   return uri
 }
 
+const getResponseData = async response => {
+  const contentType = response.headers.get('content-type')
+  const isJson =
+    contentType && /\bapplication\/(vnd\.api\+)?json\b/.test(contentType)
+  return isJson ? response.json() : response.text()
+}
+
 /**
  * Main API against the `cozy-stack` server.
  */
@@ -122,7 +129,7 @@ class CozyStackClient {
    * @throws {FetchError}
    */
   async fetch(method, path, body, opts = {}) {
-    const options = { ...opts }
+    const { throwFetchErrors, ...options } = { ...opts }
     options.method = method
     const headers = (options.headers = { ...opts.headers })
 
@@ -150,10 +157,22 @@ class CozyStackClient {
     try {
       const response = await fetcher(fullPath, options)
       if (!response.ok) {
-        this.emit(
-          'error',
-          new FetchError(response, `${response.status} ${response.statusText}`)
-        )
+        const reason = await getResponseData(response)
+        const err = new FetchError(response, reason)
+
+        // XXX: This was introduced so apps could display errors (e.g. quota
+        // errors) via cozy-ui's `useClientErrors()` hook. At the time of
+        // writing these lines, this is the only use case of this `error`
+        // event.
+        this.emit('error', err)
+
+        // XXX: we introduced `throwFetchErrors` because we've seen some direct
+        // calls to `this.fetch()` and we're not sure how these callers would
+        // react to errors being thrown but it would be best to spend some time
+        // to look at these and, in case they're not prepared, change them to
+        // handle thrown errors.
+        // We could then get rid of `throwFetchErrors`.
+        if (throwFetchErrors) throw err
       }
       return response
     } catch (err) {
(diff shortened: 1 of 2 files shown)


packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -40,6 +40,12 @@ const normalizeUri = uriArg => {
   return uri
 }
 
+const getResponseData = async response => {
+  const contentType = response.headers.get('content-type')
+  const isJson = contentType && contentType.indexOf('json') >= 0
```
paultranvan, line 45: I know you copy/pasted this code, but as you are here, I would do an exact `application/json` match here, to prevent any false positive, like "application/kindofjsonbutnotreally"

packages/cozy-stack-client/src/CozyStackClient.js
```diff
@@ -268,14 +278,9 @@ class CozyStackClient {
         body = JSON.stringify(body)
       }
     }
+    clonedOptions.throwFetchErrors = true
```
paultranvan, line 283: What is the purpose of this option if this is always set to true? 
Besides, I don't think we should make it true by default, this is quite a breaking change

paultranvan (commented):
This PR looks good and legitimate but I'm a bit worried about this BC, as it might have some deep impact for apps/libs reading the error messages

packages/cozy-stack-client/src/CozyStackClient.js
taratatach: We should probably match `application/vnd.api+json` as well though. What do you think?

packages/cozy-stack-client/src/CozyStackClient.js
taratatach: This is not where the breaking change is. This option is always set to `true` only in `fetchJSONWithCurrentToken` as this function was previously throwing errors (although the underlying call to `CozyStackClient.fetch` would also emit them).

packages/cozy-stack-client/src/CozyStackClient.js
paultranvan: Absolutely!

taratatach:
> This PR looks good and legitimate but I'm a bit worried about this BC, as it might have some deep impact for apps/libs reading the error messages

Yes I understand that. The other solution I see is to change the `reason` of fetch errors only when thrown but that would make the API quite unpredictable in my opinion.
How can we go forward? Take a look at error handling in all our apps?

packages/cozy-stack-client/src/CozyStackClient.js
paultranvan: Yes but the vast majority our fetch calls go though `fetchJSON`, calling `fetchJSONWithCurrentToken`, calling `fetch`. Thus, forcing `throwFetchErrors = true` in `fetchJSONWithCurrentToken`, will break the current error emit behaviour

taratatach: Well yes, but I was expecting apps calling `fetchJSON` to actually catch the thrown errors (otherwise we would see broken behaviors all the time :thinking: )

taratatach: I could leave the `emit` call even when `throwFetchErrors` is `true` though and throw after that depending on the value of `throwFetchErrors`.

paultranvan: > I could leave the emit call even when throwFetchErrors is true though and throw after that depending on the value of throwFetchErrors.

Sounds great!

PR #1589, opened 2025-02-20
feat: Ensure previous replication are done 

paultranvan:
If an app wants to trigger replication at different event handling, it
is nice to have some control on it, and prevent multiple parallel
replication processes.
Thus, we introduce a `waitForReplications` param, which is set to true
by default.
Thanks to this, we wait for the previous replication process to complete
before running the new one

### DIFF
--- a/packages/cozy-pouch-link/src/CozyPouchLink.js
+++ b/packages/cozy-pouch-link/src/CozyPouchLink.js
@@ -47,6 +47,7 @@ const addBasicAuth = (url, basicAuth) => {
 }
 
 const sanitized = doc => omit(doc, '_type')
+
 export const getReplicationURL = (uri, token, doctype) => {
   const basicAuth = token.toBasicAuth()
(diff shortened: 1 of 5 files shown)


packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -122,7 +123,9 @@ class PouchLink extends CozyLink {
 
     /** @private */
     this.startReplicationDebounced = debounce(
-      this._startReplication,
+      options => {
+        return this._startReplication(options)
+      },
```
Ldoppea, line 128: This change seems to be equivalent to previous code, is it necessary?

Ldoppea (approved):

packages/cozy-pouch-link/src/CozyPouchLink.js
paultranvan: Not really, I changed it for testing, and I kind of prefer explicit code. I reverted it

PR #1590, opened 2025-02-21
Add shared drive option when creating a sharing

zatteo:
Add shared drive option when creating a sharing. I also included some refactoring too.

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -1341,16 +1341,16 @@ class FileCollection extends DocumentCollection {
   }
 
   /**
-   * Get the Shared Drives directory or create it if it does not exist.
+   * Get or create the Shared Drives directory it if it does not exist.
    *
    * The Shared Drives directory is a special directory :
    * - its _id is io.cozy.files.shared-drives-dir
-   * - it contains Nextcloud drives
-   * - it contains shared drives
+   * - it can contains Nextcloud shortcuts
+   * - it can contains shared drives
    *
    * @returns {IOCozyFolder} Shared Drives directory
    */
-  getSharedDrivesDirectoryOrCreate() {
+  getOrCreateSharedDrivesDirectory() {
     return this.stackClient.fetchJSON('POST', '/files/shared-drives')
   }
 }
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -1339,6 +1339,20 @@ class FileCollection extends DocumentCollection {
 
     return { newLastSeq, pending, results }
   }
+
+  /**
+   * Get the Shared Drives directory or create it if it does not exist.
+   *
+   * The Shared Drives directory is a special directory :
+   * - its _id is io.cozy.files.shared-drives-dir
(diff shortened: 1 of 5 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -1339,6 +1339,20 @@ class FileCollection extends DocumentCollection {
...
+   * Get the Shared Drives directory or create it if it does not exist.
+   *
+   * The Shared Drives directory is a special directory :
+   * - its _id is io.cozy.files.shared-drives-dir
+   * - it contains Nextcloud drives
+   * - it contains shared drives
+   *
+   * @returns {IOCozyFolder} Shared Drives directory
+   */
+  getSharedDrivesDirectoryOrCreate() {
```
paultranvan, line 1353: I would prefer `getOrCreateSharedDrivesDirectory` , which is a more consistent naming, and I think less misleading, as the `create` at the name could easily be missed.
If yes, is it documented somewhere? This is not very intuitive from my point of view

packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -87,6 +87,7 @@ class SharingCollection extends DocumentCollection {
    * @param {Array<Recipient>=} params.readOnlyRecipients Recipients to add to the sharings with only read only access
    * @param {boolean=} params.openSharing If someone else than the owner can add a recipient to the sharing
    * @param {string=} params.appSlug Slug of the targeted app
+   * @param {boolean=} params.sharedDrive If the sharing is a shared drive
```
paultranvan, line 94: A "shared drive"  must be interpreted as a "sharing by federation", in our terminology ?

paultranvan (approved):
Some remarks about naming, but looks good overall :+1:

packages/cozy-stack-client/src/SharingCollection.js
zatteo: Exactly. Is is also the naming used on cozy-stack.

packages/cozy-stack-client/src/FileCollection.js
zatteo: Agree about renaming to `getOrCreateSharedDrivesDirectory`. I do not understand the last part of your comment ?

packages/cozy-stack-client/src/SharingCollection.js
paultranvan: Is it documented somewhere? This naming is not very intuitive from my point of view

packages/cozy-stack-client/src/FileCollection.js
paultranvan: The last part of the comment does not make any sense, because I somehow mixed up my comments. This was meant to be here: https://github.com/cozy/cozy-client/pull/1590#discussion_r1967603296

packages/cozy-stack-client/src/SharingCollection.js
zatteo: Is it being documented here https://github.com/cozy/cozy-stack/pull/4511

For the moment we (@nono and me) still think that shared drive is the best terminology. Federated sharing makes us think about federated services like Mastodon.

But we are still open to change.

Issue #1591, opened 2025-02-26
In-memory store evaluation is slow for ids queries


paultranvan:
When running a query on ids directly on the store, the performance is surprisingly slow when the store is big, e.g. `client.query(Q(doctype).getByIds(ids), { executeFromStore})`.  
A query retrieving 362 files from 25K docs in store takes +1000ms to be executed. 

This is because of how we evaluate the sift queries: we iterate over all doctype collection, and evaluate the predicate on each doc. It would be much more efficient to access the documents directly from the state , e.g. `state[doctype][id]`.
This way, we would have a O(1) performance rather than an expensive O(n).

paultranvan:
https://github.com/cozy/cozy-client/pull/1592 fixes it partially: this is now fine for `getByIds(ids), { executeFromStore}` but we did not improved the queries re-evaluation through sift, but this is less critical as the evaluation is made on incoming docs rather than the whole store. So let's close this issue.

PR #1592, opened 2025-02-26
feat: Improve store query performance for id queries

paultranvan:
See https://github.com/cozy/cozy-client/issues/1591
For a query returning 352 docs by ids from a store containing 25K docs, we measure 49ms, against +1000ms before, thanks to k * O(1) evaluation vs O(n)

This is notably useful for search use-case: all the docs are loaded in store, and we need each query to be as fast as possible

### DIFF
--- a/packages/cozy-client/src/store/documents.js
+++ b/packages/cozy-client/src/store/documents.js
@@ -99,7 +99,14 @@ const documents = (state = {}, action) => {
 
 export default documents
 
-// selector
+/**
+ * Get document from state, by its id
+ *
+ * @param {import('../types').DocumentsStateSlice} state The documents' slice
+ * @param {string} doctype The doctype to target
+ * @param {string} id The document id to get
+ * @returns {import('../types').CozyClientDocument} the document found by its id
+ */
 export const getDocumentFromSlice = (state = {}, doctype, id) => {
   if (!doctype) {
     throw new Error(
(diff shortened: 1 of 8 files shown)


packages/cozy-client/src/store/documents.js
```diff
@@ -131,6 +131,29 @@ export const getDocumentFromSlice = (state = {}, doctype, id) => {
   return state[doctype][id]
 }
 
+export const getDocumentsFromSlice = (state = {}, doctype, ids) => {
+  if (!doctype) {
+    throw new Error(
+      'getDocumentFromSlice: Cannot retrieve documents with undefined doctype'
```
taratatach, line 137:
```suggestion
      'getDocumentsFromSlice: Cannot retrieve documents with undefined doctype'
```

packages/cozy-client/src/store/documents.js
```diff
@@ -131,6 +131,29 @@ export const getDocumentFromSlice = (state = {}, doctype, id) => {
...
+    )
+  }
+  if (!ids || ids.length < 1) {
+    throw new Error(
+      'getDocumentFromSlice: Cannot retrieve documents with undefined ids'
```
taratatach, line 142:
```suggestion
      'getDocumentsFromSlice: Cannot retrieve documents with undefined ids'
```

packages/cozy-client/src/store/documents.js
```diff
@@ -131,6 +131,29 @@ export const getDocumentFromSlice = (state = {}, doctype, id) => {
...
+    )
+  }
+  if (!state[doctype]) {
+    if (process.env.NODE_ENV !== 'production') {
+      logger.info(
+        `getDocumentFromSlice: ${doctype} is absent from the store's documents. State is`,
```
taratatach, line 148:
```suggestion
        `getDocumentsFromSlice: ${doctype} is absent from the store's documents. State is`,
```

packages/cozy-client/src/store/documents.js
```diff
@@ -131,6 +131,29 @@ export const getDocumentFromSlice = (state = {}, doctype, id) => {
...
+        state
+      )
+    }
+    return null
```
taratatach, line 152: I would suggest returning an empty array to always have the same return type.

paultranvan: That is correct, as a punishement, I added the jsdoc to both `getDocumentFromSlice` and `getDocumentsFromSlice`

packages/cozy-client/src/store/documents.js
```diff
@@ -99,7 +99,14 @@ const documents = (state = {}, action) => {
 
 export default documents
 
-// selector
+/**
+ * Get documents from state, by their ids
```
taratatach, line 103:
```suggestion
 * Get document from state, by its id
```

taratatach (approved):
There's a small doc modification to make to otherwise LGTM.

packages/cozy-client/src/store/documents.js
paultranvan: Uh, I have some copy paste lazyness today... Thanks for your attention

PR #1594, opened 2025-02-27
fix: Correctly handle document's store updates

paultranvan:
We used to create a new reference on the `documents` state for any query.
This is problematic because we do a referential comparison on the updated `documents` state to evaluate if there is any update on it, resulting on all store's queries re-evaluation.
So, basically, we were doing a full queries re-evaluation for *any* query execution, even if the same query is run several time with no change.
This is particularly problematic when there are a lot of queries and/or
queries with many documents returned, as their re-evaluation will be
costful. Moreover, the document merging itself is costful.

The comparison between old and new docs is crucial for performances, so
we did some performances measures: we took a worst case where we fill
the store with 25K docs with a query and run again the same query with
no document change.
What we measure here is the time taken in the `extractAndMergeDocument` method:

- **Before: 73892 ms**
- Compare with lodash isEqual: 47530 ms
- Compare with stringify + djb2 hash: 19785 ms
- **Compare with fast-deep-equal: 2404 ms (used solution)**
- Compare with JSON.stringify: 1899 ms
- Compare with rev (for the record): 123ms

ℹ️  The JSON.stringify comparison is faster, but it does not guarantee
key order: 2 documents with same value but different order will be seen
as different. For safety, we prefer fast-deep-equal, slighly slower, but
more reliable.

ℹ️  The rev measurement is here because it would be the ideal solution:
assume any document change implies a new revision, as it is made with
CouchDB / PouchDB. However, this is more complicated in practice:
- Some docs returned by the stack do not have revision
- Some docs returned by the stack might have changed, with the same
  revision: for instance the thumbnails links for files change over
time.
- If we have a partial doc query, i.e. a query with `.select(fields)`,
  and then another query returning a common doc, the content might be
different between the 2 queries, with the same doc revision.
As this is more complicated to deal with, we do not rely on revision
yet.

We also improved the document merging. For 1K docs to merge: 
- Before: 1626ms
- After: 278ms

### DIFF
--- a/packages/cozy-client/src/store/documents.js
+++ b/packages/cozy-client/src/store/documents.js
@@ -1,11 +1,8 @@
-import keyBy from 'lodash/keyBy'
 import get from 'lodash/get'
-import isEqual from 'lodash/isEqual'
 import omit from 'lodash/omit'
 import merge from 'lodash/merge'
-
 import logger from '../logger'
-
+import fastEqual from 'fast-deep-equal'
 import { isReceivingData } from './queries'
 import { MutationTypes } from '../queries/dsl'
 import { isReceivingMutationResult } from './mutations'
@@ -29,7 +26,7 @@ const storeDocument = (state, document) => {
 
   const existingDoc = get(state, [type, properId(document)])
 
-  if (isEqual(existingDoc, document)) {
+  if (fastEqual(existingDoc, document)) {
     return state
   } else {
     return {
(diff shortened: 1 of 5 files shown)


Crash--:
> This is particularly problematic when there are a lot of queries 

This is why we should add a garbage collector for queries! You've the paper for that :p.

packages/cozy-client/src/store/documents.js
```diff
@@ -221,6 +221,7 @@ export const extractAndMergeDocument = (newData, updatedStateWithIncluded) => {
   }
 
   let haveDocumentsChanged = false
+  const deeppmerge = deepmerge()
```
zatteo, line 224: I think we should create this deepmerge() object outside `extractAndMergeDocument` to avoid creating a new one for each `extractAndMergeDocument` call ?

packages/cozy-client/src/store/documents.js
```diff
@@ -188,38 +185,65 @@ export const getCollectionFromSlice = (state = {}, doctype) => {
...
+  let haveDocumentsChanged = false
+
+  newDocsMap.forEach((newDoc, key) => {
+    const id = key
+    const currentDoc = mergedState[doctype][id]
+    if (!currentDoc) {
+      mergedState[doctype][id] = newDoc
+      haveDocumentsChanged = true
+    } else if (!fastEqual(currentDoc, newDoc)) {
+      // XXX Here we check if the doc has changed. Note this would be way faster
```
zatteo, line 233: nit: What are theses XXX ?

packages/cozy-client/src/store/documents.js
```diff
@@ -188,38 +185,65 @@ export const getCollectionFromSlice = (state = {}, doctype) => {
...
+      //    fields added by the stack
+      mergedState[doctype][id] = merge({}, mergedState[doctype][id], newDoc)
+      haveDocumentsChanged = true
     }
   })
-
-  return mergedData
+  // XXX If there is any state change, the updated state should have a new reference,
+  // as we heavily rely on referential equality.
+  // Conversely, if there is no change, the same reference should be kept
```
zatteo, line 246: nit: Maybe we could add tests that check this to avoid breaking this core behavior : 
- a state change => new reference
- no state change = > same reference

zatteo (approved):

packages/cozy-client/src/store/documents.js
paultranvan: This is something I took from @nono , I'm not sure if it means anything, but I interpret it as "this is tricky and worth reading"

packages/cozy-client/src/store/documents.js
paultranvan: You're right, this is a core function of the store handling and there is currently no test, so this is not a nit :wink:

packages/cozy-client/src/store/documents.js
nono: Yes, it means that it requires attention, and often, it is dirty (like we do something weird to avoid a bug in a lower level library or we have to do that for an historical reason).

Ldoppea (approved):

packages/cozy-client/src/store/documents.js
paultranvan: Added in [b0714ba](https://github.com/cozy/cozy-client/pull/1594/commits/b0714ba2c79231d962ec02b20e92e29edb3be933)

packages/cozy-client/src/store/documents.js
paultranvan: Yes, done in [fed634c](https://github.com/cozy/cozy-client/pull/1594/commits/fed634c0b9aefd4ba0d907edb7f0c665c0f26573)

PR #1595, opened 2025-02-28
[WIP] feat: Improve performances

Ldoppea:


packages/cozy-client/src/CozyClient.js
```diff
@@ -1157,7 +1157,18 @@ client.query(Q('io.cozy.bills'))`)
...
+          return false
+        }
+
+        if ((!document.meta?.rev && !document._rev) || enforce) {
+          return true
+        }
+
+        return false
+      })
+      for (const document of documentsToPersist) {
```
Crash--, line 1199: Is it really better? OK you don't call a method if not needed, but now you loop twice

packages/cozy-pouch-link/src/jsonapi.js
```diff
@@ -1,32 +1,40 @@
...
+    doc.relationships.referenced_by = doc.referenced_by
+  } else {
+    doc.relationships = {
+      referenced_by: doc.referenced_by
     }
   }
-  if (normalizedDoc.rev) {
-    delete normalizedDoc.rev
+  if (doc.rev) {
+    delete doc.rev
```
Crash--, line 30: setting to undefined is way more performant than calling delet(https://jsperf.app/delete-vs-undefined-vs-null/10 I don't like the `void 0` stuff)

packages/cozy-client/src/CozyClient.js
paultranvan: This is actually significantly better. After a big query retrieving 25K docs, the `persistVirtualDocuments` takes: 
- Before: 609ms
- After: 20ms

I wonder if this is because of the event loop overwhelmed by too many calls

Crash--: Oh shit, I didn't saw that the `persistVirtualDocument` was async... This is why.

nit: You can now remove the check within the `persistVirtualDocument` method since should only call it with the docs to persist ^^.

packages/cozy-pouch-link/src/jsonapi.js
paultranvan: Good to know! I did the change in another PR: https://github.com/cozy/cozy-client/pull/1596/commits/2b89a635d12441ddc445cf5d44ce7ddd77232723

PR #1598, opened 2025-03-25
Add platform option to directly query SQLite database

paultranvan:
We noticed some performances issues with PouchDB, particularly when using Mango indexes.
We also noticed that direct SQLite accesses are very fast in a react-native environement, and using native SQLite indexes is very efficient.
Therefore, we decided to offer the possibility to directly query SQLite rather than using PouchDB API, in a completely transparent way. It can be simply done like this: 

```
import { default as PouchLink, SQLiteQuery } from 'cozy-pouch-link';

const pouchLinkOpts = {
 doctypes: ['io.cozy.todos'],
 platform: {
   queryEngine: SQLiteQuery
 }
}
const pouchLink = new PouchLink({pouchLinkOpts})
const client = new CozyClient({uri, token, links: [pouchLink]})

```
Here are some performances measures, from a react-native env, on a OnePlus Nord 4 Android device. The database has 8600 documents. The given time is measured inside a `client.query` call, with no store (except for the redux store measures).
    
    |                       | PouchDB  | Sqlite | Redux store | Stack   |
    | --------------------- | -------- | ------ | ----------- | ------- |
    | 1 doc                 | 10 ms    | 10 ms  | 2 ms        | 175 ms  |
    | 100 docs (allDocs)    | 72 ms    | 34 ms  | 3 ms        | 198 ms  |
    | 1000 docs (allDocs)   | 169 ms   | 74 ms  | 10 ms       | 661 ms  |
    | 8664 docs (allDocs)   | 1073 ms  | 830 ms | 88 ms       | 2300 ms |
    | 100 docs (find)       | 162 ms   | 25 ms  | 6 ms        | 860 ms  |
    | Index creation (find) | 83803 ms | 386 ms | -           | -       |

ℹ️ The allDocs method is a sequential scan of the db
ℹ️ The find method uses a JSON index. The measures do not include the index creation time.
ℹ️ The redux store evaluation is made on a store already filled with docs
ℹ️ The stack evaluation includes network requests


packages/cozy-pouch-link/src/db/pouchdb/getDocs.js
```diff
@@ -0,0 +1,63 @@
...
+
+export const getDocs = async (db, fct, queryParams = {}) => {
+  // allDocs return an error when limit is null
+  if (queryParams.limit === null) delete queryParams.limit
+
+  const limit = queryParams.limit
+  const field = fct === 'allDocs' ? 'rows' : 'docs'
+
+  if (isAdapterBugged(db.adapter)) {
+    // FIXME: to remove?
```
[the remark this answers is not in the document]
paultranvan: There is not much to say, because I don't have the context neither, but I added some explanation

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -53,14 +59,21 @@ class PouchManager {
...
+          new this.PouchDB(
+            getDatabaseName(this.options.prefix, doctype),
+            pouchOptions
+          )
+        ]
+      })
     )
+    Object.keys(this.pouches).forEach(dbName => {
+      this.dbQueryEngines.set(dbName, this.pouches[dbName])
+    })
```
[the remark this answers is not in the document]
paultranvan: That was a good catch, and fixed in 40cc887dd1248cf2d53b91b0fdeb4774cd028fd5

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -235,6 +232,8 @@ class PouchLink extends CozyLink {
       }
     }
 
+    this.queryEngine = this.options.platform?.queryEngine || PouchDBQueryEngine
```
Ldoppea, line 235: This one seems not used anymore, you can directly use the platform member inside of PouchManager.

Also PouchManager already defaults to `PouchDBQueryEngine`, so no need to handle it here.

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -53,14 +63,30 @@ class PouchManager {
...
+        ]
+      })
     )
+
+    Object.keys(this.pouches).forEach(dbName => {
+      const doctype = getDoctypeFromDatabaseName(dbName)
+      if (this.queryEngine === PouchDBQueryEngine) {
+        // Use existing pouch instance
+        const engine = new PouchDBQueryEngine(this, doctype)
+        engine.setDB(this.pouches[dbName])
```
Ldoppea, line 83: The `PouchDbQueryEngine` already retrieves the db when using `openDb`. So there should be no need to call `setDB` here.

So this can be replaced by:
```js
        const engine = new PouchDBQueryEngine(this, doctype)
        engine.openDB(dbName)
        this.dbQueryEngines.set(dbName, engine)
```
Which is nearly the same code in `setQueryEngine`:
```js
  setQueryEngine(name, doctype) {
        const engine = new this.options.queryEngine(this, doctype)
        engine.openDB(name)
        this.dbQueryEngines.set(name, engine)
        return engine
  }
```
And because you are in a `if (this.queryEngine === PouchDBQueryEngine)` then you can remove the if and mutualize the code here.

packages/cozy-pouch-link/src/db/pouchdb/pouchdb.js
```diff
@@ -0,0 +1,110 @@
+import { createIndex, getIndexFields, getIndexName } from '../../mango'
+import DatabaseQueryEngine from '../dbInterface'
+import { isMissingPouchDBIndexError } from '../../errors'
+import { getDocsAndNormalize } from './getDocs'
+
+export default class PouchDBQueryEngine extends DatabaseQueryEngine {
+  constructor(pouchManager, doctype) {
+    super()
+    this.pouchManager = pouchManager
+    this.client = pouchManager?.client
```
Ldoppea, line 10: Why is the `?.` needed here? Seems like a typo (same comment in sql engine)

paultranvan: It was breaking tests. Not a good reason as we can just better mock, but it does not cost to add it

packages/cozy-pouch-link/src/PouchManager.js
paultranvan: Fair enough, I fixed it and moved it directly after the pouch initialization

PR #1604, opened 2025-04-30
feat: FileCollection uses proxy for shared drives

taratatach:
To make changes to files on a shared drive, clients have to call their
  stack via proxied routes (e.g. `/sharings/drives/:id/_changes` for the
  changes feed of the given shared drive).

  Instead of creating a separate collection for that we enable passing
  options to a collection and specifically a `driveId` option to the
  `FileCollection`.
  This presence of a `driveId` option will change the prefix of all
  routes called by the `FileCollection` from `/files` to
  `/sharings/drives/${driveId}`.

### DIFF
--- a/packages/cozy-stack-client/src/FileCollection.js
+++ b/packages/cozy-stack-client/src/FileCollection.js
@@ -194,9 +194,15 @@ const dirName = path => {
  * files associated to a specific document
  */
 class FileCollection extends DocumentCollection {
-  constructor(doctype, stackClient) {
-    super(doctype, stackClient)
+  constructor(doctype, stackClient, options = {}) {
+    super(doctype, stackClient, options)
     this.specialDirectories = {}
+
+    if (options.driveId != null && options.driveId !== '') {
+      this.prefix = `sharings/drives/${options.driveId}`
+    } else {
+      this.prefix = 'files'
+    }
   }
 
   /**
@@ -220,7 +226,7 @@ class FileCollection extends DocumentCollection {
     try {
       resp = await this.stackClient.fetchJSON(
         'POST',
-        uri`/files/_all_docs?include_docs=true`,
+        uri`/${this.prefix}/_all_docs?include_docs=true`,
         {
           keys: ids
         }
@@ -240,7 +246,7 @@ class FileCollection extends DocumentCollection {
   async fetchFindFiles(selector, options) {
     return this.stackClient.fetchJSON(
       'POST',
-      '/files/_find',
+      `/${this.prefix}/_find`,
       this.toMangoOptions(selector, options)
     )
   }
@@ -259,7 +265,7 @@ class FileCollection extends DocumentCollection {
     const { skip = 0 } = options
     let resp
     try {
-      const path = '/files/_find'
+      const path = `/${this.prefix}/_find`
       resp = await this.findWithMango(path, selector, options)
     } catch (error) {
       return dontThrowNotFoundError(error)
@@ -323,7 +329,7 @@ class FileCollection extends DocumentCollection {
     const refs = documents.map(d => ({ id: d._id, type: d._type }))
     const resp = await this.stackClient.fetchJSON(
       'POST',
-      uri`/files/${document._id}/relationships/referenced_by`,
+      uri`/${this.prefix}/${document._id}/relationships/referenced_by`,
       { data: refs }
     )
     return {
@@ -351,7 +357,8 @@ class FileCollection extends DocumentCollection {
     }))
     const resp = await this.stackClient.fetchJSON(
       'DELETE',
-      uri`/files/${document._id || document.id}/relationships/referenced_by`,
+      uri`/${this.prefix}/${document._id ||
+        document.id}/relationships/referenced_by`,
       { data: refs }
     )
     return {
@@ -417,7 +424,7 @@ class FileCollection extends DocumentCollection {
 
     const resp = await this.stackClient.fetchJSON(
       'DELETE',
-      uri`/files/${_id}`,
+      uri`/${this.prefix}/${_id}`,
       undefined,
       {
         headers: {
@@ -440,7 +447,7 @@ class FileCollection extends DocumentCollection {
    * Empty the Trash
    */
   emptyTrash() {
-    return this.stackClient.fetchJSON('DELETE', '/files/trash')
+    return this.stackClient.fetchJSON('DELETE', `/${this.prefix}/trash`)
   }
 
   /**
@@ -452,7 +459,7 @@ class FileCollection extends DocumentCollection {
    *
    */
   restore(id) {
-    return this.stackClient.fetchJSON('POST', uri`/files/trash/${id}`)
+    return this.stackClient.fetchJSON('POST', uri`/${this.prefix}/trash/${id}`)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -194,9 +194,15 @@ const dirName = path => {
  * files associated to a specific document
  */
 class FileCollection extends DocumentCollection {
-  constructor(doctype, stackClient) {
-    super(doctype, stackClient)
+  constructor(doctype, stackClient, options = {}) {
```
zatteo, line 197: nit: Maybe we can type with TS comment this option

zatteo:
I like it like this. It implies that shared drives have exactly same route except the prefix but if it is what will happen 👍

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -194,9 +194,15 @@ const dirName = path => {
...
+    super(doctype, stackClient, options)
     this.specialDirectories = {}
+
+    if (options.driveId != null && options.driveId !== '') {
```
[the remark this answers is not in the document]
taratatach: Ah yes, we could do that too :).

packages/cozy-stack-client/src/FileCollection.js
taratatach: Yes, agreed! I did the minimum I had to to test it and show it to you guys. I plan on adding tests (and types here then).

paultranvan:
I think it's a good idea to avoid creating a new collection :+1: 
I have 2 concerns here: 

1. How can we use this API with a `client.query`, rather than a direct `client.collection('io.cozy.files', { driveId })` ? 
2. The naming of `driveId` might be not ideal, I think I would prefer a more explicit `sharedDriveId`. Or maybe an even more explicit `prefix`? That would have the benefit of being more generic.

About the first point, maybe we could pass the prefix to the collection methods options, and do `client.query(queryDef, { pathPrefix: `/sharings/drives/${driveId}` }) ?

taratatach:
1. I forgot about `client.query` but we could pass the collection's options as argument indeed
2. I thought about using `prefix` but cozy-client abstracts away paths and I'd rather not force developers to know about them for this feature. I'm OK with using a more specific argument name like `sharedDriveId` though

zatteo:
> require the dev to know about the route path ?

I think it's better to avoid this (that's one of the main advantages of cozy-client to avoid the need to know route pathes). I would like that the user of cozy client (using a collection or a query) only has to pass a `driveId` or `sharedDriveId`.

So we would do something like : 
1. for collection : `const fileCollection = client.collection('io.cozy.files', { sharedDriveId: sharedDriveId })`
2. for queries : `Q(io.cozy.files).where({ trashed: true}).fromSharedDriveId(sharedDriveId)` 

And then maybe inside cozy-client inside the file collection, we convert the sharedDriveId to a prefix to keep the code clean and generic inside cozy-client. 

About the 2. (using a query), I do not know what is best between passing it: 
- in query definition `client.query(Q(io.cozy.files).where({ trashed: true}).fromSharedDriveId(sharedDriveId))` 
- in query definition inside the where clause `client.query(Q(io.cozy.files).where({ trashed: true, sharedDriveId: sharedDriveId }))`
- in query options `client.query(Q(io.cozy.files).where({ trashed: true }), { sharedDriveId: sharedDriveId })`

It makes sense for me that it is inside the query definition because it changes the data fetched and not the way we fetch data but I do not know what it implies/if it is possible.

taratatach:
>     * In [d63f6d6](https://github.com/cozy/cozy-client/commit/d63f6d620740e8d50efd2439080cc864736baf31) I had to change DSL method from `sharingId` to `setSharingId`. What do you think about it? It will be the main exposed API.

What about `inSharing(<id>)`? I think it better conveys the idea that the query will be run within the given sharing.

>     * Internally, we sometime use `driveId` and sometimes `sharingId`. Is it relevant to have this two names? I am not sure but I can't figure it out.

I used `sharingId` pretty much everywhere except in `FileCollection` where I preferred to used `driveId` as it is more precise (it's not any kind of file sharing).

packages/cozy-stack-client/src/FileCollection.js
```diff
@@ -79,6 +86,13 @@ import logger from './logger'
  * @property {string} _type - Doctype of the client (i.e. io.cozy.oauth.clients)
  */
 
+/**
+ * Options that can be passed to FileCollection's constructor
+ *
+ * @typedef {object} FileCollectionOptions
+ * @property {string} [driveId] - ID of the shared drive targeted by the collection
```
paultranvan, line 93: Do we have some doc about the shared drive somewhere? Typically from the stack side?

packages/cozy-pouch-link/src/helpers.js
```diff
@@ -21,7 +21,6 @@ helpers.insertBulkDocs = async (db, docs) => {
   return db.bulkDocs(docs, { new_edits: false })
 }
 
-
```
paultranvan: This commit should be squash

packages/cozy-client/src/links/StackLink.spec.js
```diff
@@ -36,31 +39,43 @@ describe('StackLink', () => {
       stackClient.collection().find.mockReset()
       await link.request(query)
       expect(stackClient.collection().find).toHaveBeenCalled()
-      expect(stackClient.collection).toHaveBeenCalledWith('io.cozy.todos')
+      expect(stackClient.collection).toHaveBeenCalledWith(
+        'io.cozy.todos',
+        expect.anything()
+      )
```
paultranvan, line 45: Since this commit actually fix tests from previous commit,  it should be rebased within the commit where it breaks

packages/cozy-client/src/queries/dsl.js
```diff
@@ -366,7 +366,7 @@ class QueryDefinition {
    * @param {string} id - The sharing doc id
    * @returns {QueryDefinition}  The QueryDefinition object.
    */
-  sharingId(id) {
+  setSharingId(id) {
```
paultranvan, line 369: I don't really like this naming TBH. I understand that we technically set the sharingId to the QueryDefinition, but here the semantic should not be on the technical set, but rather from the querying perspective. As a developer, I would not understand doing `Q('io.cozy.files').setById(id)` to query a file by its id.
Likewise, I would prefer `sharingById` here

packages/cozy-stack-client/src/FileCollection.js
taratatach: We do have documentation in `cozy-stack` about shared drives.

packages/cozy-client/src/queries/dsl.js
taratatach: I suggested `inSharing(id)`.

zatteo: Agree with `sharingById` !

paultranvan: Oh @zatteo already called it

packages/cozy-stack-client/src/FileCollection.js
paultranvan: I was wondering why I didn't find any code :) 
 Ok, so the stack's doc link could be added to the jsdoc eventually

packages/cozy-client/src/queries/dsl.js
zatteo: Sorry I answered before seeing @taratatach comment...

I think `inSharing` is more explicit but like you said it is more "functional", and cozy-client API and naming is not functional so maybe we should stick with `sharingById`.

PR #1607, opened 2025-05-07
Correctly handle queries with DataProxy

paultranvan:
This includes various fixes and improvements to correctly handle queries in the DataProxy.
See commit descriptions for more insights

:warning:  It is based on https://github.com/cozy/cozy-client/pull/1606 that should be merged first

### DIFF
--- a/docs/api/cozy-client/classes/CozyClient.md
+++ b/docs/api/cozy-client/classes/CozyClient.md
@@ -73,7 +73,7 @@ Cozy-Client will automatically call `this.login()` if provided with a token and 
 
 *Defined in*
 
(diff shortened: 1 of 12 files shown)


Ldoppea:
> [fix: Do not throw when doc has relationship but not schema](https://github.com/cozy/cozy-client/pull/1607/commits/bda653510e097f82108f7e787b369cbae3df0186)

In the flagship app I did copy the schema from cozy-home. I imagine we may want to discuss about this later in order to handle it in an homogenous way.

Ldoppea (approved):

paultranvan:
> In the flagship app I did copy the schema from cozy-home. I imagine we may want to discuss about this later in order to handle it in an homogenous way.

Yes, I think we should let the apps handle themselves the relationships (by their own cozy-clilent I mean), and avoid the need to let the dataproxy to handle a big schema for all apps. But we can further discuss the implications

PR #1608, opened 2025-05-11
fix: downgrade react-native 0.65.x to 0.63.x

pm-McFly:
react-native 0.65.x is compatible w/ react 17 only


zatteo:
Interesting because in practice we are on react-native 0.73. So even if it is more correct from a dependency point of view to downgrade it does not make really sense. It must be because we use a very small subset of React Native that we did not encounter issues.

Counter proposition 1 : update our apps to React 17

Counter proposition 2 : remove dependency to React Native in cozy-client. I was never comfortable with this dependency. Is there a way to inject the methods relying on React Native in the apps instead of having them here ?

Ldoppea:
@pm-McFly did you encounter an issue with this? Or is this "by principle"?

In cozy-client, the react-native dependency is used only in `.native.js` files and those files are loaded only in our react-native app, it is never loaded in a web environment.

More specifically this is used in `authentication/mobile.native.js`. This specific part does not use any React related mechanism, so we should never encounter any incompatibility between the cozy-client's react version and its react-native version.

pm-McFly:
> @pm-McFly did you encounter an issue with this? Or is this "by principle"?

No don't worry I won't open a discussion about which version *should* be used when or NPM/YARN debate ;)

It's mostly about the fact that I wanted to try-out cozy-ui and got stuck during dependency resolving (cf. linked issue comment with full details ;))

> This specific part does not use any React related mechanism, so we should never encounter any incompatibility between the cozy-client's react version and its react-native version.

Unfortunately this is not understood by the NPM/YARN deps resolution system:
https://github.com/cozy/cozy-ui/pull/2789#issuecomment-2872015684

pm-McFly:
btw after some trials... React version is incorrect..
need more in-depth analysis to figure out the correct one!

pm-McFly:
revert to only downgrade rn version to 0.63.5 (expecting react 16.13.1) -> need to verify other repos for react compat!

zatteo (approved):

zatteo:
Moved here https://github.com/cozy/cozy-client/pull/1609

PR #1610, opened 2025-05-22
feat: Use Prelogin API instead of well-known URL in `rootCozyUrl()` and add new `registrationDetails()` method

Ldoppea:
The `rootCozyUrl` method is responsible to verify if a given URL corresponds to an existing cozy and if yes it returns the root URL for this cozy (no slug, no path etc)

This methods is based on the Cozy's `/.well-known/change-password` route

One of the limitations of this route is when testing a Cozy that uses OIDC for login, then a redirection is done to the OIDC server

This would produce false positive due to the redirection check

This check is necessary because the cozy-stack may redirect deprecated apps to the Home app

To prevent this behavior we can use the Cozy's `/public/prelogin` route instead

This route is a public API endpoint, so it won't redirect in case of OIDC login

Also his route has the benefits of returning a JSON object that contains information about how to login to the Cozy

We want to retrieve the `isOIDC` prop from this JSON

However, instead of modifying the `rootCozyUrl` method return type, which would be a breaking change, we prefer to create a new method `registrationDetails`

This new method is responsible for both checking the validity of the Cozy and also to return its root url and a boolean specifying if the Cozy uses OIDC


packages/cozy-client/src/helpers/urlHelper.js
```diff
@@ -351,3 +363,103 @@ const removeSubdomain = url => {
...
+ * 4. getting the root URL when your user gives you an instance from OIDC partners
+ *
+ *   const details = await registrationDetails(new URL(`https://alice.someoidcpartener.com`))
+ *   // → returns { rootUrl: new URL('https://alice.someoidcpartener.com'), isOIDC: true }
+ *
+ * @param {URL} url The URL from which we'll try to get the root Cozy URL
+ *
+ * @returns {Promise<RegistrationDetails>} The root Cozy URL
+ */
+export const registrationDetails = async url => {
```
zatteo, line 394: nit: Feeling like this method name is missing a verb

zatteo (approved):
Smart way to changes things 👍

packages/cozy-client/src/helpers/urlHelper.js
Ldoppea: I did hesitate to choose a name similar to `rootCozyUrl` or to do something different like `fetchRegistrationDetails`.

I'm open to both.

taratatach (approved):

packages/cozy-client/src/helpers/urlHelper.js
Ldoppea: Done: https://github.com/cozy/cozy-client/pull/1610/commits/3420f60d342776a29ad2b4914a4dfc64835a51b5

PR #1616, opened 2025-06-27
fix: Normalize data for trash-related methods

paultranvan:
The `emptyTrash`, `restore` and `deleteFilePermanently` were not normalizing data.
Furthermore, `deleteFilePermanently` return was not including `data` object, making it non-standard.

BREAKING CHANGE: if you were relying on the `deleteFilePermanently` response, take care of handling the `data` attribute at the root of the response.


zatteo (approved):
nit for next time do nothing here: Breaking chang could be even more simple and concrete for the poor developer that is updating breaking changes, so like "deleteFilePermanently return now `{ data: result }` instead of `result` so replace `const result = deleteFilePermanently()` by `const { data: result } = deleteFilePermanently()`"

PR #1617, opened 2025-06-27
feat: Support multiple create and delete

paultranvan:
We used to support mutliple updates but not create nor delete.  Thus, we now introduce the `createAll` and `destroyAll` API methods, to respectively bulk create and delete a batch of documents.  `updateAll`is also introduced, which is a simple wrapper for `saveAll`, just like `createAll`.

:warning: The `createAll` is currently not supported for stackClient links, but works for pouchLink. 

Adding the associated dsl mutations is also useful for apps that want to handle themselves store actions. For instance, Drive would be able to dispatch a single mutation action for multiple updates, typically after the deletion of a directory. This would improve performances, by avoiding to make a costful dispatch per document.

### DIFF
--- a/docs/api/cozy-client/README.md
+++ b/docs/api/cozy-client/README.md
@@ -50,7 +50,9 @@ cozy-client
 | `ADD_REFERENCED_BY` | `string` |
 | `ADD_REFERENCES_TO` | `string` |
 | `CREATE_DOCUMENT` | `string` |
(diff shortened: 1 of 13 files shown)


paultranvan:
> Do we have a ticket to expose the bulk_docs API to the stack?

For the record, after discussion we find out it's not a stack issue (the API is already exposed), but more a CC collection issue

PR #1622, opened 2025-08-19
feat: Queue DataProxy requests when it is not ready

paultranvan:
The DataProxy needs some time to be initialized and answer requests. However, apps may ask for query execution before it is ready ; to prevent this, we had an init mechanism made in the `DataProxyProvider` to ensure the children are rendered only when the DataProxy is ready.

It works, but if something went wrong, nothing is displayed, which is confusing and hard to debug for the developer.

So, we change this by always rendering the children on the DataProxy side, and implement a queue on the `DataProxyLink` to bufferise requests, waiting for the dataproxy to be ready.

Once it is ready, all the requests in the queue are executed, in FIFO fashion.

### DIFF
--- a/packages/cozy-client/src/links/DataProxyLink.js
+++ b/packages/cozy-client/src/links/DataProxyLink.js
@@ -1,4 +1,5 @@
 import CozyLink from './CozyLink'
+import logger from 'cozy-logger'
 
 export default class DataProxyLink extends CozyLink {
   /**
@@ -8,6 +9,12 @@ export default class DataProxyLink extends CozyLink {
   constructor({ dataproxy } = {}) {
     super()
     this.dataproxy = dataproxy || null
+    this._queue = []
+    this._drainingRequests = false
+
+    if (window) {
+      window.addEventListener('message', this._onReceiveMessage)
+    }
   }
 
   registerClient(client) {
@@ -23,6 +30,7 @@ export default class DataProxyLink extends CozyLink {
    */
   registerDataProxy(dataproxy) {
     this.dataproxy = dataproxy
+    this._flushQueue()
   }
 
   async reset() {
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/links/DataProxyLink.js
```diff
@@ -23,24 +30,80 @@ export default class DataProxyLink extends CozyLink {
...
     // does nothing, we don't need any client for this kind of link
   }
 
   async request(operation, options, result, forward) {
-    if (this.dataproxy) {
-      // Send the request to the DataProxy, that will handle both web and mobile environments
-      const opts = options || {}
-      return this.dataproxy.requestLink(operation, opts)
-    } else {
-      return forward(operation, options)
```
zatteo, line 38: @paultranvan no more forward? It is an issue no?

zatteo (approved):

packages/cozy-client/src/links/DataProxyLink.js
```diff
@@ -23,24 +30,80 @@ export default class DataProxyLink extends CozyLink {
    */
   registerDataProxy(dataproxy) {
     this.dataproxy = dataproxy
+    this._flushQueue()
```
Crash--, line 33: should we await?

packages/cozy-client/src/links/DataProxyLink.js
```diff
@@ -23,24 +30,80 @@ export default class DataProxyLink extends CozyLink {
...
+          resolve(resp)
+        } catch (err) {
+          reject(err)
+        }
+      }
+    } finally {
+      this._drainingRequests = false
+      // Run it again in case queue has filled again during the flush
+      if (this._queue.length > 0) {
+        this._flushQueue()
```
Crash--, line 88: Same here, no await?

packages/cozy-client/src/links/DataProxyLink.js
```diff
@@ -23,24 +30,80 @@ export default class DataProxyLink extends CozyLink {
...
+      return
+    }
+    const eventData = event?.data
+    if (eventData && typeof eventData === 'object') {
+      if (
+        eventData.type === 'DATAPROXYMESSAGE' &&
+        eventData.payload === 'READY'
+      ) {
+        logger.info(`Received message: DataProxy is ready`)
+        this._flushQueue()
```
Crash--, line 105: same here?

packages/cozy-client/src/links/DataProxyLink.js
```diff
@@ -23,24 +30,80 @@ export default class DataProxyLink extends CozyLink {
    */
   registerDataProxy(dataproxy) {
```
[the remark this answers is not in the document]
paultranvan: This is called by the DataProxyProvider : https://github.com/cozy/cozy-libs/blob/master/packages/cozy-dataproxy-lib/src/dataproxy/DataProxyProvider.jsx#L175
This is typically done when everything is initialized, to provide the `dataproxy` instance to the link
About the throw, the promise will be rejected, so it is up to the initial requester to deal with it

packages/cozy-client/src/links/DataProxyLink.js
paultranvan: Same here, I see no reason to await , here we receive a "ready" message, so we just fire the flush

packages/cozy-client/src/links/DataProxyLink.js
Crash--: If we handle correctly the catch, yes, that's true. But I think that at the moment, this is not the case.

paultranvan: It is catch in  client.query : https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L1059

Crash-- (approved):

PR #1629, opened 2025-09-08
feat: Update date-fns

zatteo:
Update date-fns and use ^ operator to avoid  duplicate packages.

### DIFF
--- a/packages/cozy-client/package.json
+++ b/packages/cozy-client/package.json
@@ -20,7 +20,7 @@
     "@types/lodash": "^4.14.170",
     "btoa": "^1.2.1",
     "cozy-stack-client": "^60.6.0",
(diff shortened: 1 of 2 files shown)


packages/cozy-client/package.json
```diff
@@ -20,7 +20,7 @@
     "@types/lodash": "^4.14.170",
     "btoa": "^1.2.1",
     "cozy-stack-client": "^60.6.0",
-    "date-fns": "2.29.3",
+    "date-fns": "^2.30.0",
```
JF-Cozy, line 23: I think you should fix the version like we did in cozy-ui because we can have this packages in multiples lib (cozy-client, cozy-ui...) and in the app at the same time 🤔

PR #1630, opened 2025-09-19
feat: Update get full path to use in shared drive :sparkles:

lethemanh:
### Change:

Update `getFullpath` function to accept `driveId` to be used in shared drive.

### DIFF
--- a/packages/cozy-client/src/models/file.js
+++ b/packages/cozy-client/src/models/file.js
@@ -338,14 +338,19 @@ export const hasMetadataAttribute = ({ file, attribute }) =>
  * @param {string} name  - The file's name
  * @returns {Promise<string>} The full path of the file in the cozy
  **/
-export const getFullpath = async (client, dirId, name) => {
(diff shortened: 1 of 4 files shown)


packages/cozy-client/src/models/file.js
```diff
@@ -338,13 +338,15 @@ export const hasMetadataAttribute = ({ file, attribute }) =>
...
+export const getFullpath = async (client, dirId, name, driveId = undefined) => {
   if (!dirId) {
     throw new Error('You must provide a dirId')
   }
 
   const { data: parentDir } = await client.query(
-    Q(DOCTYPE_FILES).getById(dirId)
+    Q(DOCTYPE_FILES)
+      .getById(dirId)
+      .sharingById(driveId)
```
rezk2ll, line 349: or we can

```ts
  let query = Q(DOCTYPE_FILES).getById(dirId)

  if (driveId) {
    query = query.sharingById(driveId)
  }

  const { data: parentDir } = await client.query(query)
```

PR #1632, opened 2025-10-01
feat(cozy-pouch-link): Handle shared drive recipients' files

doubleface:
Now shared drives recipients files have their own "virtual" doctype like `"io.cozy.shareddrives-<driveId>"` which is first replicated using the `_changes` route for each shared drive (no `_all_docs` available)

To handle realtime addition/removal of shared drives, "addDoctype" and "removeDoctype" method have been added to the PouchLink class

### DIFF
--- a/packages/cozy-pouch-link/src/startReplication.js
+++ b/packages/cozy-pouch-link/src/startReplication.js
@@ -303,21 +303,25 @@ export const sharedDriveReplicateAllDocs = async ({
     // PouchDB.bulkDocs ignores _deleted documents with revision newer than the existing
     // document
     // The workaround is to get documents from PouchDB with correct revision and delete them
+    // see https://github.com/pouchdb/pouchdb/issues/7841
     if (toDelete.length > 0) {
       // lets try to find a document with the same _id in allDocsWithDriveId
-      for (const toDeleteDoc of toDelete) {
-        try {
-          const originalDoc = await pouch.get(toDeleteDoc._id)
-          if (originalDoc) {
-            await pouch.remove(originalDoc)
-          } else {
-            logger.error(
-              `sharedDriveReplicateAllDocs: Document ${toDeleteDoc._id} not found in local pouch`
-            )
-          }
-        } catch (error) {
+      const updatedToDeleteDocs = await pouch.bulkGet({
+        docs: toDelete.map(d => ({ id: d._id }))
+      })
+      const toBulkDelete = updatedToDeleteDocs.results
+        .map(doc => ({
+          ...doc?.docs?.[0]?.ok,
+          _deleted: true
+        }))
+        .filter(doc => doc._id)
+      if (toBulkDelete.length > 0) {
+        await pouch.bulkDocs(toBulkDelete)
+        if (toBulkDelete.length !== toDelete.length) {
           logger.error(
-            `sharedDriveReplicateAllDocs: Error deleting document ${toDeleteDoc._id}: ${error}`
+            `sharedDriveReplicateAllDocs: Error deleting documents ${toDelete
+              .map(d => d._id)
+              .join(', ')}`
           )
         }
       }
--- a/packages/cozy-pouch-link/src/startReplication.spec.js
+++ b/packages/cozy-pouch-link/src/startReplication.spec.js
@@ -414,7 +414,9 @@ describe('startReplication', () => {
 
     const mockPouch = {
       get: jest.fn(),
-      remove: jest.fn()
+      remove: jest.fn(),
+      bulkGet: jest.fn(),
+      bulkDocs: jest.fn()
     }
 
     const mockStorage = {
@@ -606,8 +608,18 @@ describe('startReplication', () => {
       })
 
       // Mock that doc-2 exists in local pouch
-      mockPouch.get.mockResolvedValue({ _id: 'doc-2', _rev: '1-abc' })
-      mockPouch.remove.mockResolvedValue()
+      mockPouch.bulkGet.mockResolvedValue({
+        results: [
+          {
+            docs: [
+              {
+                ok: { _id: 'doc-2', _rev: '1-abc', name: 'file2.txt' }
+              }
+            ]
+          }
+        ]
+      })
+      mockPouch.bulkDocs.mockResolvedValue([{ ok: true, id: 'doc-2' }])
 
       const result = await sharedDriveReplicateAllDocs({
         driveId,
@@ -624,11 +636,17 @@ describe('startReplication', () => {
       ]
 
       expect(result).toEqual(expectedDocs)
-      expect(mockPouch.get).toHaveBeenCalledWith('doc-2')
-      expect(mockPouch.remove).toHaveBeenCalledWith({
-        _id: 'doc-2',
-        _rev: '1-abc'
+      expect(mockPouch.bulkGet).toHaveBeenCalledWith({
+        docs: [{ id: 'doc-2' }]
       })
+      expect(mockPouch.bulkDocs).toHaveBeenCalledWith([
+        {
+          _id: 'doc-2',
+          _rev: '1-abc',
+          name: 'file2.txt',
+          _deleted: true
+        }
+      ])
       expect(insertBulkDocs).toHaveBeenCalledWith(
         mockPouch,
         expectedDocs.filter(doc => !doc._deleted)
@@ -644,8 +662,23 @@ describe('startReplication', () => {
         pending: false
       })
 
-      // Mock that doc-1 does not exist in local pouch
-      mockPouch.get.mockResolvedValue(undefined)
+      // Mock that doc-1 does not exist in local pouch (bulkGet returns error)
+      mockPouch.bulkGet.mockResolvedValue({
+        results: [
+          {
+            docs: [
+              {
+                error: {
+                  status: 404,
+                  name: 'not_found',
+                  message: 'missing'
+                }
+              }
+            ]
+          }
+        ]
+      })
+      mockPouch.bulkDocs.mockResolvedValue([])
 
       const result = await sharedDriveReplicateAllDocs({
         driveId,
@@ -659,8 +692,10 @@ describe('startReplication', () => {
       const expectedDocs = [{ _id: 'doc-1', _deleted: true, driveId }]
 
       expect(result).toEqual(expectedDocs)
-      expect(mockPouch.get).toHaveBeenCalledWith('doc-1')
-      expect(mockPouch.remove).not.toHaveBeenCalled()
+      expect(mockPouch.bulkGet).toHaveBeenCalledWith({
+        docs: [{ id: 'doc-1' }]
+      })
+      expect(mockPouch.bulkDocs).not.toHaveBeenCalled()
       expect(insertBulkDocs).toHaveBeenCalledWith(
         mockPouch,
         expectedDocs.filter(doc => !doc._deleted)


packages/cozy-pouch-link/src/remote.js
```diff
@@ -62,7 +62,7 @@ export const fetchRemoteLastSequence = async baseUrl => {
   const remoteUrl = new URL(`${baseUrl}/_changes`)
   const res = await fetchRemoteInstance(remoteUrl, {
     limit: 1,
-    descending: true
+    ...(baseUrl.includes('sharings/drives') ? {} : { descending: true }) // shared drives' _change route does not support descending. It look like it has no impact on the expected result
```
paultranvan, line 65: I don't understand how it can works: this is done to retrieve the lastSeq from the lastest change. So either you retrieve the wrong lastSeq, either the descending order is true by default for shared drives (which is normally false by default). @nono can you help us? This might require to add the descending order on the stack side (or maybe there is a simpler way to retrieve the lastSeq?)

doubleface: I see, since I don't do try to do the replication with PouchDb replication anymore, this change is not needed anymore. I remove it

paultranvan: > if there are some documents added during the replication

if we retrieve `lastSeq=10` and new docs are added afterwards, making `lastSeq = 12` we will get them next time we replicate as we'll ask docs since 10. But it might be problematic if it fails indeed.  

> You should use the lastSeq from last successful call of the replication for that.

Yes, it's probably better. @doubleface we can plug it [here](https://github.com/cozy/cozy-client/blob/master/packages/cozy-pouch-link/src/startReplication.js#L106) and listen for `replication.complete` to get the lastSeq. 
See the example given in https://pouchdb.com/api.html#replication

doubleface: And I use the lastSeq from the previous request for this.

packages/cozy-pouch-link/src/startReplication.js
```diff
@@ -208,3 +228,55 @@ export const replicateAllDocs = async ({ db, baseUrl, doctype, storage }) => {
...
+ *
+ * @param {Object} params - The parameters object
+ * @param {string} params.driveId - The unique identifier of the shared drive to replicate from
+ * @param {Object} params.pouch - The local PouchDB instance to replicate documents to
+ * @param {Object} params.storage - Storage interface for persisting replication state
+ * @param {string} params.doctype - The document type being replicated (e.g., 'io.cozy.files')
+ * @param {CozyClient} params.client - CozyClient instance for fetching remote documents
+ * @returns {Promise<Array>} A promise that resolves to an array of all replicated documents
+ * @throws {Error} Throws an error if driveId is not provided
+ */
```
paultranvan, line 246: Why is this necessary to make this function? When we discussed about it, I  understood that we can call the pouch/couch replication for the shared drive, is it not the case @nono ?

nono: Because we don't the paths of the files inside CouchDB, so we can't just make a request to CouchDB to get the relevant data, we need to iterate on each file, get its parent, compute its path, see if it's the shared drive, etc.

paultranvan: But isn't it what we already do, to get the change feed for each shared file?

packages/cozy-pouch-link/src/replicateOnce.js
```diff
@@ -48,15 +44,8 @@ export const replicateOnce = async pouchManager => {
       const replicationFilter = doc => {
         return !startsWith(doc._id, '_design')
       }
-      let seq = ''
-      if (initialReplication) {
-        // Before the first replication, get the last remote sequence,
-        // which will be used as a checkpoint for the next replication
-        const lastSeq = await fetchRemoteLastSequence(getReplicationURL())
```
paultranvan, line 49: I guess we can remove fetchRemoteLastSequence if it's not used elsewhere?

packages/cozy-pouch-link/src/replicateOnce.js
```diff
@@ -73,11 +62,6 @@ export const replicateOnce = async pouchManager => {
         pouchManager.storage,
         pouchManager.client
       )
-      if (seq) {
-        // We only need the sequence for the second replication, as PouchDB
-        // will use a local checkpoint for the next runs.
-        await pouchManager.storage.destroyDoctypeLastSequence(doctype)
-      }
```
paultranvan, line 78: This changes the behaviour: before we were using the lastSeq only for 2nd replication (because the first one was not done through pouch replication). Now, we rely on it for all the replications. I guess it's ok, but I'm wondering if any side effect could occur

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -799,6 +799,32 @@ class PouchLink extends CozyLink {
...
+   * adds it to the pouches, and starts replication.
+   *
+   * @param {string} doctype - The name of the doctype to add.
+   * @param {Object} options - The replication options for the doctype.
+   */
+  addDoctype(doctype, options) {
+    this.doctypes.push(doctype)
+    this.doctypesReplicationOptions[doctype] = options
+    this.pouches.addDoctype(doctype, options)
+    this.startReplication()
```
paultranvan, line 814: The replication start is arguable: I'm not sure this is something we should expect from a `addDoctype` method

packages/cozy-pouch-link/src/PouchManager.js
```diff
@@ -351,6 +351,45 @@ class PouchManager {
...
   }
+
+  /**
+   * Adds a new doctype to the list of managed doctypes, sets its replication options,
+   * creates a new PouchDB instance for it, and sets up the query engine.
+   *
+   * @param {string} doctype - The name of the doctype to add.
+   * @param {Object} options - The replication options for the doctype.
+   */
+  addDoctype(doctype, options) {
```
paultranvan, line 362: Maybe this should be called from the `init()` in PouchManager ? To avoid code duplication and forget to change this code if anything change in the future

packages/cozy-pouch-link/src/replicateOnce.js
doubleface: I don't understand. With this we still do the first replication without pouch replication. It only depends on `initialReplication` parameter which does not change

packages/cozy-pouch-link/src/CozyPouchLink.js
doubleface: But I don't we where else the replication will/should be started. Should it be started by the caller of addDoctype ?

doubleface: Ok. Let's give this responsibility to the caller,  then

packages/cozy-pouch-link/src/replicateOnce.js
doubleface: Ok, I see. Let's not change this behavior. I put the sequence removal back

paultranvan: Well if you put the `lastSeq` removal as before, you will break the shared drive next "replications", since they won't have any `lastSeq` to rely on, as they don't use the actual replication protocol

packages/cozy-pouch-link/src/CozyPouchLink.js
doubleface: Done

packages/cozy-pouch-link/src/startReplication.js
```diff
@@ -208,3 +244,73 @@ export const replicateAllDocs = async ({ db, baseUrl, doctype, storage }) => {
...
+      for (const toDeleteDoc of toDelete) {
+        const originalDoc = await pouch.get(toDeleteDoc._id)
+        if (originalDoc) {
+          await pouch.remove(originalDoc)
+        }
+      }
+    }
+
+    startDocId = newLastSeq
+    await helpers.insertBulkDocs(pouch, filteredDocs)
```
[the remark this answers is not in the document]
doubleface: I can add some unit tests mocking PouchDb and stack, but I doubt of their utility without any "browser" PouchDb since I am trying to bypass an "interface bug" between stack and PouchDb. But here are tests with full mock

packages/cozy-pouch-link/src/replicateOnce.js
```diff
@@ -69,11 +59,13 @@ export const replicateOnce = async pouchManager => {
...
+        pouchManager.storage,
+        pouchManager.client
       )
-      if (seq) {
-        // We only need the sequence for the second replication, as PouchDB
-        // will use a local checkpoint for the next runs.
+      const isSharedDrive = Boolean(replicationOptions.driveId)
+      if (seq && !isSharedDrive) {
+        // For shared drives, we always need to keep the last sequence since replication is handled manually, not by PouchDB.
+        // For standard PouchDB replications, the last sequence is only needed for the initial sync; subsequent runs use PouchDB's internal checkpointing.
```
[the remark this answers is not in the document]
doubleface: But I did not change that. the `lastSeq` was already saved on each completed replication.
It looks like the previous comment was wrong and the PouchDb replication already used the `lastSeq` saved in localStorage for every replication (the first one with alldocs expected).

Yes, it means that there is no point at destroying it here but what I don't know is if there is a need to avoid to use the lastSeq in localStorage.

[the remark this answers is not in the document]
paultranvan: > the lastSeq was already saved on each completed replication.

Well, no, that's you who added the save on each completed replication :smile: https://github.com/cozy/cozy-client/pull/1632/commits/ac5fb4787115caaa742b50150d8cbe1d2e2c24e8#diff-6f9271f35894bccd13c05f69ba8e460d9cae55ed63666aa0b2038a5b4bc5a328

packages/cozy-pouch-link/src/startReplication.js
taratatach: No, to get a shared drive's changes feed we filter the main changes feed but don't have to iterate over every file.

packages/cozy-pouch-link/src/replicateOnce.js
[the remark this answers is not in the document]
doubleface: Yes sorry. At this point, I think I just should revert [`ac5fb47` (#1632)](https://github.com/cozy/cozy-client/pull/1632/commits/ac5fb4787115caaa742b50150d8cbe1d2e2c24e8) since it has no point anymore

packages/cozy-pouch-link/src/startReplication.js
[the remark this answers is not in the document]
doubleface: Regarding the deleted doc still sent to bulkDocs, I did it because these docs where ignored by PouchDb anyway. But let's make it cleaner. Done. I also added error handling which I found missing while reviewing this part.
I also added more comments on the why of this workaround.

packages/cozy-pouch-link/src/CozyPouchLink.js
```diff
@@ -45,10 +45,12 @@ const addBasicAuth = (url, basicAuth) => {
...
 }
 
-export const getReplicationURL = (uri, token, doctype) => {
+export const getReplicationURL = (uri, token, doctype, replicationOptions) => {
   const basicAuth = token.toBasicAuth()
   const authenticatedURL = addBasicAuth(uri, basicAuth)
-  return `${authenticatedURL}/data/${doctype}`
+  if (replicationOptions?.driveId) {
+    return `${authenticatedURL}/sharings/drives/${replicationOptions?.driveId}`
+  } else return `${authenticatedURL}/data/${doctype}`
```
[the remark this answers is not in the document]
doubleface: I found the version with else clearer

packages/cozy-pouch-link/src/startReplication.js
```diff
@@ -208,3 +235,99 @@ export const replicateAllDocs = async ({ db, baseUrl, doctype, storage }) => {
...
+    // PouchDB.bulkDocs ignores _deleted documents with revision newer than the existing
+    // document
+    // The workaround is to get documents from PouchDB with correct revision and delete them
+    if (toDelete.length > 0) {
+      // lets try to find a document with the same _id in allDocsWithDriveId
+      for (const toDeleteDoc of toDelete) {
+        try {
+          const originalDoc = await pouch.get(toDeleteDoc._id)
+          if (originalDoc) {
+            await pouch.remove(originalDoc)
```
paultranvan: since this works it is not blocking, but I think we should eventually do a bulkGet and then a bulkDelete, that would be much more efficient

Crash--:
And the missing route from the stack today is `/_bulk_get` ?

packages/cozy-pouch-link/src/startReplication.js
```diff
@@ -208,3 +235,102 @@ export const replicateAllDocs = async ({ db, baseUrl, doctype, storage }) => {
...
+    }
+    // FIXME this is a workaround to allow to delete documents from the shared drive
+    // changes
+    // PouchDB.bulkDocs ignores _deleted documents with revision newer than the existing
+    // document
+    // The workaround is to get documents from PouchDB with correct revision and delete them
+    if (toDelete.length > 0) {
+      // lets try to find a document with the same _id in allDocsWithDriveId
+      const updatedToDeleteDocs = await pouch.bulkGet({
+        docs: toDelete.map(d => ({ id: d._id }))
```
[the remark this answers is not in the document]
doubleface: It looks like the `if (toDelete.length > 0)` does the job no ?

PR #1633, opened 2025-10-01
feat: Store PouchDB databases locally

paultranvan:
This is useful at cleanup time for old browsers that do not support the `indexeddb.databases()` function.


Crash-- (approved):
Great!

PR #1636, opened 2025-10-06
fix: Resolve set write permission for recipient in shared drive :bug:

lethemanh:
### Change:

Now, we can create a shared drive with write permission for recipients.

### DIFF
--- a/packages/cozy-stack-client/src/SharingCollection.js
+++ b/packages/cozy-stack-client/src/SharingCollection.js
@@ -362,14 +362,20 @@ const getSharingRulesForFile = document => {
  */
 const getSharingRulesForSharedDrive = document => {
   const { _id, name } = document
+  /**
+   * The sharing rules always set to write by default (same as the implementation from getSharingPolicyForFile)
+   * To check for read-only we based on `read_only` in member data
+   * Even in case shared drive has many recipients with differents permissions,
+   * the `rules` still set to write by default and use `read_only` to check for read-only recipients.
+   */
   return [
     {
       title: name,
       doctype: 'io.cozy.files',
       values: [_id],
       add: 'none',
-      update: 'none',
-      remove: 'none'
+      update: 'sync',
+      remove: 'sync'
     }
   ]
 }


packages/cozy-stack-client/src/SharingCollection.js
```diff
@@ -362,14 +362,20 @@ const getSharingRulesForFile = document => {
...
   return [
     {
       title: name,
       doctype: 'io.cozy.files',
       values: [_id],
       add: 'none',
-      update: 'none',
-      remove: 'none'
+      update: 'sync',
+      remove: 'sync'
```
zatteo, line 378: I do not understand what you are trying to fix here? Here it breaks description of shared drives in backend documentation https://docs.cozy.io/en/cozy-stack/shared-drives/
```
To create a shared drive (typically on the organization Cozy), we need the following steps:

    Ensure that the /Drives folder exists in the cozy instance with the [POST /files/shared-drives](https://docs.cozy.io/en/cozy-stack/files/#post-filesshared-drives) route.
    Create a folder inside it, with the name of shared drive.
    Create a sharing with the drive: true attribute, and one rule for shared folder (with none for add, update and remove attributes).
```

lethemanh: > Create a sharing with the drive: true attribute, and one rule for shared folder (with none for add, update and remove attributes).
We cannot create a shared drive folder that has write access with the rule like this. Also, you can refer to the function `hasWriteAccess` in `cozy-sharing` to see how the app checks write permission.
I think the backend document should be updated as well.

zatteo: @nono What do you think about this?

nono: I confirm that the stack documentation is correct: we should have `update: none` and `remove: none` for a shared drive. `Sync` means that when a change happens in a Cozy, it is also written to the other cozy instances of the sharing, which is not what we want for shared drive (files only on one instance).

lethemanh: So, what exact way to check write access for a shared drive? Could I use the data in `attributes.members` to check for write access?

PR #1637, opened 2025-10-06
Centralize shortcut code in cozy-client models and hooks

zatteo:


rezk2ll (approved):

packages/cozy-client/src/hooks/useFetchHomeShortcuts.jsx
```diff
@@ -0,0 +1,45 @@
+import CozyClient, { Q, useQuery } from 'cozy-client'
+
+const fetchPolicy = CozyClient.fetchPolicies.olderThan(30 * 1000)
```
paultranvan, line 3: There were previous discussions about raising up default fetch policy duration, since 30 seconds is very short. 
We have for instance 10mins in Drive: https://github.com/cozy/cozy-drive/blob/master/src/queries/index.ts#L29
Should we have a common policy here @Crash-- ?

PR #1648, opened 2025-11-12
fix(cozy-client): Failed queries not respecting fetchPolicy

rezk2ll:
### issue

When a query fails, it does not update the store, which causes the OlderThan fetch policy to always return true, and this will cause an infinite loop when trying to fetch the same query.

```ts
olderThan: delay => queryState => {
  if (!queryState || !queryState.lastUpdate) {
    return true
  } else {
    const elapsed = Date.now() - queryState.lastUpdate
    return elapsed > delay
  }
},
```

this is used everywhere, especially in the useInstanceInfo hook,
```ts
const buildSettingsByIdQuery = id => ({
  definition: Q('io.cozy.settings').getById(id),
  options: {
    as: `io.cozy.settings/${id}`,
    fetchPolicy: CozyClient.fetchPolicies.olderThan(
      DEFAULT_CACHE_TIMEOUT_QUERIES
    ),
    singleDocData: true
  }
})
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data refresh timing by considering both the latest successful update and error when determining whether information is stale.
  * Prevented unnecessary repeated fetch attempts after recent errors while still retrying once the configured delay has elapsed.

* **Tests**
  * Added coverage for missing, recent, stale, successful, and failed fetch states.

### DIFF
--- a/packages/cozy-client/src/policies.js
+++ b/packages/cozy-client/src/policies.js
@@ -17,10 +17,12 @@ const fetchPolicies = {
    * @returns {Function} Fetch policy to be used with `<Query />`
    */
   olderThan: delay => queryState => {
-    if (!queryState || !queryState.lastUpdate) {
+    const lastCheck = queryState.lastUpdate || queryState.lastErrorUpdate
+
+    if (!queryState || !lastCheck) {
       return true
     } else {
-      const elapsed = Date.now() - queryState.lastUpdate
+      const elapsed = Date.now() - lastCheck
       return elapsed > delay
     }
   },
--- a/packages/cozy-client/src/policies.js
+++ b/packages/cozy-client/src/policies.js
@@ -17,14 +17,18 @@ const fetchPolicies = {
    * @returns {Function} Fetch policy to be used with `<Query />`
    */
   olderThan: delay => queryState => {
-    const lastCheck = queryState.lastUpdate || queryState.lastErrorUpdate
-
-    if (!queryState || !lastCheck) {
+    if (!queryState) {
       return true
-    } else {
-      const elapsed = Date.now() - lastCheck
-      return elapsed > delay
     }
+    const lastFetch = Math.max(
+      queryState.lastUpdate || 0,
+      queryState.lastErrorUpdate || 0
+    )
+    if (!lastFetch) {
+      return true
+    }
+    const elapsed = Date.now() - lastFetch
+    return elapsed > delay
   },
 
   /**
--- a/packages/cozy-client/src/policies.spec.js
+++ b/packages/cozy-client/src/policies.spec.js
@@ -1,0 +1,65 @@
+import fetchPolicies from './policies'
+
+describe('fetchPolicies.olderThan', () => {
+  const NOW = 1600000000000
+  const delay = 30 * 1000
+  const policy = fetchPolicies.olderThan(delay)
+
+  beforeEach(() => {
+    jest.spyOn(Date, 'now').mockReturnValue(NOW)
+  })
+
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
+
+  it('refetches when there is no query state', () => {
+    expect(policy(undefined)).toBe(true)
+  })
+
+  it('refetches when the query has never been fetched nor errored', () => {
+    expect(policy({ lastUpdate: null, lastErrorUpdate: null })).toBe(true)
+  })
+
+  it('does not refetch a recent successful query', () => {
+    expect(policy({ lastUpdate: NOW - 1000, lastErrorUpdate: null })).toBe(
+      false
+    )
+  })
+
+  it('refetches a stale successful query', () => {
+    expect(policy({ lastUpdate: NOW - delay - 1, lastErrorUpdate: null })).toBe(
+      true
+    )
+  })
+
+  it('does not refetch right after a recent error (prevents infinite refetch loop on 500)', () => {
+    expect(policy({ lastUpdate: null, lastErrorUpdate: NOW - 1000 })).toBe(
+      false
+    )
+  })
+
+  it('refetches again once the error is older than the delay', () => {
+    expect(policy({ lastUpdate: null, lastErrorUpdate: NOW - delay - 1 })).toBe(
+      true
+    )
+  })
+
+  it('backs off on a recent error even when the last success is stale', () => {
+    expect(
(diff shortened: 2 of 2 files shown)


Crash--:
@rezk2ll I updated the PR. Is it ok for you?

Crash--:
## Why a persistent 5xx turns into an infinite request storm — full analysis

I reproduced this at runtime and traced the exact mechanism. Sharing the whole picture here since it explains both *why* the loop happens and *why* keying the backoff off the most recent of the two timestamps (rather than `lastUpdate || lastErrorUpdate`) is the correct fix.

### 1. Runtime reproduction

With the backend forced to return `500` on `GET /settings/instance` from the very first load, Twake Drive fires **~820 requests in ~4 seconds** and keeps going. The decisive detail: it is **not** the failing request that repeats on its own — a whole bundle of *unrelated* boot queries refires together on every cycle:

```
GET  /sharings/drives                       -> 200
GET  /settings/instance                     -> 500
POST /data/io.cozy.nextcloud.migrations/_find -> 200
GET  /data/io.cozy.files.settings/_normal_docs -> 404
```

Requests that *succeed* (200/404) refiring alongside the 500 means the **whole component subtree is unmounting and remounting** in a loop, not that a single query is retrying.

### 2. Root cause in the store + policy

A failed query never records a "we fetched at time T" timestamp. In the initial query state, [`lastFetch`, `lastUpdate` and `lastErrorUpdate` all start `null`](https://github.com/linagora/cozy-client/blob/c9f48ee6/packages/cozy-client/src/store/queries.js#L55-L57). On success the store sets [`lastFetch`/`lastUpdate`](https://github.com/linagora/cozy-client/blob/c9f48ee6/packages/cozy-client/src/store/queries.js#L196), but [`RECEIVE_QUERY_ERROR`](https://github.com/linagora/cozy-client/blob/c9f48ee6/packages/cozy-client/src/store/queries.js#L267-L275) only sets `fetchStatus: 'failed'`, `lastError` and `lastErrorUpdate` — **never `lastUpdate` nor `lastFetch`**.

`olderThan` on `master` looks only at `lastUpdate`, so an erroring query is treated as *never fetched* and reported stale forever. `client.query` honours the policy and only fetches when it returns `true` ([`CozyClient.js#L1008-L1010`](https://github.com/linagora/cozy-client/blob/c9f48ee6/packages/cozy-client/src/CozyClient.js#L1008-L1010)), so a persistent 5xx is re-issued on every render/mount with no negative caching.

### 3. Why it becomes a full-subtree remount loop

In the Drive app, the whole tree is wrapped in `CozyTheme` from `cozy-ui-plus`, which (when not public) reads `io.cozy.settings.instance` to pick the color scheme. Its implementation — [`CozyThemeWithQuery`](https://github.com/cozy/cozy-libs/tree/master/packages/cozy-ui-plus/src/providers/CozyTheme) — gates the entire app on that query:

```jsx
const { data, ...rest } = useQuery(buildSettingsInstanceQuery()) // io.cozy.settings.instance, olderThan(5min)
if (isQueryLoading(rest) && !hasQueryBeenLoaded(rest)) return null // <- unmounts all children
```

where [`isQueryLoading`](https://github.com/linagora/cozy-client/blob/c9f48ee6/packages/cozy-client/src/utils.js#L40-L45) is `fetchStatus ∈ {loading, pending}` and [`hasQueryBeenLoaded`](https://github.com/linagora/cozy-client/blob/c9f48ee6/packages/cozy-client/src/utils.js#L51-L52) is just `col.lastFetch`.

The key is that **the same `io.cozy.settings.instance` query (same `as`) is shared** between `CozyTheme` (the gate) and `useInstanceInfo`, which runs in a *child* of `CozyTheme`. A `useQuery` fetch fires once per mount — its effect deps are [`[as, enabled]`](https://github.com/linagora/cozy-client/blob/c9f48ee6/packages/cozy-client/src/hooks/useQuery.js#L70-L76), both stable — so the loop is driven by mount/unmount, not by re-render:

1. `CozyTheme` issues the query → shared state `loading`, `lastFetch` null → gate returns `null` → **children unmount**.
2. `500` → state `failed` (no `lastFetch`/`lastUpdate`). `failed` is not `loading` → `CozyTheme` **renders its children** → they **mount**.
3. On mount, the child `useInstanceInfo` calls `client.query('io.cozy.settings.instance')`. `olderThan` sees `lastUpdate === null` → **stale** → `client.query` re-issues → shared state flips back to `loading`.
4. `CozyTheme` (subscribed to the same state) now sees `loading` and `lastFetch` still null → gate returns `null` → **children unmount again**.
5. The re-issued fetch `500`s → `failed` → children mount again → back to step 3. **Infinite.**

Every child remount also re-runs the other boot queries (`/sharings/drives`, migrations, `files.settings`), which is exactly the bundle seen in the network trace.

Note this specifically requires the settings request to fail from the *first* load: `lastFetch` is only ever set on success, so `hasQueryBeenLoaded` stays falsy and the gate degenerates into "any `loading` unmounts the children". Had settings succeeded once, `!hasQueryBeenLoaded` would be false and there would be no loop.

### 4. Why `max(lastUpdate, lastErrorUpdate)` is the right fix

The backoff must key off **whichever happened last, success or error** — see [`policies.js#L19-L32`](https://github.com/linagora/cozy-client/blob/c9f48ee6/packages/cozy-client/src/policies.js#L19-L32).

`lastUpdate || lastErrorUpdate` is **not** equivalent: if a query succeeded a while ago (`lastUpdate` old) and then starts erroring (`lastErrorUpdate` recent), the `||` keeps the stale *success* timestamp, so `elapsed > delay` stays true and the loop survives in that case. `Math.max(lastUpdate || 0, lastErrorUpdate || 0)` fixes it. It also guards `queryState` before dereferencing it (the `const lastCheck = queryState.lastUpdate` form throws on `queryState === undefined`).

With this, after the first `500` `lastErrorUpdate` is recent → `olderThan` returns `false` → at step 3 above, `client.query` short-circuits and dispatches no load → the shared state stays `failed`, never flips back to `loading` → `CozyTheme` keeps its children mounted → **the loop stops**. The app renders (without a custom theme) on a stable error state, and the query is retried once the delay elapses.

The added `policies.spec.js` covers fresh/stale success, error backoff (the 500 loop), retry-after-delay, and the max-of-both-timestamps precedence.

rezk2ll:
> @rezk2ll I updated the PR. Is it ok for you?

thanks @Crash-- let's merge :rocket:

Crash-- (approved):

PR #1682, opened 2026-05-12
fix(pouch): pass clean doctypes to onSync (not prefixed pouch dbNames)

Crash--:
## Bug

`replicateOnce` zips per-doctype replication results with the keys of
`pouchManager.pouches`, which are the FULL prefixed Pouch dbNames
(e.g. `instance__doctype__io.cozy.todos`). The downstream consumer
(`handleOnSync` in `CozyPouchLink`) forwards that prefixed name as
the doctype to `normalizeDocs` + `client.setData`. The store's
`autoQueryUpdater` compares `query.definition.doctype` (clean, e.g.
`'io.cozy.todos'`) to the incoming `doc._type` (prefixed) -> mismatch
for every active query -> SKIP.

Symptom in a downstream consumer (twake-drive-mobile): periodic-sync
replication correctly pulls new revisions into local Pouch SQLite
(verified at the SQLite level), but the React UI never refreshes
because the store's auto-update path bails for every query. Only
pull-to-refresh works (it goes through the link chain which re-reads
from local Pouch directly).

Trace from the consumer with a temporary diagnostic log:

```
[handleOnSync] setData ["instance__doctype__io.cozy.files=1"]
[autoQueryUpdater] SKIP doctype mismatch io.cozy.files/...
  io.cozy.files vs data._type= instance__doctype__io.cozy.files
```

## Fix

In `replicateOnce.js`, map the dbNames through
`getDoctypeFromDatabaseName` before zipping. `doctypeUpdated` is
now keyed by the clean doctype, the downstream chain sees the
correct `doc._type`, `autoQueryUpdater` matches, queries auto-
update on every sync round.

## Tests

- New `replicateOnce.spec.js` regression test: assert `onSync` is
  called with clean doctype keys, not prefixed dbNames.
- Updated `PouchManager.spec.js` — the existing assertion
  `"should call on sync with doctype updates"` was encoding the
  bug (it asserted the prefixed key `'cozy.tools__doctype__io.cozy.todos'`).
  Switched to the clean `'io.cozy.todos'`. Without this fix the
  test passes but documents broken behavior.

Full `cozy-pouch-link` jest suite: 188 tests passing.

## Note

Discovered while working on the parent-lookup guard fix in #1681 —
originally bundled there but split out into its own PR for clean
review/merge.

### DIFF
--- a/packages/cozy-pouch-link/src/replicateOnce.js
+++ b/packages/cozy-pouch-link/src/replicateOnce.js
@@ -87,7 +87,9 @@ export const replicateOnce = async pouchManager => {
   )
 
   // Waiting on each replication
-  const doctypes = Object.keys(pouchManager.pouches)
+  const doctypes = Object.keys(pouchManager.pouches).map(dbName =>
+    getDoctypeFromDatabaseName(dbName)
+  )
   const promises = Object.values(pouchManager.replications)
   try {
     const res = await allSettled(promises)
--- a/packages/cozy-pouch-link/src/PouchManager.spec.js
+++ b/packages/cozy-pouch-link/src/PouchManager.spec.js
@@ -235,7 +235,7 @@ describe('PouchManager', () => {
     manager.updateSyncInfo('io.cozy.todos')
     await manager.replicateOnce()
     expect(onSync).toHaveBeenCalledWith({
-      'cozy.tools__doctype__io.cozy.todos': [
+      'io.cozy.todos': [
         {
           _id: '1',
           name: 'Make replication work'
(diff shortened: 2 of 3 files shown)


zatteo (approved):

packages/cozy-pouch-link/src/replicateOnce.js
```diff
@@ -87,7 +87,9 @@ export const replicateOnce = async pouchManager => {
   )
 
   // Waiting on each replication
-  const doctypes = Object.keys(pouchManager.pouches)
```
Crash--, line 90: @doubleface Do you use replicateOnce for the shared drive replication?

doubleface: Not directly but the search engine in dataproxy calls startReplication from PouchLink which calls startReplication which call replicateOnce on each pouches

Crash--: how can I test that I didn't break anything for that? I think we don't care because search is reading directly from Pouch directly not relying on "queries".

PR #1684, opened 2026-05-26
Fix AI assistant creation and deletion

paultranvan:
- For creation : the newly created assistant was not returned, so the caller could not use it
- For edition : the UI were not updated because of the store bypass

### DIFF
--- a/packages/cozy-client/src/models/assistant.js
+++ b/packages/cozy-client/src/models/assistant.js
@@ -21,7 +21,7 @@ const ACCOUNT_DOCTYPE = 'io.cozy.accounts'
  *
  * @param {CozyClient} client - An instance of CozyClient
  * @param {Assistant} assistantData - Data for the new assistant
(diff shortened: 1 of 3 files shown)


packages/cozy-client/src/models/assistant.js
```diff
@@ -21,7 +21,7 @@ const ACCOUNT_DOCTYPE = 'io.cozy.accounts'
  *
  * @param {CozyClient} client - An instance of CozyClient
  * @param {Assistant} assistantData - Data for the new assistant
- * @returns {Promise<void>} - A promise that resolves when the assistant is created
+ * @returns {Promise<object>} - A promise that resolves with the created assistant document
```
zatteo: I am sure our common friend could type properly this

zatteo (approved):

packages/cozy-client/src/models/assistant.js
paultranvan: And the Assistant is already typed, so I had no excuse

PR #1689, opened 2026-06-12
feat(permissions): Add shared drive link lookup by ids

doubleface:
This is needed to allow recipients of  shared drives to see the links they created on files or folders inside of shared drives

## Summary by CodeRabbit

* **New Features**
  * Add a PermissionCollection method to retrieve sharing/permission links by document IDs for shared drives, returning normalized permission entries.

* **Tests**
  * Add coverage for shared-drive behavior: validates input, handles empty ID lists without network calls, enforces shared-drive configuration, and verifies request/response mapping.

* **Documentation**
  * Update API docs to describe the new method, its route behavior for shared drives, return format, and error cases.

### DIFF
--- a/packages/cozy-stack-client/src/PermissionCollection.js
+++ b/packages/cozy-stack-client/src/PermissionCollection.js
@@ -6,6 +6,11 @@ import logger from './logger'
 
 const normalizePermission = perm => normalizeDoc(perm, 'io.cozy.permissions')
 
+const normalizePermissionsResponse = resp => ({
+  ...resp,
+  data: resp.data.map(normalizePermission)
+})
+
 /**
  * Options that can be passed to PermissionCollection's constructor
  *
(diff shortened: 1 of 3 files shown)


packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -153,6 +153,46 @@ class PermissionCollection extends DocumentCollection {
...
+    if (!this.prefix) {
+      throw new Error('findLinksByIds is only supported for shared drives')
+    }
+
+    if (!Array.isArray(ids)) {
+      throw new TypeError('ids must be an array')
+    }
+
+    if (ids.length === 0) {
+      return { data: [] }
```
rezk2ll: why empty state returns { data: [] } while pupolated state returns { ...resp, data: ... }, why should have cosistent return shape

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -153,6 +153,46 @@ class PermissionCollection extends DocumentCollection {
...
+   * URL becomes `/sharings/drives/:driveId/permissions?ids=...` when the
+   * collection is built with `{ driveId }`.
+   *
+   * The returned permissions are normalized (have `_id` and `_type`).
+   *
+   * @param {string[]} ids - The list of document ids to look up
+   * @returns {Promise} The matching permissions
+   */
+  async findLinksByIds(ids) {
+    if (!this.prefix) {
```
rezk2ll, line 176: i think this should come after Array.isArray(ids) check no?

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -153,6 +153,46 @@ class PermissionCollection extends DocumentCollection {
...
+    if (!Array.isArray(ids)) {
+      throw new TypeError('ids must be an array')
+    }
+
+    if (ids.length === 0) {
+      return { data: [] }
+    }
+
+    const searchParams = new URLSearchParams()
+    searchParams.append('ids', ids.join(','))
```
rezk2ll, line 189: i know it cannot happen today but if have ids containing ',' this will break ( in theory )

packages/cozy-stack-client/src/PermissionCollection.js
```diff
@@ -153,6 +153,46 @@ class PermissionCollection extends DocumentCollection {
...
+   * This is the shared-drive equivalent of `findLinksByDoctype` when you
+   * already know the document ids. It uses the same `this.prefix`, so the
+   * URL becomes `/sharings/drives/:driveId/permissions?ids=...` when the
+   * collection is built with `{ driveId }`.
+   *
+   * The returned permissions are normalized (have `_id` and `_type`).
+   *
+   * @param {string[]} ids - The list of document ids to look up
+   * @returns {Promise} The matching permissions
+   */
```
rezk2ll, line 174: missing @throws

packages/cozy-stack-client/src/PermissionCollection.js
doubleface: Right but stack api expects this at the moment

packages/cozy-stack-client/src/PermissionCollection.js
doubleface: Why? If no prefix is set, we don't even have to look at params since this api is not available outside shared drives

packages/cozy-stack-client/src/PermissionCollection.js
doubleface: Ok, let's normize the response

rezk2ll (approved):

PR #1696, opened 2026-07-07
feat(cozy-stack-client): Add SharingCollection read-only methods for groups

zatteo:
Wrap the POST and DELETE /sharings/:id/groups/:index/readonly endpoints exposed by the stack to toggle a sharing group's read_only flag.

## Summary by CodeRabbit

* **New Features**
  * Added group-level sharing controls to set a group as read-only or restore read/write access.
* **Documentation**
  * Updated the public API docs to include the new group permission methods.
* **Tests**
  * Added coverage for the new group permission actions to verify the correct behavior.


Crash-- (approved):

Crash--:
@zatteo why this is not merged?

zatteo:
Because this was part of https://github.com/linagora/cozy-stack/pull/4847 that was closed because it will be done another way in the stack.

PR #1697, opened 2026-07-25
fix(store): resolve getById queries by _id index on setData

Crash--:
## Summary

On a full sync, `client.setData(data)` dispatches `RECEIVE_QUERY_RESULT` with `queryId: null`, which re-evaluates every active query. `updateData` ran `groupBy(newData, makeFilterDocumentFn(query.definition))`, a sift scan over the whole `newData` array, once per query. With ~1500 getById queries active against ~9840 docs that is ~15M evaluations per dispatch and blocks the thread for ~38s.

getById/getByIds queries do not need to scan all of `newData`: this indexes `newData` by `_id` once per dispatch and resolves them with an O(1) lookup, mirroring the O(1) getById access already used by `executeQueryFromState`. Selector queries keep the sift scan but memoize the partition per definition within a dispatch. The `_deleted` / `_type === doctype` matching semantics and the autoUpdate add/remove/update behaviour are unchanged.

Measured on a real device: setData 38588ms to 247ms (~150x).

## Tests

- `yarn jest packages/cozy-client/src/store` gives 104 passing, snapshots unchanged.
- Added `computeMatchPartition` unit tests: getById hit / missing / wrong doctype / deleted, getByIds, selector, and per-dispatch memoization.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved automatic query updates for faster data refreshes.
  * Optimized lookups for queries targeting specific records.
  * Improved handling of deleted, missing, or mismatched records during updates.

* **Bug Fixes**
  * Ensured query results correctly distinguish matching and non-matching records across supported query types.

### DIFF
--- a/packages/cozy-client/src/store/queries.js
+++ b/packages/cozy-client/src/store/queries.js
@@ -442,16 +442,6 @@ export const makeSorterFromDefinition = definition => {
   }
 }
 
-// Within a single dispatch, `newData` is the same array reference for every
-// active query. A `getById`/`getByIds` query does not need to sift-scan the
-// whole `newData` array to know if its documents are in there: it only cares
-// about a handful of known ids. We therefore index `newData` by `_id` once per
-// dispatch and resolve those queries with an O(1) lookup, mirroring the O(1)
-// getById access already used by `executeQueryFromState`. Selector queries keep
-// the sift scan but memoize the partition per distinct definition, so several
-// queries sharing the same selector only scan `newData` once. On a full-sync
-// dispatch (thousands of getById queries against thousands of docs) this turns
-// an O(queries * docs) scan into O(queries + docs).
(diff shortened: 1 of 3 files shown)


Crash--:
This extends #1591. That issue's O(1)-by-id fix only landed in `executeQueryFromState` (the read path). The same O(n) sift scan remained in `updateData` (the `RECEIVE_QUERY_RESULT` / `setData` path), where ~1500 getById queries each scanning every doc blocked ~38s on a real device during a full sync. This PR applies the same `_id`-index lookup there (38588ms to 247ms).

Side note recorded in #1698: upgrading sift to v17 was investigated as an alternative and is a perf regression (2.8x slower exec), so this id-index approach, not a sift bump, is the right fix.

zatteo (approved):

Issue #1698, opened 2026-07-25
sift 6 to 17 upgrade is a performance regression (keep sift@6)


Crash--:
## Summary

While investigating store query-evaluation performance, I attempted to upgrade `sift`
from 6.0.0 to 17.1.3 (the current major). The port is functionally clean, but v17 is a
significant performance regression for our usage, so we should stay on sift@6. Recording
this so the upgrade is not attempted again (for example via the Dependency Dashboard,
#1014).

## The port works (but do not ship it)

`getSelectorFilterFn` (`packages/cozy-client/src/store/queries.js`) relies on the v6 global
`sift.use({ $gtnull })` registry, which was removed after v6. It ports cleanly to v17 via
`options.operations`:

```js
import sift, { createEqualsOperation } from 'sift'

const siftOperations = {
  $gtnull: (_params, ownerQuery, options) =>
    createEqualsOperation(actualValue => !!actualValue, ownerQuery, options)
}
// ...
return sift(convert$gtNullSelectors(selectors), { operations: siftOperations })
```

The full cozy-client suite passes on this branch: 960/960 tests, 19 snapshots unchanged.
The `{$gt: null}` semantics (matches when the field is truthy) are identical to v6.

## Benchmark: v17 is slower

Selector `{ dir_id, type: 'file', name: { $gtnull: null } }` over 10 000 docs, Node 20,
median of 5 runs:

| | sift 6.0.0 | sift 17.1.3 |
|---|---|---|
| compile x5000 | 2.7 ms | 58.7 ms (21x slower) |
| exec: filter 10 000 docs x500 | 415 ms | 1149 ms (2.8x slower) |

Both match the same 142 docs. The premise "modern sift compiles to faster testers" does
not hold here: v6 is a very optimized implementation, and the post-v6 rewrite (operation
tree, tree-shaking, operators passed via options) traded raw throughput for flexibility.

Reproducible bench (`npm i sift6@npm:sift@6.0.0 sift17@npm:sift@17.1.3`, then `node bench.js`):

```js
const sift6 = require('sift6')
const sift17mod = require('sift17')
const sift17 = sift17mod.default || sift17mod
const createEqualsOperation = sift17mod.createEqualsOperation

const N_DOCS = 10000, COMPILE_ITERS = 5000, EXEC_ITERS = 500
const docs = []
for (let i = 0; i < N_DOCS; i++) {
  docs.push({
    _id: 'id' + i, _type: 'io.cozy.files',
[...]

PR #1699, opened 2026-07-25
test: add store query benchmarks and PR-vs-master CI comparison

Crash--:
## Summary

There are no perf tests in the repo today. This adds a dependency-free benchmark suite for the store query reducer and a CI job that compares the PR against master on the same runner (ratios, robust to runner noise).

Benchmarks run against the real built reducer:
- `receiveQueryResult-getById-heavy` is the guarded metric (a regression >25% fails the job): ~1000 getById + 200 selector queries over 10k docs, the hot path that #1697 takes from ~38s to ~247ms.
- `selector-eval-folder-listing` and a compiled-sift canary are informational (they catch a sift bump like the one rejected in #1698).

The `perf` job benchmarks the PR, checks out the base SHA and benchmarks it back-to-back on the same runner, then a committed compare script posts a single table comment and fails only on a guarded regression. The comparator is a small Node script (no third-party action, no secrets). The 'no baseline on master yet' case is handled: this PR just publishes its numbers as the baseline.

## Summary by CodeRabbit

- **New Features**
  - Added automated performance benchmarking for pull requests.
  - Performance results are compared with the base branch and shared directly in pull request comments.
  - Builds can now detect and flag significant performance regressions.

- **Documentation**
  - Added guidance for running benchmarks locally and interpreting comparison results.

- **Chores**
  - Added a convenient benchmark command for generating performance reports.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -50,3 +50,77 @@ jobs:
           git remote set-url origin https://cozy-bot:$GH_TOKEN@github.com/cozy/cozy-client.git
           echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
           yarn lerna publish --yes -m "chore: Publish [skip ci]"
+
+  perf:
+    name: Performance benchmarks
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Enable Corepack
+        run: corepack enable
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+      - name: Record PR ref
+        id: prref
+        run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+      - name: Install dependencies (PR)
+        run: yarn install --immutable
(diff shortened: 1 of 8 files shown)


Crash--:
On the two workflow findings:

- **Stability (`||` masking)**: fixed — the base-branch step now guards on the benchmark harness existing (`benchmarks/run.js`), so only the "no benchmarks on base yet" bootstrap case is tolerated; a real build/install failure fails the job.
- **Security**: reviewed — the analysis didn't surface a concrete actionable issue. The perf job runs on `pull_request` (not `pull_request_target`), so a fork PR gets a read-only token; the `github-script` comment step uses the default scoped token. Happy to pin action SHAs or tighten `permissions:` if you have a specific concern in mind.

Crash--:
Hardened the perf job's token handling in 708363f:
- `persist-credentials: false` on checkout so the workflow token isn't left in `.git/config` for the untrusted PR code (`yarn install`/`yarn bench`) to read; the comment step uses github-script's ambient token, not git creds.
- Base/PR SHAs passed via env vars instead of `${{ github.event… }}` interpolation in the shell (command-injection defense).

For context on the residual risk: the job runs on `pull_request` (not `pull_request_target`) with least-privilege `permissions` (`contents: read`, `pull-requests: write`). GitHub gives fork PRs a **read-only** token (and withholds secrets), so untrusted fork code can't abuse the write scope; same-repo branch PRs run trusted code. If you want the gold-standard isolation, we can split into a `workflow_run` job that posts the comment without ever running PR code — happy to do that as a follow-up.

zatteo (approved):

PR #1702, opened 2026-07-26
feat(pouch): make the native SQLite query engine usable on op-sqlite

Crash--:
Makes cozy-pouch-link's native SQLite query engine (`SQLiteQuery`,
`db/sqlite/sqliteDb.native.js`) actually work with `pouchdb-adapter-react-native-sqlite`
(op-sqlite). Wired as `platform.queryEngine`, it previously threw `no such table:
by-sequence`, `database is locked`, and `UNIQUE constraint failed`. Now mango indexes
are built by SQLite (`CREATE INDEX` on `by-sequence`, off the JS thread) instead of
pouch-find materialising a **separate mapreduce-view SQLite DB per index**
(`<db>-mrview-<hash>.sqlite`, ~9840 row writes).

## Result (measured on a ~9840-file RN replica, iOS 18.3 sim)

| | pouch-find (mrview) | native engine |
|---|---|---|
| folder-listing index build | ~4500 ms | **~11 ms** |
| `readUI` total (first sync) | ~5300 ms | **~345 ms** |
| max JS-thread stall | ~2200 ms | **~160 ms** |
| separate view DB files | 6+ | **0** |
| lock / `no such table` errors | — | **0** |

Query results verified identical (folder listing, sort, thumbnails).

## The fixes

1. **Own connection in WAL mode.** `openDB` opens the engine's own op-sqlite connection
   on the same file, lazily, with `PRAGMA journal_mode=WAL` + `busy_timeout`. WAL lets it
   coexist with the adapter's writer without `database is locked` — the two-connection
   setup in rollback mode was the root of the mismatch + locks. **Self-contained: no
   dependency on the adapter, no global.**
2. **Winning-rev reads.** `getById`/`getByIds`/`allDocs`/`find` JOIN `document-store` on
   `winningseq` so only the WINNING revision of each doc is returned — correct under
   conflicts (a highest-rev JS pick can be wrong). Both sides are indexed
   (`by-sequence.seq` is the PK; the adapter indexes `document-store.winningseq`), so the
   join stays cheap. mango scans `by-sequence` via the requested index, then the join keeps
   only winning rows (matches CouchDB semantics).
3. **Non-unique docid index.** PouchDB's `by-sequence` keeps a doc's rev history, so several
   rows share `(doc_id, deleted)`; the `UNIQUE` index threw on any multi-rev doc.
4. **Graceful early queries.** `executeSQL` maps `no such table` / `database is locked` to
   an empty result so the earliest queries (before `by-sequence` exists) don't reject uncaught.

App side: set `platform.queryEngine = SQLiteQuery` (import from `cozy-pouch-link`; runtime
export, absent from types).

## Summary by CodeRabbit

- **Bug Fixes**
  - Queries now consistently return only the current winning revision, preventing stale results.
  - Improved selector-to-SQL translation for safer value quoting, correct boolean precedence, and accurate handling of empty `$in/$nin`.
  - Sorting now respects mixed directions per field.
  - More resilient database startup with WAL, a busy timeout, lazy initialization, and safer “missing table” error handling.
- **Tests**
  - Added/updated SQL translation coverage for native SQLite selector compatibility, including documented unsupported selector cases.

### DIFF
--- a/packages/cozy-pouch-link/src/db/sqlite/sql.js
+++ b/packages/cozy-pouch-link/src/db/sqlite/sql.js
@@ -86,6 +86,15 @@ export const parseResults = (
   }
 }
 
+// Quote a value for inline SQL. String single quotes are doubled ('' ) so a
+// value like "l'ete.txt" cannot break out of the literal (SQL error) or inject.
+const quoteSQLValue = value => {
+  if (typeof value === 'string') {
+    return `'${value.replace(/'/g, "''")}'`
+  }
+  return value
+}
+
 const parseCondition = (field, condition) => {
   const conditions = []
 
@@ -95,10 +104,15 @@ const parseCondition = (field, condition) => {
       let sqlOp = MANGO_TO_SQL_OP[operator]
 
       if (operator === '$in' || operator === '$nin') {
-        const values = condition[operator]
-          .map(v => (typeof v === 'string' ? `'${v}'` : v))
-          .join(', ')
-        conditions.push(`${sqlField} ${sqlOp} (${values})`)
+        const list = condition[operator] || []
+        if (list.length === 0) {
+          // "IN ()" is a syntax error. $in [] matches nothing (0 = false),
+          // $nin [] matches everything (1 = true).
+          conditions.push(operator === '$in' ? '0' : '1')
+        } else {
+          const values = list.map(quoteSQLValue).join(', ')
+          conditions.push(`${sqlField} ${sqlOp} (${values})`)
+        }
       } else if (operator === '$exists') {
         const value = condition[operator]
         if (value) {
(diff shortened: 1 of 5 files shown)


Crash--:
Thanks for the review — addressed in da7c3fa / 2e53133:

- **getByIds `IN` clause**: fixed — one quoted value per id (was one `"a,b"` literal). Added getById/getByIds tests.
- **executeSQL swallowing errors**: narrowed — only the transient pre-init `no such table` is mapped to empty; `database is locked` now surfaces (WAL + `busy_timeout` handle contention).
- **SQL fixtures**: updated to the winningseq-join shape.

Left for a follow-up (noted as heavy-lift): turning the lazy handle setup (WAL/timeout/index DDL) into a single awaited init promise. Today it's fire-and-forget; the only failure mode is a missing index (non-fatal — SQLite still returns correct rows, just unindexed) or WAL not applied. Happy to do the async-init refactor if you'd like it in this PR rather than a follow-up.

Crash--:
The two remaining heavy-lift review points — awaiting a single init promise (sqliteDb.native.js) and not mapping schema/lock failures to empty reads (sql.js executeSQL) — are the same robustness concern and are legitimate. Rather than expand this PR, I've captured them as a follow-up: linagora/cozy-client#1703.

Crash--:
Pushed `sql.mango.spec.js` + four mango-translation correctness fixes (commit a442603):

- escape single quotes in string values (a filename like `l'ete.txt` broke the literal / was an injection vector),
- parenthesise the mango expression so a top-level `$or` keeps the `DELETED = 0` guard (was leaking deleted docs),
- empty `$in` → constant instead of `IN ()` syntax error,
- per-field sort direction (mixed asc/desc was wrong).

Existing `sql.spec.js` expectations updated for the new parenthesised/where + per-field-sort shapes. The remaining unsupported operators (`$not`/`$regex`/`$elemMatch`/`$all`/`$size`, nested-object equality, `$eq null`) are `.skip` tests documenting the target SQL, tracked in #1704.

`yarn jest packages/cozy-pouch-link/src/db/sqlite/` → 42 passed, 7 skipped. Lint clean, `yarn types` no diff.

zatteo (approved):

PR #1709, opened 2026-07-30
feat: add effective recipients query support

shepilov:
## Summary

- Wraps the two new read-only endpoints from [linagora/cozy-stack#4885](https://github.com/linagora/cozy-stack/pull/4885): `GET /sharings/recipients/:file-id` and its shared-drive equivalent `GET /sharings/drives/:driveId/recipients/:file-id`.
- Returns the combined, deduplicated list of people who can access a file or folder, including access inherited from parent shared folders (`name`, `email`, `instance`, `status`, `read_only`, `can_edit_here`, `sources[]`).
- Exposes the projection through `Q("io.cozy.sharings.recipients").effectiveRecipients(fileId, { driveId })` for `client.query()` and `useQuery()` consumers.
- Stores results as file and drive-scoped virtual documents, replaces the complete snapshot after each successful refresh, and excludes the computed type from Pouch persistence.
- Flattens response items directly as computed projections because no generic `/data` collection backs them.

## Summary by CodeRabbit

* **New Features**
  * Added effective-recipient queries for files and folders, including direct and inherited access.
  * Added support for retrieving recipients within shared drives.
  * Added normalized recipient details, including permissions, access sources, and editability.
  * Added query support for selecting effective recipients by file or folder ID.
* **Documentation**
  * Updated API documentation and type definitions for effective-recipient features.
* **Bug Fixes**
  * Effective-recipient results are handled as fresh snapshots and are not persisted for offline use.


JF-Cozy:
is it ready for review @shepilov?

BTW, would it be possible to expose this through the query layer as well?

In twake-drive we'd like to consume this via `client.query()` / `useQuery` rather than a direct `client.collection()` call:
direct collection calls bypass the store (no `as` alias, no `fetchPolicy`, no cache mutualization), and `client.query` is what works across desktop and mobile.

The DSL already has a precedent for special-cased resolution (`sharingById`), so something like:

```js
Q('io.cozy.sharings').effectiveRecipients(fileId, driveId)
```

What do you think?

This way it could be resolved by the stack link to `SharingCollection.fetchEffectiveRecipients`, with the result stored under e.g. `as: 'io.cozy.sharings/effective-recipients/${fileId}'`. Apps would then get loading/error states and cache for free in the Share modal. It's the same fetchJSON transport underneath

docs/api/cozy-client/classes/QueryDefinition.md
```diff
@@ -273,7 +273,33 @@ See https://docs.cozy.io/en/tutorials/data/queries/#sort-data-with-mango
...
-[packages/cozy-client/src/queries/dsl.js:80](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/queries/dsl.js#L80)
+[packages/cozy-client/src/queries/dsl.js:81](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/queries/dsl.js#L81)
+
+***
+
+### effectiveRecipients
+
+▸ **effectiveRecipients**(`fileId`, `options?`): [`QueryDefinition`](QueryDefinition.md)
+
+Query the computed recipients who can access a file or folder.
```
JF-Cozy: It doesn't just retrieve the list of parent recipients for a subfolder—it retrieves the list of all recipients regardless of the situation, right? So from now on, should we use this every time, in every cases? Also when fetching the recipient for a parent folder for example? Or do we need to be more specific in the doc?

packages/cozy-client/src/links/StackLink.js
```diff
@@ -123,7 +129,7 @@ export default class StackLink extends CozyLink {
    * @param {QueryDefinition} query - Query to execute
    * @returns {Promise<import("../types").ClientResponse>}
    */
-  executeQuery(query) {
+  async executeQuery(query) {
```
JF-Cozy, line 132: Does changing this method to async really have that much of an impact? No need to `await executeQuery(` elsewhere?

packages/cozy-client/src/const.js
JF-Cozy, line 1: nit: don't know if we already expose all of them in the main barrel of the lib, but if we do, we should add DOCTYPE_SHARING_RECIPIENTS

JF-Cozy (approved):

packages/cozy-client/src/store/queries.spec.js
```diff
@@ -142,6 +142,36 @@ describe('queries reducer', () => {
...
+      const recipient1 = {
+        _id: 'file-1/recipient-1',
+        _type: 'io.cozy.sharings.recipients'
+      }
+      const recipient2 = {
+        _id: 'file-1/recipient-2',
+        _type: 'io.cozy.sharings.recipients'
+      }
+      const queryDefinition = Q(
+        'io.cozy.sharings.recipients'
```
JF-Cozy, line 156: why not Q('io.cozy.sharings').getEffectiveRecipients as we already have Q('io.cozy.sharings').sharingById or getSharingById or something like that?

Oh ok because `sharingById` is a different case: it queries actual `io.cozy.sharings` documents (mango on real docs), while effective recipients are a computed projection, not stored documents — the stack returns them with
`type: 'io.cozy.sharings.recipients'`

docs/api/cozy-client/classes/QueryDefinition.md
shepilov: Yes, we should use it whenever we need to display who can access a file or folder

packages/cozy-client/src/links/StackLink.js
shepilov: We already have code `await this.executeQuery(` but function was without async

packages/cozy-client/src/const.js
shepilov: we don't

PR #1711, opened 2026-08-27
feat(models): add the io.cozy.banners model

rezk2ll:
## Context

`io.cozy.banners` holds the platform messages shown to a user across applications. Every client that reads the doctype has to order the documents, evaluate a validity window, honour a dismissal, fall back on values it does not know and skip a shape it cannot read. Left to each application, those rules get written several times and drift.

## Solution

A model of pure functions for the read side, plus `dismiss` which retries once on a conflict since the field is idempotent.

The spec runs the contract vectors published with the doctype, so this implementation and the ones on other platforms are checked against the same cases rather than against each other. Polling is left to the caller: the model only decides what is visible at a given time.

## Summary by CodeRabbit

- **New Features**
  - Added support for retrieving trusted banners.
  - Banners now respect scheduling, priority, dismissal status, supported versions, and display surfaces.
  - Added fallbacks for unknown categories, severities, and surfaces.
  - Validated secure call-to-action links.
  - Added reliable banner dismissal with conflict recovery.

- **Documentation**
  - Added API documentation and TypeScript types for banners.

- **Tests**
  - Added coverage for banner visibility, ordering, filtering, fallbacks, trust, version handling, and dismissal behavior.

### DIFF
--- a/docs/api/cozy-client/modules/models.banner.md
+++ b/docs/api/cozy-client/modules/models.banner.md
@@ -1,0 +1,489 @@
+[cozy-client](../README.md) / [models](models.md) / banner
+
+# Namespace: banner
+
+[models](models.md).banner
+
+## Interfaces
+
+*   [Banner](../interfaces/models.banner.Banner.md)
+*   [BannerCta](../interfaces/models.banner.BannerCta.md)
+*   [DismissOptions](../interfaces/models.banner.DismissOptions.md)
+*   [VisibleBannersOptions](../interfaces/models.banner.VisibleBannersOptions.md)
+
+## Type aliases
+
+### BannerCategory
+
+Ƭ **BannerCategory**<>: `"quota"` | `"billing"` | `"trial"` | `"account"` | `"system"`
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:48](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L48)
+
+***
+
+### BannerClient
+
+Ƭ **BannerClient**<>: `default`
+
+A CozyClient instance
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:66](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L66)
+
+***
+
+### BannerSeverity
+
+Ƭ **BannerSeverity**<>: `"info"` | `"warning"` | `"error"`
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:52](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L52)
+
+***
+
+### BannerSurface
+
+Ƭ **BannerSurface**<>: `"banner"` | `"modal"`
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:56](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L56)
+
+## Variables
+
+### BANNERS_DOCTYPE
+
+• `Const` **BANNERS_DOCTYPE**: `"io.cozy.banners"`
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:7](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L7)
+
+***
+
+### BANNERS_QUERY_NAME
+
+• `Const` **BANNERS_QUERY_NAME**: `string`
+
+The name the banners query is stored under. It is namespaced rather than the
+bare doctype: `client.query` keys the store entry on this name but keys the
+in-flight promise on the definition, so an unrelated consumer naming its own
+`io.cozy.banners` query would make this one resolve nothing while that query
+is loading.
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:101](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L101)
+
+***
+
+### KNOWN_SEVERITIES
+
+• `Const` **KNOWN_SEVERITIES**: readonly [`BannerSeverity`](models.banner.md#bannerseverity)\[]
+
+The severities the doctype defines. A value outside this list is rendered
+with DEFAULT_SEVERITY. Frozen because the fallbacks read it at call time, so
+a consumer mutating it would change what every other consumer renders.
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:22](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L22)
+
+***
+
+### KNOWN_SURFACES
+
+• `Const` **KNOWN_SURFACES**: readonly [`BannerSurface`](models.banner.md#bannersurface)\[]
+
+The surfaces the doctype defines. A value outside this list is rendered on
+DEFAULT_SURFACE. Frozen for the same reason as KNOWN_SEVERITIES.
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:31](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L31)
+
+***
+
+### SUPPORTED_DOCTYPE_VERSION
+
+• `Const` **SUPPORTED_DOCTYPE_VERSION**: `1`
+
+The version of the io.cozy.banners shape this code was written against.
+A document declaring a higher cozyMetadata.doctypeVersion is skipped.
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:13](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L13)
+
+## Functions
+
+### buildBannersQuery
+
+▸ **buildBannersQuery**(): `Query`
+
+Query for the banners of the current instance.
+
+The query is named so every consumer shares one store entry rather than
+minting a new one per call, and the limit is lifted because the whole
+collection is read in full.
+
+*Returns*
+
+`Query`
+
+A named query
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:112](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L112)
+
+***
+
+### byPriority
+
+▸ **byPriority**(`a`, `b`): `number`
+
+Highest priority first, then bannerId ascending so every client orders
+identically when priorities are equal.
+
+The tie break compares code units rather than calling localeCompare, whose
+result depends on the runtime's locale and ICU build. A missing or non
+numeric priority sorts as 0 instead of producing NaN, which would make the
+comparator intransitive and the sort order engine defined.
+
+Documents that lost their bannerId fall back on \_id, as the deduplication
+does, so two of them do not compare equal and land in whatever order the
+database returned them in.
+
+*Parameters*
+
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `a` | [`Banner`](../interfaces/models.banner.Banner.md) | A banner |
+| `b` | [`Banner`](../interfaces/models.banner.Banner.md) | Another banner |
+
+*Returns*
+
+`number`
+
+The comparison result
+
+*Defined in*
+
+[packages/cozy-client/src/models/banner.js:255](https://github.com/linagora/cozy-client/blob/master/packages/cozy-client/src/models/banner.js#L255)
+
+***
+
+### dismiss
+
+▸ **dismiss**(`client`, `banner`, `options?`): `Promise`<{ `data`: [`Banner`](../interfaces/models.banner.Banner.md)  }>
+
+Records a dismissal. The field is idempotent, so a conflict is resolved by
+reading the document again and writing the value on the fresh revision.
+
+The dismissal is written to the document it is given. During a doctypeVersion
+migration the stack materializes the same bannerId under several versions,
+and the sibling documents are not reached from here.
+
+*Parameters*
+
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `client` | [`CozyClient`](../classes/CozyClient.md) | A CozyClient instance |
+| `banner` | [`Banner`](../interfaces/models.banner.Banner.md) & { `_id`: `string`  } | The banner to dismiss, as stored |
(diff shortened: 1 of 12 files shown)


packages/cozy-client/types/models/index.d.ts
```diff
@@ -1,4 +1,5 @@
 export const triggers: typeof trigger;
+import * as banner from "./banner";
```
MontaGhanmy, line 2: can you run `yarn types` and `yarn docs` before this goes in? `types/models/banner.d.ts` isn't there, so the `./banner` import here doesn't resolve (and `models.banner.md` is missing the same way).

ci stays green on it because the new files come back untracked and the check is just `git diff --exit-code`.

rezk2ll: good catch, thanks

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,350 @@
...
+ * The severities the doctype defines. A value outside this list is rendered
+ * with DEFAULT_SEVERITY.
+ */
+export const KNOWN_SEVERITIES = ['info', 'warning', 'error']
+
+/**
+ * The surfaces the doctype defines. A value outside this list is rendered on
+ * DEFAULT_SURFACE.
+ */
+export const KNOWN_SURFACES = ['banner', 'inline', 'modal']
```
zatteo: `inline` has been removed from spec yesterday

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,350 @@
...
+  const startsAt = toTime(banner?.startsAt)
+  const endsAt = toTime(banner?.endsAt)
+  if (Number.isNaN(startsAt) || Number.isNaN(endsAt)) return false
+
+  const at = now.getTime()
+  if (Number.isNaN(at)) return false
+  if (startsAt !== null && at < startsAt) return false
+  if (endsAt !== null && at >= endsAt) return false
+  return true
+}
```
zatteo, line 234: Can you check if in `date-fns` 2.x package that is already used in cozy-client there is a function to do this? Maybe it's worth it to use it.

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,350 @@
...
+ * A conflict reaches this code in two shapes: a FetchError from the stack,
+ * which carries the HTTP status, and a PouchDB conflict, which the pouch link
+ * rewraps into a plain Error that only keeps the message.
+ *
+ * @param {Error & {status?: number}} error - The error to classify
+ * @returns {boolean} True when the write lost a revision race
+ */
+const isConflict = error =>
+  error?.status === 409 || /Document update conflict/.test(error?.message ?? '')
+
```
zatteo, line 482: nit: Do something with `isDocumentUpdateConflict` ? idk

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,350 @@
...
+/**
+ * The links disagree on how a missing document comes back: the stack link
+ * resolves `{data: null}` while the pouch link resolves `{data: []}`, and an
+ * empty array is truthy. Both mean there is nothing to write to.
```
zatteo, line 508: Is it normal that both link return different on missing documents?

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,356 @@
...
+ * The version of the io.cozy.banners shape this code was written against.
+ * A document declaring a higher cozyMetadata.doctypeVersion is skipped.
+ */
+export const SUPPORTED_DOCTYPE_VERSION = 1
+
+/**
+ * The severities the doctype defines. A value outside this list is rendered
+ * with DEFAULT_SEVERITY.
+ */
+export const KNOWN_SEVERITIES = ['info', 'warning', 'error']
```
Crash--, line 22: I know it's stupid but with this code if you take a look at the generated type, you don't have any type... you just have `string[]`. jsdoc...

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,356 @@
...
+ * The query is named so every consumer shares one store entry rather than
+ * minting a new one per call, and the limit is lifted because the whole
+ * collection is read in full.
+ *
+ * @returns {{definition: object, options: object}} A named query
+ */
+export const buildBannersQuery = () => ({
+  definition: Q(BANNERS_DOCTYPE).limitBy(null),
+  options: { as: BANNERS_DOCTYPE }
+})
```
Crash--, line 115: so you want the consumer app to call `buildBannersQuery` and then use isInWindow, severity and things like that by themself? I think it's not a good idea.

rezk2ll: fair, now it's only exported just for testing ( or later? )

packages/cozy-client/src/models/banner.js
rezk2ll: thanks, i added union types

packages/cozy-client/src/models/banner.js
rezk2ll: good catch

packages/cozy-client/src/models/banner.js
rezk2ll: and from here also

packages/cozy-client/src/models/banner.js
rezk2ll: it's because the stack link returns { data: null } for a missing document, while the pouch link returns { data: [] }.

i created a separate issue for it here https://github.com/linagora/cozy-client/issues/1713

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,396 @@
...
+    )
+    // The stack deletes a banner when its condition clears, which is one of
+    // the ways the first write conflicts. There is nothing left to dismiss.
+    if (!fresh) return { data: null }
+    // Another client got there first, and the field is idempotent.
+    if (isDismissed(fresh)) return { data: fresh }
+
+    return client.save({ _type: BANNERS_DOCTYPE, ...fresh, dismissedAt })
+  }
+}
```
Crash--, line 635: I still don't understand how the third party app will consume this API.

Why not something like: 
```
banner.getActiveBanner() 
```

And in this getActiveBanner we've the request to fetch the banner with a good filter / index for the mango query and else? Because I don't think that doing the filter & isInTime manually is a good idea. 

And I don't think that the caller app should know all these informations. The caller app should just ask for the banner to display.

packages/cozy-client/src/models/banner.js
rezk2ll: looked into it. isDocumentUpdateConflict is only exported from Collection.js, not from the package root, so using it here means a deep import or widening cozy-stack-client's public API.

also it matches on error.message alone and calls .match without a guard, so it throws on an error with no message.

so imo not worth

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,390 @@
...
+ *
+ * Note that this only proves who created the document. An application holding
+ * write permission can still rewrite the fields of a stack authored one, which
+ * the platform cannot prevent, so a client escapes text and never interprets
+ * markup in it.
+ *
+ * @param {Banner} banner - The document to check
+ * @returns {boolean} True when the stack is the author
+ */
+export const isTrusted = banner => getCreatedByApp(banner) === STACK_APP
```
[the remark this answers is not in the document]
rezk2ll: it cannot be fixed in cozy-client :p therefore what i did:

dismiss no longer writes the caller's object. It re-reads the stored document and writes that back with only dismissedAt changed so a caller holding a resolved banner cannot overwrite the stack's severity, surface or cta ( Verified against the live stack locally, after dismissing, the document still reads severity: critical, surface: inline and keeps its original javascript: CTA. That closes the accidental-corruption path )

docs/api/cozy-client/modules/models.banner.md
```diff
@@ -0,0 +1,482 @@
...
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `client` | [`CozyClient`](../classes/CozyClient.md) | A CozyClient instance |
+| `options` | [`VisibleBannersOptions`](../interfaces/models.banner.VisibleBannersOptions.md) | - |
+
+*Returns*
+
+`Promise`<[`Banner`](../interfaces/models.banner.Banner.md)>
+
+The highest priority banner, or null
```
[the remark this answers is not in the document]
rezk2ll: @coderabbitai The source already declares it: @returns {Promise<Banner|null>}

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,601 @@
...
```
MontaGhanmy, line 138: the comment above is right that an app can write a document that looks like a platform message, but `createdByApp` is set by the app itself, not by the stack. cozy-client keeps whatever `cozyMetadata` the caller passes in (`CozyClient.js:771`), so an app can just put `createdByApp: 'stack'` and get through this check. it's still worth having, it just catches mistakes rather than a real attempt. could we say that in the JSDoc so nobody treats it as a guarantee?

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,601 @@
...
+  return first ?? null
+}
+
+/**
+ * Records a dismissal. The field is idempotent, so a conflict is resolved by
+ * reading the document again and writing the value on the fresh revision.
+ *
+ * The dismissal is written to the document it is given. During a doctypeVersion
+ * migration the stack materializes the same bannerId under several versions,
+ * and the sibling documents are not reached from here.
```
MontaGhanmy, line 560: the doctype says the dismissal should be written to every version of the `bannerId`, and here we only write the document we were given. it doesn't hurt us, since our dismissed set is keyed on `bannerId` and catches the siblings on read, but a client that reads only the older version would still show the banner. is that a deliberate call?

rezk2ll: right

MontaGhanmy (approved):

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,635 @@
...
+Object.freeze(KNOWN_SEVERITIES)
+
+/**
+ * The surfaces the doctype defines. A value outside this list is rendered on
+ * DEFAULT_SURFACE. Frozen for the same reason as KNOWN_SEVERITIES.
+ *
+ * @type {readonly BannerSurface[]}
+ */
+export const KNOWN_SURFACES = ['banner', 'modal']
+Object.freeze(KNOWN_SURFACES)
```
Crash--, line 32: but why?!

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,635 @@
...
+ * different definition. Both are indistinguishable from an empty collection
+ * once the response is read, so they are rejected here instead: reporting a
+ * failed read as "there is nothing" hides a banner the user should see, and
+ * reports a dismissal that never happened as done.
+ *
+ * @param {object} response - The response of a query
+ * @returns {object} The same response
+ * @throws {Error} When the query resolved nothing
+ */
+const requireResponse = response => {
```
Crash--, line 496: don't add this kind of wrapper... Why do you need that?

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,635 @@
...
+ * @param {BannerClient} client - A CozyClient instance
+ * @param {VisibleBannersOptions} [options] - Options
+ * @returns {Promise<Banner[]>} The banners to display, ready to render
+ * @throws {Error} When the doctype cannot be read, a missing permission included
+ */
+export const getActiveBanners = async (client, options = {}) => {
+  const { definition, options: queryOptions } = buildBannersQuery()
+  const data = requireResponse(await client.query(definition, queryOptions))
+    ?.data
+  return getVisibleBanners(Array.isArray(data) ? data : [], options)
```
Crash--, line 538: why we're not doing that from a mango query?!

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,635 @@
...
+  const siblings = await findSiblings(client, banner)
+
+  let result = { data: null }
+  for (const sibling of siblings) {
+    const written = await writeDismissal(client, sibling._id, dismissedAt)
+    // The caller is answered with the document it asked about.
+    if (sibling._id === id) result = written
+  }
+  return result
+}
```
Crash--, line 585: I don't understand why this kind of stuff is needed

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,635 @@
...
+/**
+ * Every stored document sharing the banner's identity, the one the caller holds
+ * included. The whole collection is read because the versions of a bannerId are
+ * not reachable by id, and the doctype keeps the collection small on purpose.
+ *
+ * @param {BannerClient} client - A CozyClient instance
+ * @param {Banner} banner - The banner being dismissed
+ * @returns {Promise<Banner[]>} The documents to write the dismissal to
+ */
+const findSiblings = async (client, banner) => {
```
Crash--, line 596: same. why do we need that?

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,635 @@
...
+ * empty array is truthy. Both mean there is nothing to write to.
+ *
+ * @param {object} response - The response of a getById query
+ * @returns {Banner|null} The document, or null when there is none
+ */
+const readOne = response => {
+  const data = requireResponse(response)?.data
+  const doc = Array.isArray(data) ? data[0] : data
+  return doc?._id ? doc : null
+}
```
Crash--, line 517: Why do we need this kind of helper?

packages/cozy-client/src/models/banner.js
```diff
@@ -0,0 +1,635 @@
...
+ * passes (CozyClient.js, ensureCozyMetadata spreads it last), so an application
+ * can set `createdByApp: 'stack'` on a document it authored itself and pass
+ * this check. It catches an application writing under its own name, which is
+ * the ordinary mistake, not one deliberately impersonating the stack. Enforcing
+ * that needs the doctype to be stack write only on the server side.
+ *
+ * It also only speaks to who created the document. An application holding write
+ * permission can rewrite the fields of a stack authored one, which the platform
+ * cannot prevent, so a client escapes text and never interprets markup in it.
+ *
```
Crash--, line 134: No application will have this right. We review carrefully all the manifest.webapp published to the registry. So this is not something we should be worried about. 

We can delete this code.

PR #1721, opened 2026-09-15
fix: Allow restoration of magic folder again

zatteo:
Normally when a magic folder is sent to the trash, when the user try to create it again, it should be restored. But this was not working due to this FIXME that is not relevant anymore.

_Found during_ https://github.com/linagora/cozy-stack/pull/4921

## Summary by CodeRabbit

- **Bug Fixes**
  - File deletion no longer removes associated “referenced by” links, preserving those references after the file is deleted.


doubleface (approved):

Crash-- (approved):
Let’s see 😅

zatteo:
> Let’s see 😅

Praying 🙏

PR #1722, opened 2026-09-15
fix: Resolve includes on a missing document to a null document

paultranvan:
A getById on a document that does not exist does not throw: Collection.get answers a 404 with `{ data: null }`. When the query also carried an `include`, fetchRelationships then walked that null document and attachRelationships read its `_id`, so the query rejected with "TypeError: Cannot read properties of null (reading '_id')" instead of resolving to the same null document as a query without include.

fetchRelationships now returns the response untouched when there is no document to include on, and attachRelationships passes a null document through, the way both already did for an empty list.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of queries for missing documents by preserving null results without attempting relationship processing.
  - Empty collection results continue to return safely.
  - Relationship data is no longer attached when a document is unavailable.

- **Tests**
  - Added coverage for missing-document queries and relationship handling.

- **Documentation**
  - Updated API reference links to reflect the latest source locations.


packages/cozy-client/src/CozyClient.spec.js
```diff
@@ -1296,6 +1296,20 @@ describe('CozyClient', () => {
       expect(requestHandler.mock.calls[0][0]).toBe(query)
     })
 
+    it('should resolve to a null document when a getById with includes misses', async () => {
+      // Collection.get answers a 404 with a null document instead of throwing
```
zatteo: nit: Useless AI comment

zatteo (approved):

packages/cozy-client/src/CozyClient.js
```diff
@@ -1207,7 +1207,9 @@ client.query(Q('io.cozy.bills'))`)
    */
   async fetchRelationships(response, relationshipsByName) {
     const isSingleDoc = !Array.isArray(response.data)
-    if (!isSingleDoc && response.data.length === 0) {
+    // A getById on a missing document resolves with a null `data` rather
+    // than throwing (see Collection.get): there is nothing to include on it.
+    if (isSingleDoc ? !response.data : response.data.length === 0) {
```
Crash--, line 1212: I read this line 5 times, and I'm not sure to really understand it yet

paultranvan: The comment was probably too narrow. I think it's clearer now?

packages/cozy-client/src/CozyClient.spec.js
paultranvan: It's really a pain, I updated my global AI guideline for less blabla, I'll see how it goes

packages/cozy-client/src/CozyClient.js
Crash--: by reading it 

if (
  (isSingleDoc && !response.data) ||
  (!isSingleDoc && response.data.length === 0)


I got it.

Crash-- (approved):

